PR #4, opened 2021-03-26
Add AWS tags

cnunciato:


leezen:
@cnunciato Ah, I didn't realize this was only in the CI account. In that case, it doesn't matter. For the prod account, we would want the same set of tags we apply to the PR buckets in `pulumi/docs`

PR #17, opened 2021-04-06
Make sure exported property matches Python naming conventions

davidwrede:
Ported from https://github.com/pulumi/docs/pull/5598

In addition to matching Python naming conventions, this fix also ensures the `curl` statement on https://www.pulumi.com/docs/get-started/azure/deploy-changes/ executes successfully when choosing Python as the language for the tutorial.


davidwrede:
Fixed in #106.

PR #21, opened 2021-04-06
fix console docs nav

susanev:
### Proposed changes

fixes #5565

intent is to go from this...
current-console-docs-nav

to this...
future-console-docs-nav

- [x] remove 3 category landing pages
- [x] move content if needed that was on those landing pages
- [x] expose individual pages in left nav
- [x] fix refs
- [x] fix redirects
- [x] move plan information to index page

... in the future i think we should consider moving org roles to the org page, and stack permissions to the projects and stacks page. also is it weird that the saml information isn't with these docs? anyways, wont be taking any of that on in this pr.


davidwrede (commented):
FYI: I updated the directory structure change of the topics because of the move from docs -> pulumi-hugo.

Also, I did a bit of rewriting to tie the Getting Started section and the section that talked about plans so that it flows a bit better. LMKWYT.

susanev:
thank you for cleaning up my mess!! ill take a closer look later today.

davidwrede (approved):

PR #23, opened 2021-04-06
Bump Hugo to 0.82

cnunciato:
To align with https://github.com/pulumi/docs/pull/5671, which upgrades Hugo to 0.82 as well.

https://github.com/gohugoio/hugo/releases


MitchellGerdisch (approved):
LGTM

PR #25, opened 2021-04-06
Update banners and add CTAs for PulumiUP in docs and blog pages

zchase:
This PR updates the banner/alert to promote PulumiUP and adds some CTAs for PulumiUP in the right nav of the docs and blog pages. 

This PR was originally two PRs in the old repo:
- https://github.com/pulumi/docs/pull/5663
- https://github.com/pulumi/docs/pull/5637


cnunciato (commented):
A few suggestions, but nothing majorly blocking code-wise. However it does seem like we could do better with the actual image; we rarely (never?) use sharp edges or black borders, and the button looks a bit Windows-Paint-y compared to the style we're using for orange buttons elsewhere: 

image

For that matter, why not just use regular Tailwind-styled HTML for this, since it's the querystring that seems to be the important part? Is an image actually necessary?

cnunciato:
Just added an image to consider as an alternative. Lemme know what you think.

cnunciato (approved):
Giving a 👍 to help move it along, though it'd be nice to get some of these other things addressed as well. If the image looks okay to you all, feel free to merge when you're comfortable.

PR #26, opened 2021-04-06
Python s3.BucketObject takes source argument, not content

jaybuff:
Using content will result in this error:

    Diagnostics:
      aws:s3:BucketObject (index.html):
        error: unexpected asset content


davidwrede (approved):
Thanks for the fix!

PR #30, opened 2021-04-07
Update 3.0 migration instructions to match beta

leezen:
We'll want to actually have the migration guide explain beta usage until we get to the actual 3.0 release.


stack72 (approved):

stack72:
Great catch!!

PR #32, opened 2021-04-07
Remove duplicate CSS config

cnunciato:
The files we're actually using for asset builds are in `assets/config`, so these can be removed, and `vendor/fontawesome` folder was renamed to just `fontawesome`.


cnunciato:
Oops! I think I deleted the wrong files. Let's try that again...

cnunciato:
All righty, that looks better I think. 😅

davidwrede (approved):

PR #50, opened 2021-04-13
Update integration test guide to account for Automation API

leezen:
Fixes https://github.com/pulumi/home/issues/1293

### DIFF
--- a/themes/default/content/docs/guides/testing/integration.md
+++ b/themes/default/content/docs/guides/testing/integration.md
@@ -146,12 +146,6 @@ If you are using GitHub, Pulumi offers a [GitHub App]({{< relref "/docs/guides/c
 By leveraging Pulumi for your core acceptance test workflow, you'll
 unlock new automation capabilities that improve your team's productivity and confidence in the quality of changes.
 
-## Full Examples
-
-A minimal example of using Pulumi's Go integration test framework is available in the examples repository: [Integration Testing in Go](https://github.com/pulumi/examples/tree/05ae8e1803d7f44cecac69589175e416e421cdfe/testing-integration).
-
-Our own [integration test suite](https://github.com/pulumi/examples/blob/05ae8e1803d7f44cecac69589175e416e421cdfe/misc/test/examples_test.go) can be a source of further inspiration.
-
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/guides/testing/integration.md
```diff
@@ -154,3 +151,22 @@ unlock new automation capabilities that improve your team's productivity and con
...
+**automation.ts**
+```typescript
+export async function getDeployment(): Promise<Deployment> {
+  const stack = await LocalWorkspace.createOrSelectStack(args);
+  
+  return stack.exportStack();
+}
+```
+
+You can then iterate through the `Deployment` object to check that the expected resources and property values are present.
```
EvanBoyle, line 172: Suggested addition: 

We even write tests inside pulumi using Automation API and language-specific frameworks like `mocha`. Take a look at the source for yourself:

- [NodeJS tests](https://github.com/pulumi/pulumi/blob/d93e5acf675fe8f63cfcf0c028cf996f97baf621/sdk/nodejs/tests/automation/localWorkspace.spec.ts#L189)
- [Go tests](https://github.com/pulumi/pulumi/blob/d93e5acf675fe8f63cfcf0c028cf996f97baf621/sdk/go/x/auto/local_workspace_test.go#L734)
- [Python tests](https://github.com/pulumi/pulumi/blob/0c4e1a33e058a5b129ba009c5a58793200c09fff/sdk/python/lib/test/automation/test_local_workspace.py#L291)
- [C# tests](https://github.com/pulumi/pulumi/blob/0e2d9185870d1cc343c3f75499ec4265ba072292/sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs#L345)

(URLs will change with 3.0)

EvanBoyle (approved):
Looks great. Only suggestion is to point towards the language specific tests we write for automation API itself. These are technically integration test, and are useful working references (rather than having to create a bunch of new examples).

themes/default/content/docs/guides/testing/integration.md
leezen: Great suggestion!

PR #63, opened 2021-04-14
Update azure aks helm examples

t0yv0:
This brings the new Azure aks/helm examples in as you requested @mikhailshilkov 

TypeScript example was already in

TODOS:

- [x] I have run `mktutorial` in a hacky way, is there official way to run it? I also manually edited what mktutorial wanted to add, which an be a minor nuisance for the next invocation (it wanted to add more examples from the repo); should I just bring those in as well?

- [x] IF we want to remove examples without helm chart, like "Azure Kubernetes Service (AKS) Cluster using the native Azure Provider" what is the best way to remove them? I think if I use remove them from here then mktutorial will bring them back in. Perhaps I can PR to examples repo and rename those examples using a naming scheme such that mktutorial won't pick them up?

- [x] Do we want "Featured Tutorials" links to direct to -aks-helm* examples instead of plain aks examples or leave as is? 

Thank you!


mikhailshilkov:
> It wanted to add more examples from the repo, should I just bring those in as well?

Yes.

> IF we want to remove examples without helm chart

We don't want to remove anything.

>  Do we want "Featured Tutorials" links to direct to -aks-helm* examples instead of plain aks examples or leave as is?

Yes, we want to change to aks-helm.

Thank you!

t0yv0:
Thanks - then this does the trick I think.

mikhailshilkov:
Thank you!

The large "Azure Kubernetes Service" link still leads to the old example. Also, could you update the same link on http://pulumi-hugo-origin-pr-63-b1188801.s3-website.us-west-2.amazonaws.com/docs/tutorials/ ?

t0yv0:
Got two more links redirected!

Now one more interesting place remains:

```
> If you are looking to use Kubernetes on Azure, see [the AKS tutorial]({{< relref "../kubernetes/aks" >}}).
```

That one seems actually https://www.pulumi.com/docs/tutorials/kubernetes/aks/ quite detailed, but AFAIK it uses the classic Azure provider. Should we replace that with azure-py-aks-helm as well?

t0yv0:
@mikhailshilkov --^ thanks

mikhailshilkov (approved):
Looks great!

t0yv0:
OK merging for now. If we want to redirect the "AKS tutorial" I can do so separately.

PR #81, opened 2021-04-19
Fixed dereference issue in K8s Getting Started tutorial

davidwrede:
Got a nil pointer dereference runtime error when running through the tutorial with the new ApplyT changes, so I added in this check.

Note: I wasn't sure what to return if the minikube `ClusterIP` value is nil, so I just return an empty string. Happy to change it to whatever the best practice is.

### DIFF
--- a/themes/default/content/docs/get-started/kubernetes/modify-program.md
+++ b/themes/default/content/docs/get-started/kubernetes/modify-program.md
@@ -233,7 +233,10 @@ func main() {
 
         if isMinikube {
             ip = frontend.Spec.ApplyT(func(val *corev1.ServiceSpec) string {
-                return *val.ClusterIP
+                if val.ClusterIP != nil {
+				    return *val.ClusterIP
+			    }
+			    return ""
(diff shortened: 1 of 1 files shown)


lblackstone (approved):

themes/default/content/docs/get-started/kubernetes/modify-program.md
```diff
@@ -233,7 +233,10 @@ func main() {
 
         if isMinikube {
             ip = frontend.Spec.ApplyT(func(val *corev1.ServiceSpec) string {
-                return *val.ClusterIP
+                if val.ClusterIP != nil {
+				    return *val.ClusterIP
```
joeduffy, line 237: Are tabs vs spaces off here?

davidwrede: Yeah, I double checked my editor settings in VS Code (and they right from within the editor), so I'll take a stab at just updating them in TextEdit or something.

PR #100, opened 2021-04-20
Remove countdown and update button text on PulumiUP landing page

zchase:
Just as the title states. Now that the event is over we should remove the countdown.


cnunciato (approved):
TBH I kind of love it, but alas, yeah.

PR #103, opened 2021-04-20
Remove warning about unchecking gRPC FUSE warning

chrsmith:
We had a warning about a Docker x macOS bug. However, it has since been fixed. (Or at the very least, we cannot reproduce the problems you encounter if you have that option checked.) So removing it from the docs.

image

cc https://github.com/pulumi/pulumi-ee/issues/10


jkisk (approved):
Chris ran me through this on my Mac and I can confirm that this works, with the gRPC FUSE feature switched on.

PR #106, opened 2021-04-20
Update deploy-changes.md

litvos:
use earlier defined variable staticEndpoint


davidwrede (approved):
Thanks for the quick fix!

Issue #86, opened 2021-04-20
Landing page has a gap on a large screen


mikhailshilkov:
I have a 3840x2160 monitor and I open the website half-width full-height, this is what I see now:

![image](https://user-images.githubusercontent.com/1454008/115358901-6bd65980-a1be-11eb-99fe-72ffeb9d5fb3.png)

I think we could reduce the spacing between the top section and the bottom one to make it look better and fit more useful content.

I'm on a Mac/Firefox.

susanev:
@kimberleyamackenzie would love your thoughts on this

kimberleyamackenzie:
Yeah, I think maybe we just ought to define a max-height for the homepage hero.  I'll play around with some numbers and make a PR.

susanev:
this was fixed.

PR #107, opened 2021-04-20
Add max height for homepage hero

kimberleyamackenzie:
This addresses https://github.com/pulumi/pulumi-hugo/issues/86, to decrease the gap between the homepage hero and content on especially tall screens.

With this change, for a screen height of 2150:
Screen Shot 2021-04-20 at 12 40 17 PM


mikhailshilkov:
Yes, this looks better to me 👍 

image

susanev (approved):

PR #112, opened 2021-04-21
update logo in page title

sean1588:
I noticed the old logo in the page title, so figured I'd open a PR. I _think_ this will update it, but not an expert in this repo so.... 

this is how it looks as it is currently deployed. I just added this as an example so others know which logo I am talking about.
Screenshot from 2021-04-20 18-31-35


susanev:
@cnunciato do you know where this image displays on the website? that will help me figure out if we want just the cube only or with the background color like @sean1588 did here.

sean1588:
from the generated site preview...

Screenshot from 2021-04-20 18-43-01

susanev (approved):

PR #114, opened 2021-04-21
pulling serverless blog posts at request of Aaron

spara:


aaronkao (approved):
These are the reasons this blog is being pulled:

1/ I don't agree with the org size view because I have seen sufficiently large and not-young companies adopt serverless (e.g., Nike, Fannie Mae, Fender, Comcast). 
2/ The architecture comment is also glossing over the fact that yes it isn't lift and shift, but there are many apps that can easily be refactored over. 
3/ I would approach architecture from a use case point of view like (e.g., data processing or it automation or batch workloads work great). 
4/ On costs, the data I collected at aws showed significant cost savings unless you were genius at using auto scaling and instance selection. 
5/ There is a lot of things that make serverless more secure than an instance...there are still things that the customer has to secure, but more of the security in the stack is in the cloud providers hands. So I really don't agree with the security comments.

PR #118, opened 2021-04-21
Update the /pulumi-up landing page.

zchase:
This PR updates the /pulumi-up landing page and I also snuck in a fix to some cut off text on the pricing page by removing the max-height animation, which turns out was not a great idea given the varying sizes of the text blocks. I'll revisit and layer some JS on top to bring the animation back, but it doesn't look like a pure CSS animation is feasible here.

Fixes https://github.com/pulumi/marketing/issues/36


themes/default/static/logos/tech/jfrog.svg
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.0//EN" "http://www.w3.org/TR/2001/REC-SVG-20010904/DTD/svg10.dtd">
+<!-- Creator: CorelDRAW 2018 (64-Bit) -->
```
cnunciato, line 3: Whoa, CorelDRAW -- there's a name from the past.

themes/default/content/pulumi-up/_index.md
```diff
@@ -23,7 +23,7 @@ event_sections:
 
     - title: The Tools
       act: II
-      speakers: [ "luke-hoban", "sophia-parafina", "komal-ali" ]
+      speakers: [ "luke-hoban", "lee-zen", "komal-ali" ]
```
cnunciato, line 26: I see we have this layout on the page now:

image

... but is it right? Should Komal's image be below Luke's? It looks somehow off -- although I'm guessing the intent may be to "center" the second row? (If so, ignore me.)

themes/default/assets/sass/_accordion.scss
```diff
@@ -13,6 +13,10 @@
         transition: max-height 1s ease-out;
     }
 
+    .accordion-item-body-no-animation {
```
cnunciato, line 16: TBH, I much prefer the behavior _without_ the animation, as it actually felt pretty sluggish to me. This feels snappy and does what I expect, so I'd personally suggest we just nix the animation altogether.

cnunciato (approved):
LGTM, though I defer to @susanev on the SVGs, as I know we did some work in the redesign to make things more consistent size-wise. (And I also just ❤️ Figma's super-clean SVG exports.)

themes/default/assets/sass/_accordion.scss
zchase: 👍 Yeah I agree with you. I'll cut a different PR to update the other accordions once this makes it through.

themes/default/content/pulumi-up/_index.md
zchase: 👍 Yeah I'll need to play with the alignment to try and clean this up.

zchase:
@cnunciato @susanev I am going to merge this as-is so that we can get the recording onto the website (something we promised to get out this week). Let's circle back on the SVG and alignment issues.

PR #129, opened 2021-04-22
Update social card image

zchase:
The current twitter card was rendering weird because the sizing was off. This PR adds a correctly sized image for the Twitter cards.

current-twitter-card

One thing to note is that I started looking at this because I noticed that our Twitter cards still had the old branding. I learned that Twitter caches these cards so when we update a Twitter card we should run it through the [validator](https://cards-dev.twitter.com/validator) so that it clears the cache.


cnunciato:
Should this maybe be a bit larger? I only ask because when I run it through the validator, it looks a little soft-edged to me:

image

susanev:
@cnunciato what do you mean by soft-edged?

@zchase lets use the stacked logo I did for wendy for this space.
https://drive.google.com/file/d/1_xg4Y5ho_uLm4cL2k33QKazw5Q8IhrcB/view?usp=sharing

cnunciato:
@susanev Sorry, what i meant was that it looks "lossy", but I think the image is actually too small, so it's being upscaled:

From https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary:

> Images for this Card support an aspect ratio of 1:1 with minimum dimensions of 144x144 or maximum of 4096x4096 pixels

So maybe a PNG of >= 2X the minimum would be good (say, 300x300), since it looks like SVG isn't supported.

cnunciato:
I'll push something onto this branch so we can see how it looks.

susanev:
looks good to me

cnunciato:
Yeah that looks pretty good to me -- what do you think @susanev?

image

cnunciato:
@zchase If this looks good to you, feel free to merge away.

cnunciato:
Merging just to move this along -- @zchase if you see anything that needs fixing, let me know and I'll address in a follow-up!

PR #141, opened 2021-04-26
Fix MockCallArgs property in Golang example

lenardchristopher:
The `MockCallArgs` does not have `Inputs` as a property. This should be `Args`.

Source code: https://github.com/pulumi/pulumi/blob/master/sdk/go/pulumi/mocks.go#L34

Doc page: https://www.pulumi.com/docs/guides/testing/unit/#add-mocks _select golang examples_


stack72 (approved):
Thanks so much for this! you are correct here @lenardchristopher

Issue #145, opened 2021-04-27
Add Libvirt provider to the cloud providers page


stack72:
Now that the provider is published, we need to ensure there's a cloud provider entry for this

PR #147, opened 2021-04-27
Adding Libvirt provider to cloud-providers page

stack72:
Fixes: #145


davidwrede (commented):
Page looks good (just a minor nit). Btw, do we need a setup page for libvirt too?

stack72:
> Page looks good (just a minor nit). Btw, do we need a setup page for libvirt too?

WE don't need a setup page for it - it's a utility like cloudinit that we can interact with rather than a SaaS / PaaS system

davidwrede (approved):

PR #155, opened 2021-04-27
Change from `bucket-path` to `bucket-name`

jwfuller:
Change the variable name to better match AWS documentation. We want a user to enter `s3://my-bucket` rather than `s3://my-bucket.s3-us-east-1.amazonaws.com`


davidwrede (approved):
Good catch!

lukehoban:
This change likely still makes sense - but note that I believe the reason it used the term "bucket-path" was that it could be of the form `s3://my-bucket/somepath/someotherpath`.  And in fact that form is often useful to store multiple projects' state in the bucket.  Might be worth also adding a note that the bucket name can include a path inside the bucket.

davidwrede:
Yup, good point. I'll get a PR up soon.

PR #158, opened 2021-04-28
Note that bucket name for S3 backend can contain multiple folders

davidwrede:


cnunciato (approved):

themes/default/content/docs/intro/concepts/state.md
```diff
@@ -156,6 +156,10 @@ To use the [AWS S3](https://aws.amazon.com/s3/) backend, pass the `s3://<bucket-
 $ pulumi login s3://<bucket-name>
 ```
 
+{{% notes type="info"%}}
+The `bucket-name` value can include multiple folders, such as `my-bucket/teamprojects/project1/`. This is useful when storing multiple projects' state in the same bucket.
```
cnunciato, line 160: Is the trailing slash okay? Just figured I'd ask the question, just in case, since I don't often use the S3 backend.

davidwrede: I tested it both ways and it seemed to work. In fact, the S3 UI shows folders with a trailing slash.

davidwrede: While it works both ways, I'll remove the trailing slash. S3 UI specifically doesn't want people to add that in when creating their folders, so I don't want people to think they should put that in their folder name(s).

PR #165, opened 2021-04-29
Update _index.md

aaronkao:
Corrected on-premise to on-premises. A premise is a statement or idea taken to be true and on which an argument or reasoning may be based. Premises is a house or building, together with its land and outbuildings, occupied by a business or considered in an official context.


cnunciato (approved):
⭐ 💯

PR #167, opened 2021-04-29
Fix Stencil async warnings

cnunciato:
Stencil's `@Method` decorator [expects methods to be defined `async`](https://stenciljs.com/docs/methods), so this change just declares them async to suppress the warnings we see in the console in development.

Update: Turned out we had a bit of logic to clean up as well (e.g., only use the `autoplay` module when autoplay is enabled), so did that, and tightened up some typings as well.

### DIFF
--- a/themes/default/components/src/components/convert/convert.e2e.ts
+++ b/themes/default/components/src/components/convert/convert.e2e.ts
@@ -1,11 +1,11 @@
-import { newE2EPage } from '@stencil/core/testing';
+import { newE2EPage } from "@stencil/core/testing";
 
-describe('pulumi-convert', () => {
-  it('renders', async () => {
-    const page = await newE2EPage();
-    await page.setContent('<pulumi-convert></pulumi-convert>');
+describe("pulumi-convert", () => {
+    it("renders", async () => {
+        const page = await newE2EPage();
+        await page.setContent(`<pulumi-convert from="tf" endpoint="https://someserver/somepath"></pulumi-convert>`);
 
-    const element = await page.find('pulumi-convert');
-    expect(element).toHaveClass('hydrated');
-  });
+        const element = await page.find("pulumi-convert");
+        expect(element).toHaveClass("hydrated");
+    });
 });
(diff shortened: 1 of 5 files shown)


cnunciato:
@zchase please take a look I think I've reviewed all the places where these components are used, but it'd be good to get your 👀 on this stuff as well.

themes/default/components/src/components/convert/convert.e2e.ts
```diff
@@ -1,11 +1,11 @@
...
+import { newE2EPage } from "@stencil/core/testing";
 
-describe('pulumi-convert', () => {
-  it('renders', async () => {
-    const page = await newE2EPage();
-    await page.setContent('<pulumi-convert></pulumi-convert>');
+describe("pulumi-convert", () => {
+    it("renders", async () => {
+        const page = await newE2EPage();
+        await page.setContent(`<pulumi-convert from="tf" endpoint="https://someserver/somepath"></pulumi-convert>`);
```
cnunciato, line 6: To quiet the `pulumi-convert` test, which was complaining about these attributes being missing.

themes/default/components/src/components/swiper/swiper.tsx
```diff
@@ -68,45 +71,39 @@ export class Swiper {
 
         SwiperJS.use(modules);
 
-        this.containerClass = `swiper-container-${this.swiperHash}`;
-        this.nextBtnClass = `swiper-button-next-${this.swiperHash}`;
-        this.previousBtnClass = `swiper-button-prev-${this.swiperHash}`;
+        this.containerClass = `swiper-container-${this.swiperID}`;
```
cnunciato, line 74: Just for clarity, since this seems to refer to some sort of randomish unique identifier.

zchase: 👍 Without the ID you get some unexpected behavior with multiple swipers on one page.

zchase (approved):
LGTM. All the places we use the swiper seems to be working as expected in preview. Thanks for cleaning this up, its been in the back of my head for a bit.

PR #173, opened 2021-04-30
Fix Python syntax in dependsOn example

Levitron:


stack72 (approved):
thanks so much for this!

PR #175, opened 2021-04-30
[CloudFlare] Update install instructions

gunar:
CloudFlare API Keys are legacy but luckily pulumi-cloudflare already accepts API Tokens.


stack72 (approved):

stack72:
Thanks so much for this @gunar - you are indeed correct here!!

gunar:
Damn you're fast! Cheers

PR #179, opened 2021-05-01
Fix digitalocean example

mhitza:
TypeScript reported that `do` is a reserved keyword and the homepage example didn't work out of the box as a result of that. The full API reference examples use `digitalocean` already.


stack72 (approved):
Thanks so much for this!

PR #180, opened 2021-05-03
clean up contact us page

### LINKED ISSUE #119: Update contact page to help user with best contact option
### Context 

We frequently have people use our general Contact Us form for technical support items and also don't distinguish between sales inquiries and general inquiries.

### Suggestion 

Our page at https://www.pulumi.com/contact/ should provide help people get to the right form or option as quick as possible - e.g.:
- **Technical Support** - should link to https://support.pulumi.com/
- **Sales** - should take the user to a sales-specific form (can be the same generic form for now)
- **Other Inquiries** - should take the user to a generic contact form (likely the same form used right now)

### Mockup

To better illustrate what I have in mind, here's a mockup. I'm not tied to the styling or wording, but do think it should be this simple.

image

susanev:
fixes #119

@clstokes pls take a look at this, I did not use your designs. but I did add some details about who should use the form and who should use the support portal. happy to talk through the design decisions over zoom if you'd like.

also @clstokes can I pls have access to fix that support landing page? its using the wrong fonts and a few other things, id love to get that cleaned up.

@zchase feel free to add commits if I got anything wrong, and I really want to remove the requirement to include a phone number on that form, is that possible?

### DIFF
--- a/themes/default/layouts/page/contact.html
+++ b/themes/default/layouts/page/contact.html
@@ -1,4 +1,3 @@
 {{ define "main" }}
     {{ partial "contact-us.html" . }}
-    {{ partial "social.html" . }}
 {{ end }}
--- a/themes/default/layouts/partials/contact-us.html
+++ b/themes/default/layouts/partials/contact-us.html
@@ -5,7 +5,7 @@
                 <h2 class="text-4xl">How can we help?</h2>
                 <p>Have general questions? Want to talk about pricing or partnerships?</br> Reach out to us using this form.</p>
 
-                <div class="hs-form hs-form-fg-dark w-1/2 m-auto">
+                <div class="hs-form hs-form-fg-dark md:w-1/2 sm:w-full m-auto">
                     {{ partial "form-section" (dict "form" .Params.form) }}
                 </div>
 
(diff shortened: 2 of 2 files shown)


zchase:
@susanev I wouldn't mind if we but the text that calls out if the person is looking for technical support above the form, because I think some of motivation for this work is that people were submitting Contact Us when they should be submitting a ticket. I also would like to implement something like Cameron's original design, but as a component on our marketing pages. I'd like to consolidated as many forms as possible, as it will make it operationally easier to manage all the requests. For example, we have a contact form on the bottom of `/aws` and I'd like that section to be replaced w/ something close to Cameron's mock.

Outside of that, I think this is good. As far as removing the phone number that is up to the sales team and we don't need a PR to change it. We can change it directly from HubSpot.

susanev:
@zchase lets talk more about Cameron's design over zoom some time soon, I have some usability concerns about it.

i can move the technical support up, but do we have more data on where people were right before they came to contact us? that might help us figure out why people are landing here when they are hitting issues.

susanev:
@zchase ive moved the support direction up (though i do want to call out that I think its a bit confusing). but either way, can you add a heading to the form as well something like `Contact Us` or similar?

susanev:
im closing this because it seems multiple folks are working on this... will come back when the requirements are more clear.

PR #181, opened 2021-05-03
Remove Docs from the Learn menu

cnunciato:
WIth Docs as a top-level item now, we can remove it from Learn.


susanev:
@cnunciato pretty sure you need to remove it from the mobile menu too, a bit farther down on the same file

cnunciato:
> pretty sure you need to remove it from the mobile menu too

Ah, thanks! Missed that.

susanev (approved):

PR #182, opened 2021-05-03
redirect /support to https://support.pulumi.com/

### LINKED ISSUE #111: Redirect support page to support portal
Currently https://www.pulumi.com/support/ is a contact form which goes to a generic email alias when users should use https://support.pulumi.com/ instead.

The `/support` page is still linked from various locations and indexed by search engines. Can we 301 redirect (or whichever status is best) to https://support.pulumi.com/ to help users get to support faster?

susanev:
with this change if users go to /support they should be directed to https://support.pulumi.com/

fixes #111


cnunciato (commented):
@susanev Can you remove [this page](https://github.com/pulumi/pulumi-hugo/blob/master/themes/default/content/support.md) as well? It's the content file in use today.

cnunciato:
Note that with this approach, we'll still generate an HTML page (containing  a `meta` tag that tells the browser to redirect):

```
<meta http-equiv=refresh content="0; url=https://support.pulumi.com/">
```

.. so if you happened to `curl` the preview URL at `/support`, you'd get a 200, and not the 301 the issue requests:

```
curl -I http://pulumi-hugo-origin-pr-182-ddfd9dcf.s3-website.us-west-2.amazonaws.com/support/
HTTP/1.1 200 OK
```

However, when we release the website "for real" (via pulumi/docs) , we run a post-deploy job that converts all of these `meta`-style redirects to proper HTTP 301s, so while things might not look technically correct here, they should be fine when this ultimately goes out.

cnunciato (approved):
🚀

PR #190, opened 2021-05-04
add tutorial to cloud buttons

susanev:
## overview

recently some folks shared that they had no idea where the cloud buttons would take them, was it to sign in, to the cloud websites, something else?

this adds the word tutorial to those buttons to make it more clear where it will send folks (though it would probably be better to be like aws + pulumi tutorial...)

let me know what yall think or if you have another idea

note: the aws logo is a special beast, ill fiddle with it later.

## before
before-docs-landing

before-get-started

## after
after-docs-landing

after-get-started

### DIFF
--- a/themes/default/content/docs/_index.md
+++ b/themes/default/content/docs/_index.md
@@ -22,15 +22,19 @@ Pulumi is an <a href="https://github.com/pulumi/pulumi" target="_blank">open sou
     <div class="md:flex justify-between items-center">
         <a class="block rounded hover:bg-gray-200 transition-all my-2 py-4 text-center px-6" href="{{< relref "/docs/get-started/aws" >}}">
             <img class="inline-block h-8 w-auto -mb-2" src="/logos/tech/aws.svg">
+            <p class="mb-0  mt-2.5">Tutorial</p>
         </a>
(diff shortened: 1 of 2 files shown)


zchase (approved):
Seems reasonable to me. My only question is whether it should be labeled `Guided` instead of `Tutorial` because we refer to these things as Guides in other places.

davidwrede:
Correct, these should be labeled as tutorials as they are step-by-step instructions centered around one scenario/goal.

lukehoban:
I'll admit I find the word "Tutorial" in these contexts doesn't feel right.  It's really "get started with AWS".  

In the content itself we call it a "guide" - and I think that would visually look better here - though the UX of the text additions feels a bit out of place to me generally - especially on the main docs page.  That said - I don't have immediate suggestions on alternatives.  Are there other options to consider for addressing the feedback here?

susanev:
@lukehoban im not sure that "get started with AWS" explains where we are going to send you either though...

i can spend more time here on improving the design more drastically, but I was thinking this might be good for a quick fix first. does that sound good? or do you want me to hold off until the design is improved?

cnunciato:
I'd prefer we held off on merging this until the design is improved more holistically, yeah. Happy to hop in and help with that if you like.

susanev:
im going to close this, but i do think we are missing a quick opportunity to make this better. i will work on redesigning our buttons for starting the tutorials.

PR #193, opened 2021-05-04
Don't super-size button labels

### LINKED ISSUE #170: get started button issues on wide screens
originally reported by @isaac-pulumi 

if you extend your window size large enough the get started button text wraps
it should never wrap

we believe it starts wrapping around 2020px

image

cnunciato:
There's not really a need to make buttons huge on wide screens, especially when everything else stays the same size. 

Fixes #170.


cnunciato:
Here's what the size looks like now (with these changes) at 3440 wide:

image

susanev (approved):
thank you!!

Issue #198, opened 2021-05-06
policy related docs and getting started guide dont mention the enterprise requirement


susanev:
## overview

a user reported through community slack that they invested a bunch of time setting up policy packs before they learned they were only available on enterprise plans. 

also noticed a numbering issue on this page: https://www.pulumi.com/docs/get-started/crossguard/enforcing-a-policy-pack/ when reviewing

these were the docs pages they mentioned

- https://www.pulumi.com/docs/get-started/crossguard/
- https://www.pulumi.com/docs/guides/crossguard/
- https://www.pulumi.com/docs/get-started/crossguard/enforcing-a-policy-pack/

## proposed fixes

- make it more clear in our docs what requires enterprise
- make policy-related features more clear on the pricing page

PR #200, opened 2021-05-06
fix numbering

susanev:
noticed this numbering issue when reviewing the policy docs
/docs/get-started/crossguard/enforcing-a-policy-pack/
fixes part of #198 

## before
before

## after
after


cnunciato (approved):
Nice catch!

PR #208, opened 2021-05-07
Add an automation API landing page

joeduffy:
This is just an initial sketch. We'll make it better soon.


zchase (approved):
🚢

susanev:
i have a few comments on this..
- is there a reason to use automation for the URL instead of automation-api to match the heading?
- I think the scenarios and 3 callouts below them have too much content (and it makes the page overwhelming), we were trying to keep all of those to no more than 3 lines

PR #209, opened 2021-05-07
Size code elements relatively in headings

cnunciato:
Little thing I've been meaning to fix forever.

image

  👇👇 

image


susanev (approved):
thanks for fixing this! separately, @davidwrede im curious if you have thoughts on headings using code styles in them?

davidwrede:
@susanev From a visual standpoint, I think we should stay away from inline code blocks in a heading, but sometimes it's not a hard and fast rule.

PR #213, opened 2021-05-10
Relax the Hugo version restriction

cnunciato:
We currently impose a restriction on Hugo version, pinning to the same version as what's used in CI. But this is technically unnecessary (Hugo's built-in `minVersion` warns when the CLI is too far behind what's required by the modules in use in a given site), and given how often Hugo updates, and how hard it is to install a specific version on certain platforms, it makes life more difficult for contributors. So this change just relaxes that restriction and leaves it up to `minVersion` to alert when necessary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ clean:
 	./scripts/clean.sh
 
 .PHONY: ensure
-ensure: clean
+ensure:
 	./scripts/ensure.sh
 
 .PHONY: lint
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -3,7 +3,7 @@ clean:
 	./scripts/clean.sh
 
 .PHONY: ensure
-ensure: clean
```
cnunciato, line 6: Removing this because it's not necessary in CI, and there are times when you want to ensure but _not_ clean.

PR #215, opened 2021-05-11
Clean up tags

infin8x:
Fixes pulumi/docs#2532

Cleans up our tags to a smaller, easier-to-use set. I optimized for:

- Removing individual technologies/providers _unless_ we have substantial content for that technology (e.g. Kubernetes)
- Removing references to individual cloud resources
- Removing "obvious" tags like `pulumi` and `infrastructure`
- Removing customer names (and consolidating around the `guest-post` tag for guest posts)

I'd love feedback. Did I cut a tag you found useful? Are there even more tags we should cut?

image

### DIFF
--- a/themes/default/content/blog/amazon-ecr-public/index.md
+++ b/themes/default/content/blog/amazon-ecr-public/index.md
@@ -4,7 +4,7 @@ date: "2020-12-01"
 meta_desc: "Pulumi container images now available on Amazon ECR Public"
 meta_image: "pulumi-images-ecr.png"
 authors: ["paul-stack"]
-tags: ["aws", "containers", "ecr"]
+tags: ["aws", "docker-containers"]
 ---
 
 At re:Invent, the AWS team unveiled the new Amazon Elastic Container Registry Public (Amazon ECR Public), creating a new
--- a/themes/default/content/blog/amazon-ecr-public/index.md
+++ b/themes/default/content/blog/amazon-ecr-public/index.md
@@ -4,7 +4,7 @@ date: "2020-12-01"
 meta_desc: "Pulumi container images now available on Amazon ECR Public"
 meta_image: "pulumi-images-ecr.png"
 authors: ["paul-stack"]
-tags: ["aws", "docker-containers"]
+tags: ["aws", "containers"]
 ---
 
 At re:Invent, the AWS team unveiled the new Amazon Elastic Container Registry Public (Amazon ECR Public), creating a new
--- a/themes/default/content/blog/announcing-crossguard-preview/index.md
+++ b/themes/default/content/blog/announcing-crossguard-preview/index.md
@@ -4,7 +4,7 @@ date: 2019-12-02
 meta_desc: "Today we are announcing Pulumi CrossGuard, a Policy as Code solution that enforces custom infrastructure policies, is available for all users to preview."
 meta_image: crossguard.svg
 authors: ["erin-krengel"]
-tags: ["policy-as-code", "new-features"]
+tags: ["policy-as-code", "features"]
 ---
 
 Over the past few months, we have been hard at work on Pulumi CrossGuard, a Policy as Code solution. Using CrossGuard, you can express flexible business and security rules using code. CrossGuard enables organization administrators to enforce these policies across their organization or just on specific stacks. CrossGuard allows you to verify or enforce custom policies on changes before they are applied to your resources. CrossGuard is 100% open source and available to all users of Pulumi, including the Community Edition. Advanced organization-wide policy management features are available to Team Pro and Enterprise customers.
--- a/themes/default/content/blog/announcing-crossguard-preview/index.md
+++ b/themes/default/content/blog/announcing-crossguard-preview/index.md
@@ -4,7 +4,7 @@ date: 2019-12-02
 meta_desc: "Today we are announcing Pulumi CrossGuard, a Policy as Code solution that enforces custom infrastructure policies, is available for all users to preview."
 meta_image: crossguard.svg
 authors: ["erin-krengel"]
-tags: ["crossguard", "policy-as-code", "New-Features", "Pulumi-News", "Features"]
+tags: ["policy-as-code", "policy-as-code", "New-Features", "Pulumi-News", "Features"]
 ---
 
 Over the past few months, we have been hard at work on Pulumi CrossGuard, a Policy as Code solution. Using CrossGuard, you can express flexible business and security rules using code. CrossGuard enables organization administrators to enforce these policies across their organization or just on specific stacks. CrossGuard allows you to verify or enforce custom policies on changes before they are applied to your resources. CrossGuard is 100% open source and available to all users of Pulumi, including the Community Edition. Advanced organization-wide policy management features are available to Team Pro and Enterprise customers.
--- a/themes/default/content/blog/announcing-crossguard-preview/index.md
+++ b/themes/default/content/blog/announcing-crossguard-preview/index.md
@@ -4,7 +4,7 @@ date: 2019-12-02
 meta_desc: "Today we are announcing Pulumi CrossGuard, a Policy as Code solution that enforces custom infrastructure policies, is available for all users to preview."
 meta_image: crossguard.svg
 authors: ["erin-krengel"]
-tags: ["policy-as-code", "policy-as-code", "New-Features", "Pulumi-News", "Features"]
+tags: ["policy-as-code", "new-features"]
 ---
 
 Over the past few months, we have been hard at work on Pulumi CrossGuard, a Policy as Code solution. Using CrossGuard, you can express flexible business and security rules using code. CrossGuard enables organization administrators to enforce these policies across their organization or just on specific stacks. CrossGuard allows you to verify or enforce custom policies on changes before they are applied to your resources. CrossGuard is 100% open source and available to all users of Pulumi, including the Community Edition. Advanced organization-wide policy management features are available to Team Pro and Enterprise customers.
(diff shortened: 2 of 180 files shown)


themes/default/content/blog/amazon-ecr-public/index.md
```diff
@@ -4,7 +4,7 @@ date: "2020-12-01"
 meta_desc: "Pulumi container images now available on Amazon ECR Public"
 meta_image: "pulumi-images-ecr.png"
 authors: ["paul-stack"]
-tags: ["aws", "containers", "ecr"]
+tags: ["aws", "docker-containers"]
```
lukehoban, line 7: Not sure there's a need to rename `containers` to `Docker-containers`.

themes/default/content/blog/announcing-crossguard-preview/index.md
```diff
@@ -4,7 +4,7 @@ date: 2019-12-02
 meta_desc: "Today we are announcing Pulumi CrossGuard, a Policy as Code solution that enforces custom infrastructure policies, is available for all users to preview."
 meta_image: crossguard.svg
 authors: ["erin-krengel"]
-tags: ["crossguard", "policy-as-code", "New-Features", "Pulumi-News", "Features"]
+tags: ["policy-as-code", "new-features"]
```
lukehoban, line 7: Was it intentional to remove Pulumi-news here (even though the tag still exists)?

themes/default/content/blog/announcing-pulumi-0.15-kubernetes-cicd-openstack-and-more/index.md
```diff
@@ -3,7 +3,7 @@ title: "Announcing Pulumi 0.15"
 date: "2018-08-15"
 meta_desc: "Pulumi can now deploy and manage Kubernetes resources using the same familiar programming model supported for AWS, Azure, and Google Cloud Platform."
 authors: ["luke-hoban"]
-tags: ["Pulumi", "New-Features", "CI/CD"]
+tags: ["New-Features", "continuous-delivery"]
```
lukehoban, line 6: Are we lower-casing these by default now?

themes/default/content/blog/aws-lambda-efs/index.md
```diff
@@ -7,8 +7,6 @@ authors:
     - luke-hoban
 tags:
     - AWS
-    - Lambda
```
lukehoban, line 10: I think we still have "serverless" that would make sense here?

themes/default/content/blog/enforcing-different-kinds-of-policies-for-cloud-resources/index.md
```diff
@@ -6,7 +6,7 @@ meta_image: crossguard.png
 authors:
     - justin-vanpatten
 tags:
-    - crossguard
+    - policy-as-code
```
lukehoban, line 9: I think this duplicates the next line?

themes/default/content/blog/keep-your-secrets-secure-by-default/index.md
```diff
@@ -7,7 +7,6 @@ meta_image: secure_by_default.png
 authors:
     - alex-mullans
 tags:
-    - secrets
```
lukehoban, line 10: I'll admit that I recently used this tag to find all the posts we had on secrets - and there are quite a few. I'm not sure exactly how it fits with the principles - but I feel like this makes sense as a label (similar to policy-as-code).

lukehoban (commented):
I like it!  Definitely interested in @spara's thoughts as well.

themes/default/content/blog/announcing-crossguard-preview/index.md
susanev: should this just be features? since they wont stay "new" for very long, or maybe that's not a concern?

tusharshahrs:
The following blog post:    [Refactoring Pulumi Code with `aliases`](http://pulumi-hugo-origin-pr-215-f3e758e7.s3-website.us-west-2.amazonaws.com/blog/tag/testing/#refactoring-pulumi-code-with-aliases)

has the tag: `testing`.  Can we switch this to a new tag called:  `aliases`?  We have a resource option called `aliases` that it will match. 

Reason: 
This issue shows up on many customer calls and slack questions and this shows people how to resolve renaming.  No one thinks to look for the word `testing`.

themes/default/content/blog/announcing-crossguard-preview/index.md
infin8x: @lukehoban yes. I've reserved `pulumi-news` for company news (funding, SOC2, etc.) and focused `new-features` for... new features.

@susanev that's a good point. I usually view blogs as a "point-in-time" medium. Put another way, I could make a case for keeping `new-features` since, at the time the blog post was authored, the feature was, in fact, new. But, I like shorter tags in general, so I'll adjust to just `features`.

themes/default/content/blog/announcing-pulumi-0.15-kubernetes-cicd-openstack-and-more/index.md
infin8x: The blog system lowercases them by default, so it doesn't really matter what's in code. My find-and-replace defaults to lowercase as well.

themes/default/content/blog/aws-lambda-efs/index.md
infin8x: Quite right. Fixed.

themes/default/content/blog/amazon-ecr-public/index.md
infin8x: Fair. Fixed.

themes/default/content/blog/keep-your-secrets-secure-by-default/index.md
infin8x: I can see the case for secrets as a feature. Will bring it back.

susanev:
are we doing redirects for the tags ur removing?

spara (commented):
I think the list looks ok, but I would suggest removing the following tags because they a bit generic or not usefull:

- cloud-native (everything we do is pretty much cloud native)
- features
- guest-post
- pulumi-interns

cnunciato:
> are we doing redirects for the tags ur removing?

I personally wouldn't mind just letting these 404 and fall out of Google, since by removing them we're effectively saying we don't want them to be discovered anymore. If there are any we consider especially important and that we have good redirects for, we could redirect them, but in general I'd be okay just letting them go. (And FWIW, there isn't really a Hugo-based way to do these kinds of redirects, since tag pages aren't authored in the way regular pages are; we'd need to write custom redirect rules to do this.)

infin8x:
@spara thanks! 

> cloud-native (everything we do is pretty much cloud native)

Agreed, I almost cut this and wanted to see if anyone else suggested - will cut now.

> features

I'm torn here - I like the general banner for "blog posts about new features, which may or may not ever be blogged about enough to get their own tags" - like `enums`. I'd like to keep this unless you feel strongly.

> guest-post

I don't feel strongly here, but I did think it'd be useful to keep this so we could see the history of guest posts over time.

> pulumi-interns

I liked this as a hiring mechanism that we can point future interns to. I'd like to keep this unless you feel strongly.

infin8x:
@cnunciato when you have a minute, can you help me correctly noindex the tags pages? I found `tag.html` and also the `block_external_search_index` variable, but not sure if I can make them work together since the latter seems to only be applicable to Markdown frontmatter, not HTML.

cnunciato:
@infin8x We could probably add a robots.txt exclusion for them; I can push a commit that does that, either here or as a separate PR. 

But given we're trimming these back to be more relevant, do we want/need to exclude them from indexing? Just want to be sure I understand what the goal is.

infin8x:
Hmm, good point. If we had landing pages in the docs for most tags (esp. the features ones), I would be tempted to say yes, since I'd rather land someone on the primary docs page. But, since we don't, and now that tags are consolidated and a bit more useful, perhaps we can wait for now.

OK - I've convinced myself. Will merge this in, we can discuss `noindex`ing stuff later.

joeduffy:
TBH, I would have just ranked the tags, and presented only the "Top N." That way we're not finding ourselves in a situation down the road where someone added a tag here and there -- for instance, why keep "digitalocean" but not "mysql"? I guess because we expect more DigitalOcean content -- and furthermore, why not encourage as much metadata tagging as someone wants to add? It also doesn't encounter the redirects issue, would be less churn (smaller changelist), etc ....

infin8x:
@joeduffy as the tags are currently designed, having "as much metadata as someone wants to add" led to the massive (and, I would say, hard-to-use) list of tags we had. This tactical change gets us to an easy-to-scan set of tags that's Top N-ish. The top file changed in the PR is `BLOGGING.md`, which discusses when to add tags, what tags we have, etc.

spara (approved):

joeduffy:
Correct, it led to this situation because we didn’t have the very simple thing I suggested: limit the index to the top N tags. We solved the busy index problem by losing valuable metadata about the posts. If I want to run a campaign in the future for microservice related content, I now don’t have those tags. This change feels like a step backwards and a mistake.

joeduffy:
This is an example of what I had in mind, fwiw:

```diff
diff --git a/themes/default/layouts/partials/blog/sidebar.html b/themes/default/layouts/partials/blog/sidebar.html
index e0893e40..fb1bfbcb 100644
--- a/themes/default/layouts/partials/blog/sidebar.html
+++ b/themes/default/layouts/partials/blog/sidebar.html
@@ -45,12 +45,12 @@
         </div>
         <hr class="my-8">
         <ul class="list-none p-0 mb-4 inline-flex flex-wrap text-xs">
-            {{ range $tag, $_ := .Site.Taxonomies.tags }}
-                {{ if $tag }}
+            {{ range first 20 .Site.Taxonomies.tags.ByCount }}
+                {{ if .Name }}
                     <li>
-                        <a data-track="blog-tag-{{ $tag | urlize }}" class="tag tag-blog text-xs {{ if eq (lower $tag) (urlize ((lower $.Title))) }} active {{ end }}" href="{{ .Page.Permalink }}">
-                            {{ $tag }}
-                        </a>
+                        <a data-track="blog-tag-{{ $.Name | urlize }}" class="tag tag-blog text-xs {{ if eq (lower .Name) (urlize ((lower $.Title))) }} active {{ end }}" href="{{ .Page.Permalink }}">
+                            {{ .Name }}
+                        </a> ({{ .Count }})
                     </li>
                 {{ end }}
             {{ end }}
```

PR #222, opened 2021-05-12
Clean up the webinar/event listings and add upcoming events.

zchase:
Just as the title states I cleaned up the old event listings and then added new ones based on the Marketing calendar.

@jaxxstorm I used the descriptions from the doc you wrote but I know people are actively tweaking the descriptions. We really need to get the reg forms open so we can promote these workshops, so hopefully you are ok with shipping this as-is and updating as necessary.

### DIFF
--- a/themes/default/content/resources/aws-howdy-partner-twitch/index.md
+++ b/themes/default/content/resources/aws-howdy-partner-twitch/index.md
@@ -3,6 +3,9 @@
 title: "AWS - Howdy Partner"
 meta_desc: "Join Pulumi VP of Engineering Lee Zen as he shows off some of the newest features in Pulumi 3.0 and how those features make building on AWS easier than ever."
 
+aliases:
+    - /resources/aws-howdy-partner-twitch-2020-06-01
+
 # A featured webinar will display first in the list.
 featured: false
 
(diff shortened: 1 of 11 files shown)


themes/default/content/resources/aws-howdy-partner-twitch/index.md
```diff
@@ -1,13 +1,16 @@
 ---
 # Name of the webinar.
 title: "AWS - Howdy Partner"
-meta_desc: "Watch Pulumi VP of Engineering Lee Zen build serverless applications in real time, using both Python and TypeScript."
+meta_desc: "Join Pulumi VP of Engineering Lee Zen as he shows off some of the newest features in Pulumi 3.0 and how those features make building on AWS easier than ever."
```
cnunciato, line 4: Nice description! 💯

themes/default/content/resources/cncf-live-building-your-dream-cloud-native-platform/index.md
```diff
@@ -0,0 +1,75 @@
...
+    description: ""
+
+    # The webinar presenters
+    presenters:
+        - name: ""
+          role: ""
+
+    # A bullet point list containing what the user will learn during the webinar.
+    learn:
+        - ""
```
cnunciato, line 74: Should this be empty? May not matter, but figured I'd point it out just in case.

cnunciato (approved):
LGTM!

themes/default/content/resources/cncf-live-building-your-dream-cloud-native-platform/index.md
zchase: Ah yeah it is external so the `main` values don't need to filled in. I usually leave the section in there in case we end up with a recording and then I fill everything in when if we list the recording.

PR #229, opened 2021-05-13
add secret decrypted to audit log event list

### LINKED ISSUE #230: Update audit logs doc to include decrypted event
Update the audit logs docs to include the new secret decrypted event in the events list

sean1588:
Update audit log event list to include secret decrypted event.

link to page from preview below: http://pulumi-hugo-origin-pr-229-3882ffbd.s3-website.us-west-2.amazonaws.com/docs/intro/console/auditing/

fixes #230

### DIFF
--- a/themes/default/content/docs/intro/console/auditing.md
+++ b/themes/default/content/docs/intro/console/auditing.md
@@ -172,7 +172,7 @@ This is a list of the audit log events currently being recorded.
 | Policy Pack Deleted                      | indicates the deletion of a policy pack                                                                              |
 | Policy Pack Disabled                     | indicates the disabling of a policy pack                                                                             |
 | Policy Pack Enabled                      | indicates the enabling of a policy pack                                                                              |
-| Secret Decrypted                         | indicates the decryption of a secret value associated with the specified stack                                                                  |
+| Secret Decrypted                         | indicates the decryption of a secret value associated with a stack                                                                  |
 | Stack Collaborator Added                 | indicates the adding of a collaborator to a stack                                                               |
 | Stack Collaborator Permissions Changed   | indicates a change in permissions for a stack collaborator                                         |
 | Stack Collaborator Removed               | indicates the removal of a collaborator to a stack                                                            |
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/intro/console/auditing.md
```diff
@@ -172,6 +172,7 @@ This is a list of the audit log events currently being recorded.
 | Policy Pack Deleted                      | indicates the deletion of a policy pack                                                                              |
 | Policy Pack Disabled                     | indicates the disabling of a policy pack                                                                             |
 | Policy Pack Enabled                      | indicates the enabling of a policy pack                                                                              |
+| Secret Decrypted                         | indicates the decryption of a secret value associated with the specified stack                                                                  |
```
susanev, line 175: looks like other items in this table use a mix of `to a stack` and  `for a stack`, we should probably align with one of those instead of introducing a new format

susanev (approved):

susanev:
thx for the link in ur description!!

themes/default/content/docs/intro/console/auditing.md
infin8x: Not sure I follow what you mean here? The other items discuss adding/removing something to/from the stack. Since secrets are technically a separate entity, doesn't seem right to say "indicates the decryption of a secret value from a stack". Can you suggest a wording?

infin8x (approved):
LGTM

themes/default/content/docs/intro/console/auditing.md
davidwrede: We could do something like "indicates the decryption of a secret value associated with a stack"

davidwrede (approved):

themes/default/content/docs/intro/console/auditing.md
infin8x:
```suggestion
| Secret Decrypted                         | indicates the decryption of a secret value associated with a stack                                                                  |
```

I like that - @susanev WDYT?

Added it as a suggestion so Sean can make the change quickly if we're in agreement.

susanev: oh i was not trying to block, I approved when I left the comment. thx for the suggestion dave!

sean1588:
Thank you @infin8x for getting it merged. :smile:

Issue #234, opened 2021-05-14
Layout width not making it easy to read resource and method names


pierskarsenbarg:
Not sure if it's just Azure-Native giving long resource and method names, but [this page in the docs](https://www.pulumi.com/docs/reference/pkg/azure-native/web/) looks like this, even on a wide screen monitor:

![image](https://user-images.githubusercontent.com/109047/118290640-8f937380-b4ce-11eb-910b-5b1e33a69bf2.png)

And some of the resources have similar names

![image](https://user-images.githubusercontent.com/109047/118290706-a3d77080-b4ce-11eb-9769-a6741ef375a5.png)

Which makes it tricky to work out which one is the correct one

## Expected behavior

I should be able to see the full names on the page

susanev:
@cnunciato @davidwrede do either of you know why the multi-column layouts were chosen for these pages? do we believe folks browse these pages? or are they usually looking for a specific thing?

at a minimum, we absolutely need tooltips on hover whenever we cut text and append `...`

pierskarsenbarg:
You can mouse over and get something: 

![image](https://user-images.githubusercontent.com/109047/118291797-e2b9f600-b4cf-11eb-90ff-5f5ec1826430.png)

But it's not great

cnunciato:
@susanev People absolutely do browse these pages, yes. We chose multi-column mainly for information density, since the ~90% case is that resource names fall well within 1/3 of the page. We could offer a toggle to switch between list and column view, though, and even make it sticky if we wanted.

susanev:
chatted with @cnunciato we are going to switch to 2-column for now, which wont totally solve it but will make it better quickly. this will also open things up a bit and make it easier for folks to consume the information.

pierskarsenbarg:
Closing this issue because it's looking a lot better now

PR #235, opened 2021-05-14
change resource docs from 3-col to 2-col layout

susanev:
## overview

changing the docs to present data in a 2-col layout instead of 3-col layout which should make the content a bit easier for folks to consume

part of #234

you wont be able to use the link to view the pages, but I've included a few screenshots

### before
3-col

### after
2-col

### DIFF
--- a/themes/default/assets/sass/_api-nodejs.scss
+++ b/themes/default/assets/sass/_api-nodejs.scss
@@ -36,15 +36,18 @@ h3.api {
 ul.api {
     @apply list-none p-0 m-0 overflow-hidden;
 
-    columns: 2 330px;
+    columns: 1;
+
+    @screen xl {
+        columns: 2;
+    }
 
     li {
         @apply truncate mt-0;
 
         a {
-            @apply btn truncate bg-transparent inline-block p-1;
+            @apply btn truncate bg-transparent inline-block p-1 max-w-full;
 
-            max-width: 330px;
         }
 
         .symbol {
--- a/themes/default/assets/sass/_api-nodejs.scss
+++ b/themes/default/assets/sass/_api-nodejs.scss
@@ -36,7 +36,7 @@ h3.api {
 ul.api {
     @apply list-none p-0 m-0 overflow-hidden;
 
-    columns: 3 220px;
+    columns: 2 330px;
 
     li {
         @apply truncate mt-0;
@@ -44,7 +44,7 @@ ul.api {
         a {
             @apply btn truncate bg-transparent inline-block p-1;
 
-            max-width: 250px;
+            max-width: 330px;
         }
 
         .symbol {


pierskarsenbarg (approved):
LGTM. We're never going to catch all the resource names, but it looks like this is the vast majority of them 👍

themes/default/assets/sass/_api-nodejs.scss
```diff
@@ -36,15 +36,15 @@ h3.api {
 ul.api {
     @apply list-none p-0 m-0 overflow-hidden;
 
-    columns: 3 220px;
+    columns: 2 330px;
```
cnunciato, line 39: Rather than hard-code a width for these, we might as well just explicitly say, "it's one column by default, two at desktop width and up": 

```suggestion
    columns: 1;

    @screen xl {
        columns: 2;
    }
```

https://user-images.githubusercontent.com/274700/118325205-fc4f4380-b4b7-11eb-844b-116d7bd95254.mp4

themes/default/assets/sass/_api-nodejs.scss
```diff
@@ -36,15 +36,15 @@ h3.api {
...
 
     li {
         @apply truncate mt-0;
 
         a {
             @apply btn truncate bg-transparent inline-block p-1;
 
-            max-width: 250px;
+            max-width: 330px;
```
cnunciato, line 47:
```suggestion
```

themes/default/assets/sass/_api-nodejs.scss
```diff
@@ -36,15 +36,15 @@ h3.api {
...
     @apply list-none p-0 m-0 overflow-hidden;
 
-    columns: 3 220px;
+    columns: 2 330px;
 
     li {
         @apply truncate mt-0;
 
         a {
             @apply btn truncate bg-transparent inline-block p-1;
```
cnunciato, line 45: Similarly we should be able to just say "max-width of 100%" and get the same result:

```suggestion
            @apply btn truncate bg-transparent inline-block p-1 max-w-full;
```

cnunciato (changes requested):
Couple lil' suggestions to remove explicit widths (unless we need them for some reason).

themes/default/assets/sass/_api-nodejs.scss
susanev: oh okay, i was trying to align with what was already there

themes/default/assets/sass/_api-nodejs.scss
susanev: oh okay, i was trying to align with what was already there

cnunciato (approved):
✨

lukehoban:
The two preview links above seem to be missing all API reference docs - is that expected?

susanev:
@lukehoban yah we wont be able to preview it until it gets to the docs repo

PR #242, opened 2021-05-17
improve get started page in docs

susanev:
## overview
these are problems ive identified on the current page
- we throw a lot of content at users, and everything is pretty loud, we have cloud buttons, a video, and then cards as links to other pages
- the cloud buttons do not at all indicate what will happen or where you will go when you use them
- the cards at the bottom do not look interactive at all until you hover
- the text explaining the resources gets totally lost because of the presentation
- accessibility best practices state that links should look like links and buttons like buttons. generally speaking, ux best practices state that buttons should take an action on the page you are on, and links should send you to another page

fixes https://github.com/pulumi/pulumi-service/issues/6707

## proposed solution
- use links instead of custom buttons and cards that use clear link text to indicate where we will send users and why they may want to look at that thing

## feedback
let me know what yall think, super happy to adjust any of the content but would really really like to try a simplified version of this page

### before
get-started-before

### after
get-started-after

### DIFF
--- a/themes/default/content/docs/get-started/_index.md
+++ b/themes/default/content/docs/get-started/_index.md
@@ -32,6 +32,7 @@ We have guides for AWS, Azure, Google Cloud and Kubernetes using TypeScript, Jav
 - Learn about all the other clouds we support in the [Cloud Provider Docs]({{< relref "/docs/intro/cloud-providers" >}})
 
 ### Demo Video
+
 Not ready to get started yet? Watch a five-minute demo video instead.
 
 <div class="rounded-md shadow border border-gray-300 w-3/4" style="position: relative; padding-bottom: 40.25%; height: 0; overflow: hidden;">
--- a/themes/default/content/docs/get-started/_index.md
+++ b/themes/default/content/docs/get-started/_index.md
@@ -15,32 +15,24 @@ aliases:
 
 Pulumi is a modern infrastructure as code platform that allows you to use familiar programming languages and tools to build, deploy, and manage cloud infrastructure.
 
-Select one of the following options to deploy a simple application in your cloud using Pulumi:
+We have guides for AWS, Azure, Google Cloud and Kubernetes using TypeScript, JavaScipt, Python, Go or C#.
 
-<div class="tiles flex-wrap mt-4">
-    <div class="pb-4 md:pr-4 md:w-1/2">
-        <a class="tile p-8" href="{{< relref "/docs/get-started/aws" >}}">
-            <img class="h-10 mx-auto" src="/logos/tech/aws.svg" alt="AWS">
-        </a>
-    </div>
-    <div class="pb-4 md:w-1/2">
-        <a class="tile p-8" href="{{< relref "/docs/get-started/kubernetes" >}}">
-            <img class="h-10 mx-auto" src="/logos/tech/k8s.svg" alt="Kubernetes">
-        </a>
-    </div>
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/get-started/_index.md
```diff
@@ -15,32 +15,25 @@ aliases:
...
-            <img class="h-10 mx-auto" src="/logos/tech/azure.svg" alt="Azure">
-        </a>
-    </div>
-    <div class="pb-4 md:w-1/2">
-        <a class="tile p-8" href="{{< relref "/docs/get-started/gcp" >}}">
-            <img class="h-10 mx-auto" src="/logos/tech/gcp.svg" alt="Google Cloud">
-        </a>
-    </div>
-</div>
+## Deploy an Application Guides
```
cnunciato, line 20: This is technically fine, but not super engaging, as it's expressed in terms of nouns ("Guides", "AWS Guide") rather than verbs ("Deploy an application", "Deploy to AWS"). As a visitor who's probably just clicked a button with an imperative verb on it ("get started") on a page whose title is also "Get Started", I might be more motivated to keep moving forward if the words kept telling me what I'll _do_ by following one of these guides, rather than just listing them. The navigation buttons and page titles of each guide follow this pattern as well -- install, configure, create, review, etc.

```suggestion
## Deploy an Application
```

themes/default/content/docs/get-started/_index.md
```diff
@@ -15,32 +15,25 @@ aliases:
...
+
+## More Get Started Resources
 
-Or, watch how to do it in this video walkthrough:
+- Learn about [Pulumi Architecture & Concepts]({{< relref "/docs/intro/concepts" >}})
+- Learn how to adopt Pulumi in the [Adopting Pulumi Guide]({{< relref "/docs/guides/adopting" >}})
+- Sign up for one of our [Live Workshops]({{< relref "/resources/introduction-to-pulumi" >}})
+- Learn about all the other clouds we support in the [Cloud Provider Docs]({{< relref "/docs/intro/cloud-providers" >}})
+
+### Demo Video
```
cnunciato, line 34: I'd personally keep this where it is now, in terms of page hierarchy -- in other words, as the second most important thing on the page to potential new users, ahead of the deeper dives into architecture & concepts, adopting, etc. Unless we have signal we _shouldn't_ do this, it'd make more sense to me to assume that someone new(ish) to this page is looking mainly for high-level things -- a hands-on intro or a demo of one -- and secondarily for lower-level ones like how it works, how to adopt it, and so on.

cnunciato (commented):
Love the reduction of visual noise here. ❤️  A few comments to consider in terms of language and info hierarchy to think about.

themes/default/content/docs/get-started/_index.md
susanev: i was struggling with this, because i think the thumbnail is heavy, so that's why I placed it lower. i was also considering only using a link, then it could def be second. ill think about it some more.

cnunciato: Yeah. Another thing that gets lost here that I didn't notice at first is that the provider logos actually are pretty strong, visually -- it'd be great to be able to keep those if we can.

susanev:
as of now, the plan is to do an a/b test with these updates after we are finished with the current a/b test.
adding do-not-merge label for now

susanev:
ill come back to this at some point in the future

PR #243, opened 2021-05-17
remove cloud buttons from docs landing page

susanev:
## overview

in https://github.com/pulumi/pulumi-hugo/pull/190#issuecomment-833925032 it was mentioned that we should remove the "cloud buttons" from the docs landing page, this pr does that.

### before
docs-landing-before

### after
docs-landing-after

### DIFF
--- a/themes/default/content/docs/_index.md
+++ b/themes/default/content/docs/_index.md
@@ -18,23 +18,6 @@ Pulumi is an <a href="https://github.com/pulumi/pulumi" target="_blank">open sou
     <a class="btn btn-lg mx-1 my-1" href="{{< relref "/docs/get-started" >}}">Get Started</a>
 </div>
 
-<div class="bg-gray-100 rounded max-w-6xl my-4 px-4 py-2">
-    <div class="md:flex justify-between items-center">
-        <a class="block rounded hover:bg-gray-200 transition-all my-2 py-4 text-center px-6" href="{{< relref "/docs/get-started/aws" >}}">
-            <img class="inline-block h-8 w-auto -mb-2" src="/logos/tech/aws.svg">
-        </a>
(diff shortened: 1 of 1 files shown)


susanev:
honestly, id like to remove the 3-column layout too, anyone have opinions about that?

cnunciato (approved):
💯 🙌

lukehoban:
This feels like it net reduces now "nice" the page looks.  I'd be inclined to take a step back and think about what a nicer version of this page actually looks like - instead of just removing these.  Frankly having it be clear on this page that Pulumi is a tool for working with AWS, Azure, GCP and Kubernetes feels like a very good zeroth-order takeaway, which is lost if these are removed.

cnunciato:
> id like to remove the 3-column layout too, anyone have opinions about that?

These are admittedly text heavy, but there's also good introductory-level info here that we probably needn't just throw out. So if there were a way to trim things back a bit, that might be nice.

cnunciato:
>  Frankly having it be clear on this page that Pulumi is a tool for working with AWS, Azure, GCP and Kubernetes feels like a very good zeroth-order takeaway

I do agree this would be nice -- and if there were some way to indicate support for more providers visually, even better. I interpreted this PR as being aimed at removing something it's not super clear how to interact with, but I agree there's some useful signal there that'd be lost with just deleting it.

lukehoban:
> I interpreted this PR as being aimed at removing something it's not super clear how to interact with

I suppose my perspective is that users either:
1. See them as helpful zero-th order images that help them understand what Pulumi is, but don't realize they can click them - still a net positive they are there.
2. Know/guess they can click them, and have success getting started with Pulumi for their cloud.

If we can do something to move more folks from (1) to (2) - great.  But it doesn't seem that removing these entirely actually helps anything - at least until we do a bigger redesign of this page.

susanev:
closing this based on an [internal slack convo](https://pulumi.slack.com/archives/C85BS3LJZ/p1621279516042200)
this page does need some work in general though, so ill attempt to revisit in the future

PR #244, opened 2021-05-17
Migrate Azure ACI example to be azure-native

stack72:
Fixes: #5187

Uses: https://github.com/pulumi/examples/blob/master/azure-ts-aci/index.ts


mikhailshilkov (approved):
LGTM

PR #250, opened 2021-05-19
Fix C# unit testing snippet with mocks

mikhailshilkov:
The current example is invalid C#


t0yv0:
I take it we do not compile these examples? When we have time sounds like we could introduce a check to compile these https://github.com/pulumi/pulumi-hugo/issues/251

t0yv0 (approved):

PR #252, opened 2021-05-19
Specify we should use the project ID with GCP

cnunciato:
This just specifies the GCP project field to use when configuring `gcloud`.


davidwrede (approved):
Nice fix!

PR #254, opened 2021-05-19
Fix rendering issue in the GCP guide

cnunciato:
The missing newline after the `choosable` shortcode is preventing the markdown from being rendered properly on these pages. This should fix it up.


davidwrede (approved):

chrsmith:
> The missing newline after the choosable shortcode is preventing the markdown from being rendered properly on these pages. This should fix it up.

Not knowing much about Hugo shortcodes, I assume there isn't a way we can make the shortcode itself less dependent on that newline being available immediately afterwards? Or is it something like it looks for the next `<p>` element, which is only created if there is a blank line afterwards?

PR #255, opened 2021-05-19
Put output declarations on separate lines

cnunciato:
Our `csharp` templates [give these attributes lines of their own](https://github.com/pulumi/templates/blob/master/gcp-csharp/MyStack.cs#L15-L16), and this seems to be the convention as well. This just adjusts our guide pages to be consistent.


davidwrede (approved):
Nice catch!

PR #256, opened 2021-05-19
Update App Runner example code

leezen:


cnunciato (approved):
It'd be nice if we could create an example in pulumi/examples for this as well, just to have something finished to refer to. I had to stitch things together after emptying a templated project file and then guess a bit as to the right versions to import, which made this bumpier than it might've been otherwise; someone less familiar with Pulumi would probably stumble through this and likely get frustrated.

But it does work! So 👍 .

cnunciato:
Actually, this may not be right after all -- got an error just after saying `update` after a successful preview: 

image

themes/default/content/blog/deploy-applications-with-aws-app-runner/index.md
```diff
@@ -107,42 +106,58 @@ Where index.html is simply:
...
+}, nil)
+if err != nil {
+    return err
+}
+
+// Build and push our image into the ECR repo.
+dockerImage, err := docker.NewImage(ctx, "sampleapp", &docker.ImageArgs{
+    ImageName: repo.RepositoryUrl,
+    Build: docker.DockerBuildArgs{
+        Context: pulumi.String("./app"),
```
cnunciato, line 120: We should mention in the post that the Dockerfile and index.html need to live in this folder. Otherwise, the update fails.

leezen: Ah, thanks for pointing that out. Will update the text to make this clear.

cnunciato (approved):

PR #263, opened 2021-05-21
Upgrade our CSS toolchain

### LINKED ISSUE #83: upgrade tailwind

cnunciato:
Upgrades TailwindCSS, PostCSS, PurgeCSS, Autoprefixer, and CSSNano to their latest versions, and makes a few minor adjustments to make everything continue to lay out right.

https://tailwindcss.com/docs/upgrading-to-v2

Fixes #83.

### DIFF
--- a/themes/default/assets/config/postcss.config.js
+++ b/themes/default/assets/config/postcss.config.js
@@ -4,72 +4,66 @@
  * at the end to do more transformations.
  */
 
-// Note whether we're in "production" mode, so we can avoid doing post-processing and
-// optimization during development.
-const isProduction = process.env.NODE_ENV === "production";
+const plugins = [
 
-module.exports = {
-    plugins: [
-        // TailwindCSS
-        require("tailwindcss")("./assets/config/tailwind.config.js"),
+    // TailwindCSS
+    require("tailwindcss")("./assets/config/tailwind.config.js"),
 
-        // Apply vendor prefixes for CSS features that aren't
-        // fully supported yet.
-        isProduction ? require("autoprefixer")({
-            overrideBrowserslist: [
-                "last 2 versions"
-            ]
-        }) : null,
+    // Apply vendor prefixes for CSS features that aren't
+    // fully supported yet.
+    require("autoprefixer")({
+        overrideBrowserslist: [
+            "last 2 versions"
+        ]
+    }),
 
-        // Remove whitespace, comments and other safe things.
-        // https://cssnano.co/guides/getting-started/
-        isProduction ? require("cssnano")({
-            preset: [
-                "default",
-                {
-                    discardComments: {
-                        removeAll: true,
-                    },
-                },
-            ],
-        }) : null,
+    // Remove whitespace, comments and other safe things.
+    // https://cssnano.co/guides/getting-started/
+    //
+    // Commented temporarily due to unexpected removal of certain CSS variables required
+    // by the current design. See https://github.com/pulumi/pulumi-hugo/pull/263#issuecomment-849948317
+    // for additional context.
+    // require("cssnano")({
+    //     preset: [
+    //         "default",
+    //         {
+    //             discardComments: {
+    //                 removeAll: true,
+    //             },
+    //         },
+    //     ],
+    // }),
 
-        // Use PurgeCSS to remove unused classes for better page speed.
-        // Docs: https://purgecss.com/plugins/postcss.html
-        isProduction ? require("@fullhuman/postcss-purgecss")({
-            // Specify the paths to all of the template files in your project
-            content: [
-                // All layout files.
-                "./layouts/**/*.html",
+    // Use PurgeCSS to remove unused classes for better page speed.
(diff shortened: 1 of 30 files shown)


themes/default/assets/config/tailwind.config.js
```diff
@@ -20,9 +20,20 @@ const brand = {
 
 const white = defaultTheme.colors.white;
 const black = defaultTheme.colors.black;
-const red = defaultTheme.colors.red;
 const transparent = defaultTheme.colors.transparent;
 
+const red = {
```
cnunciato, line 25: Up to now, we've been using Tailwind's v1 defaults for `red`, which changed in v2. These are the actual v1 values. https://tailwindcss.com/docs/upgrading-to-v2#configure-your-color-palette-explicitly

themes/default/assets/config/tailwind.config.js
```diff
@@ -65,7 +76,7 @@ const fuchsia = {
     300: "#e5b7ce",
     400: "#d794b6",
     500: "#ca709d",
-    600: brand.fuschia,
+    600: brand.fuchsia,
```
cnunciato, line 79: 🤦

themes/default/assets/config/tailwind.config.js
```diff
@@ -165,22 +177,16 @@ module.exports = {
...
-            '50': '50vh',
-            '75': '75vh',
-            '100': '100vh',
-        }
+            "25": "25vh",
+            "50": "50vh",
+            "75": "75vh",
+            "100": "100vh",
+        },
+        screens: {
```
cnunciato, line 185: v2 also adds a `2xl` container width, which messes with our overall layout. Declaring these explicitly (https://tailwindcss.com/docs/upgrading-to-v2#configure-your-breakpoints-explicitly) gives us the v1 settings, and then in the Docs, where we do want a wider layout, we apply that using the v2 breakpoint. (Also in this PR.)

themes/default/assets/sass/_container.scss
```diff
@@ -0,0 +1,19 @@
+/* purgecss start ignore */
```
cnunciato, line 1: After a half-hour of trying to make this work in postcss.config.js, I gave up. This is valid, and works, so I decided to go with it.

themes/default/assets/sass/_header.scss
```diff
@@ -86,12 +83,24 @@
         @apply hidden;
 
         @screen lg {
-            @apply flex list-none justify-center;
-            width: calc((100% / 7) * 5);
+            @apply flex items-center flex-grow justify-evenly mx-8;
```
cnunciato, line 86: The changes in this file are just to fine-tune the main nav on desktop and mobile.

themes/default/layouts/partials/header.html
```diff
@@ -22,77 +22,72 @@
...
                     </li>
-
                     <li class="header-nav-item-drop-down">
                         <span class="dropdown-title">Learn</span>
                         <span class="dropdown-menu-caret"></span>
-                        <div class="header-nav-drop-down-menu learn-menu">
+                        <div class="header-nav-drop-down-menu">
                             <ul>
-                                <a href="{{ relref . "/blog" }}">
-                                    <li class="header-dropdown-item mb-3">
```
cnunciato, line 77: Noticed this was invalid (`ul > a > li` rather than `ul > li > a`) so fixed it up.

themes/default/assets/config/postcss.config.js
```diff
@@ -4,95 +4,92 @@
  * at the end to do more transformations.
  */
 
-// Note whether we're in "production" mode, so we can avoid doing post-processing and
-// optimization during development.
-const isProduction = process.env.NODE_ENV === "production";
```
cnunciato, line 9: Minification/purge only takes a few seconds in pulumi-hugo now, so this enables it in development as well.

zchase: 🎉 woohoo!

zchase (changes requested):
Overall looks great and thank you for doing this!! One thing I noticed that should be fixed up before we ship is that we have seem to lost the hover effect on our primary buttons. Looking through the changes it wasn't clear to me why this happen, as it didn't seem like you changed anything in that area.

cnunciato:
Took a bit to track down what's happening here, but it looks like CSSNano is stripping out the variable references that Tailwind 2 uses for things like background and border opacity, which we use in these gradient buttons. 

On the website today, the background value references the CSS variable `--bg-opacity`:

image

But with CSSNano enabled -- which we actually aren't using on the website right now (it was erroneously removed with the rebrand; this PR restores it) -- it's missing:

image

cnunciato:
@zchase I think I'm going to disable CSSNano for this PR and come back to this as a separate issue. Doing this only results in a 10K difference in file size and this seems reasonable given the overall gains we get from re-enabling PurgeCSS, etc. Will definitely make sure it's addressed, though.

zchase (approved):
LGTM 🚢

PR #266, opened 2021-05-21
Instructions on local jaeger

t0yv0:
Currently Jaeger has the advantage of correlating traces from all Pulumi sub-processes. Linked pulumi/pulumi#7109

### DIFF
--- a/themes/default/content/docs/troubleshooting/_index.md
+++ b/themes/default/content/docs/troubleshooting/_index.md
@@ -93,7 +93,6 @@ $ pulumi up --tracing http://localhost:9411/api/v1/spans
 
 To view a trace locally navigate to the [Jaeger UI](http://localhost:16686/search).
 
-
 ## Common Problems
 
 This section covers a few problems that can arise when working with Pulumi.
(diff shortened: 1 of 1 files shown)


t0yv0:
@viveklak following up on our discussion, here is a method that works today to explore correlated sub-process traces. I'll await review here on whether we want this exposed in public docs, if not, this will move to internal wiki until we can fix the default appdash-based tracing to work equally well.

pgavlin:
I do think that we should add this

I also think that it would probably be worth having the CLI serve a collection endpoint and plumb that endpoint through to its subprocesses s.t. we can collect the same data when tracing to a file.

pgavlin (approved):

viveklak (approved):

viveklak:
Yeah no concerns merging this. Also agree with Pat's propoosal to add a collection endpoint to the CLI. Could you cut a ticket to track that?

t0yv0:
It think it's https://github.com/pulumi/pulumi/issues/7109 - I probably can take care of this shortly.

PR #267, opened 2021-05-22
Fix headings on legal pages

cnunciato:
They're currently following the styling of marketing pages, which makes their headings look hugs. This should fix 'em up.


zchase (approved):
LGTM, thanks for fixing this up! I vaguely remember say I would do this during the re-design work, but must have lost track of it. 🎉

PR #269, opened 2021-05-24
Update modify-program.md

KadoBOT:
Change `bucketObject` to `_`. Since it's not used elsewhere, compilation will fail. Since `err` was already defined, `:=` has to be changed to `=`


stack72 (approved):
thanks so much for this! You are indeed correct

stack72:
It looks like it may need an update on themes/default/content/docs/get-started/gcp/deploy-changes.md as well if you'd be kind enough to update for that?

stack72:
This looks great! Thanks so much - I've got this running locally

PR #273, opened 2021-05-25
Add Aurelien Requiem to the team

aureq:
Hummmm, maybe worth checking the accented `é`


cnunciato (commented):
@aureq Do you have a photo you can share? If you have a JPG that's at least 200x200, you can put it in `themes/default/static/images/team` and it'll show up on the About page. (Feel free to DM me in Slack if you need anything cropped or adjusted or anything.)

themes/default/data/team/team/aurelien-requiem.toml
```diff
@@ -0,0 +1,11 @@
+id = "aureq"
```
cnunciato, line 1: We use this value to link to the image, so if you change it to `aurelien-requiem`, things will "just work". 😄 

```suggestion
id = "aurelien-requiem"
```

cnunciato (commented):
One little thing, then I think we can merge!

cnunciato (approved):

PR #274, opened 2021-05-25
[Blog] Refactoring IaC

spara:


aaronkao:
Typo on Pulumi 

> An advantage of Pulumu is that moving a resource across file boundaries does not recreate the resource because a resource’s identity is based on its object ID, irrespective of its file location.

spara:
> Typo on Pulumi
> 
> > An advantage of Pulumu is that moving a resource across file boundaries does not recreate the resource because a resource’s identity is based on its object ID, irrespective of its file location.

fixed

pierskarsenbarg (approved):
LGTM

PR #282, opened 2021-05-29
Lazy-load sentry

cnunciato:
We currently load Sentry in the default way, which blocks page rendering until the Sentry SDK finishes loading. This change uses a different approach that loads the Sentry SDK later, as an attempt to improve page performance. 

Fixes https://github.com/pulumi/docs/issues/5959. (🤞)

### DIFF
--- a/themes/default/layouts/partials/head.html
+++ b/themes/default/layouts/partials/head.html
@@ -103,14 +103,16 @@
     <!-- Our CSS bundle. -->
     <link rel="stylesheet" href="{{ getenv "REL_CSS_BUNDLE" }}">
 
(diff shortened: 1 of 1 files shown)


ringods:
@cnunciato I experienced slow loads on some pages myself as well. I checked the preview site and it definitely feels much snappier.

For instance, on this page, I had performance problems before, but not anymore:
http://pulumi-hugo-origin-pr-282-69b224b0.s3-website.us-west-2.amazonaws.com/docs/intro/concepts/resources/

jkisk (approved):
Looks good, double checked docs here: https://docs.sentry.io/platforms/javascript/install/lazy-load-sentry/

PR #285, opened 2021-06-01
Add Pulumi Package Schema documentation

lukehoban:
Fixes pulumi/docs#5997.

### DIFF
--- a/themes/default/content/docs/guides/pulumi-packages/schema.md
+++ b/themes/default/content/docs/guides/pulumi-packages/schema.md
@@ -91,6 +91,7 @@ An example of the Pulumi Package Schema is below.  This schema describes a packa
 ```
 
 Complete schema examples that include a much wider range of schema configuration styles are availble in these existing packages:
+
 * [AWS](https://github.com/pulumi/pulumi-aws/blob/master/provider/cmd/pulumi-resource-aws/schema.json) - Bridged Provider Package
 * [Azure Native](https://github.com/pulumi/pulumi-azure-native/blob/master/provider/cmd/pulumi-resource-azure-native/schema.json) - Native Pulumi Provider Package
 * [EKS](https://github.com/pulumi/pulumi-eks/blob/master/provider/cmd/pulumi-resource-eks/schema.json) - Component Package
--- a/themes/default/content/docs/guides/pulumi-packages/schema.md
+++ b/themes/default/content/docs/guides/pulumi-packages/schema.md
@@ -12,36 +12,52 @@ Pulumi Packages are described by a package schema, which is used to drive code g
 
 ## Example
 
-An example of the Pulumi Package Schema is below.  This schema describes a package named `xyz`, with a single resource `StaticPage` which is a [component]({{< relref "/docs/intro/concepts/resources#components" >}}), and which has a required input property `indexContent` of type `string`  and a required (always populated) output property `websiteUrl` of type `string`.  The schema supports generation of SDKs for `csharp`, `go`, `nodejs` and `python`, with metadata cconfiguring each of the generated SDKs in the corresponding sections.
+An example of the Pulumi Package Schema is below.  This schema describes a package named `apigateway`, with a single resource `RestAPI` which is a [component]({{< relref "/docs/intro/concepts/resources#components" >}}), and which has a required input property `routes` and required (always populated) output properties `url` of type `string` and `api` which references the external type of the AWS API Gateway [`RestAPI`](https://www.pulumi.com/docs/reference/pkg/aws/apigateway/restapi/) resource.  The type of the `routes` input is a custom type named `EventHandlerRoute` defined in the schema, which is an object type with `path`,`method` and `function` properties. The schema supports generation of SDKs for `csharp`, `go`, `nodejs` and `python`, with metadata configuring each of the generated SDKs in the corresponding sections.
 
 ```json
 {
-    "name": "xyz",
+    "name": "apigateway",
+    "types": {
+        "apigateway:index:EventHandlerRoute": {
+            "type": "object",
+            "properties": {
+                "path": {
+                    "type": "string"
+                },
+                "method": {
+                    "type": "string"
+                },
+                "function": {
+                    "$ref": "/aws/v3.30.0/schema.json#/resources/aws:lambda%2Ffunction:Function"
+                }
+            }
+        }
+    },
     "resources": {
-        "xyz:index:StaticPage": {
+        "apigateway:index:RestAPI": {
             "isComponent": true,
             "inputProperties": {
-                "indexContent": {
-                    "type": "string",
-                    "description": "The HTML content for index.html."
+                "routes": {
+                    "type": "array",
+                    "items": {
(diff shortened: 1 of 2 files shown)


lukehoban:
@cnunciato A couple of the tables are overflowing the main section - do you know of a way to tell Hugo to force them not to overflow (perhaps requiring wordbreaks)?

themes/default/content/docs/guides/pulumi-packages/schema.md
```diff
@@ -0,0 +1,321 @@
...
+    userguides:
+        parent: pulumi-packages
+        weight: 2
+---
+
+Pulumi Packages are described by a package schema, which is used to drive code generation for SDKs in each supported Pulumi language, as well as generation of langauge-agnostic package documentation.  This schema can be manually authored (for component packages) or generated from some other source (such as a cloud provider's API specifications for a native Pulumi resource provider).  Packages can expose resources and functions, define types used by these resources and functions, and provide packaging metadata for language-specific SDKs.
+
+## Example
+
+An example of the Pulumi Package Schema is below.  This schema describes a package named `xyz`, with a single resource `StaticPage` which is a [component]({{< relref "/docs/intro/concepts/resources#components" >}}), and which has a required input property `indexContent` of type `string`  and a required (always populated) output property `websiteUrl` of type `string`.  The schema supports generation of SDKs for `csharp`, `go`, `nodejs` and `python`, with metadata cconfiguring each of the generated SDKs in the corresponding sections.
```
justinvp, line 15:
```suggestion
An example of the Pulumi Package Schema is below.  This schema describes a package named `xyz`, with a single resource `StaticPage` which is a [component]({{< relref "/docs/intro/concepts/resources#components" >}}), and which has a required input property `indexContent` of type `string`  and a required (always populated) output property `websiteUrl` of type `string`.  The schema supports generation of SDKs for `csharp`, `go`, `nodejs` and `python`, with metadata configuring each of the generated SDKs in the corresponding sections.
```

themes/default/content/docs/guides/pulumi-packages/schema.md
```diff
@@ -0,0 +1,321 @@
...
+| `packageName`          | `string`      | No       | PackageName is an override for the name of the generated python package.                                                 |
+| `requires`             | `map[string]` | No       | Description for the NPM package.                                                                                         |
+| `readme`               | `string`      | No       | Readme contains the text for the package's README.md files.                                                              |
+| `moduleNameOverrides`  | `map[string]` | No       | Optional overrides for Pulumi module names (e.g. `{ "flowcontrol.apiserver.k8s.io/v1alpha1": "flowcontrol/v1alpha1" }`). |
+| `compatibility`        | `string`      | No       | Toggle compatibility mode for a specified target.                                                                        |
+| `usesIOClasses`        | `boolean`     | No       | Deprecated: This bool is no longer needed since all providers now use input/output classes.                              |
```
justinvp, line 269: I don't think we need to document this. This is deadcode that needs to be deleted from the implementation.

themes/default/content/docs/guides/pulumi-packages/schema.md
```diff
@@ -0,0 +1,321 @@
...
+| `emitPulumiPluginFile` | `boolean`     | No       | Indicates whether the pulumiplugin.json file should be generated.                                                        |
```
justinvp, line 270: I'm tempted to not document this.

I'd originally added this option so providers could opt-in to generating the file because it requires an additional change to each provider's `Makefile` to replace the `${PLUGIN_VERSION}` string in the file with the actual version. We haven't opted-in any of our providers yet.

However, .NET and Go will soon be generating this file too (I'm prepping the PR now), and I'm planning to just go ahead and generate it by default (without an option) for those languages, and changing Python to do the same along with deleting the option.

themes/default/content/docs/guides/pulumi-packages/schema.md
```diff
@@ -0,0 +1,321 @@
...
+
+### PropertyLanguage
+
+Language-specific info for a property.
+
+For `python`:
+
+| Property  | Type      | Required | Description                                          |
+|-----------|-----------|----------|------------------------------------------------------|
+| `mapCase` | `boolean` | No       | Whether to use case mapping tables for the property. |
```
justinvp, line 299: This is dead code that I've been meaning to clean up. I don't think we need to document it.

justinvp (approved):
LGTM

As a follow-up change, it would be useful to show some examples of how to specify more complex types like arrays, maps, and enums, and how to reference resources/types (internal and external).

I wonder if it'd be worth linking to some of our providers' schemas, as examples, like:

- [AWS](https://github.com/pulumi/pulumi-aws/blob/master/provider/cmd/pulumi-resource-aws/schema.json)
- [EKS](https://github.com/pulumi/pulumi-eks/blob/master/provider/cmd/pulumi-resource-eks/schema.json)
- etc.

cnunciato:
@lukehoban not at present, but even so, forcing the text to wrap would still make some of these pretty hard to read.

Since these are all property lists, though, what about using the same lists we use in the resource docs? I've been meaning to make it easier to write these lists by hand, so I just threw together a couple of shortcodes as a POC that would probably work. Something like:

```
{{% property-list %}}

{{% property-list-item name="name" type="string" required="yes" %}}
Name is the unqualified name of the package (e.g. "aws", "azure", "gcp", "kubernetes", "random").
{{% /property-list-item %}}

{{% property-list-item name="meta" type="[`Metadata`](#metadata)" required="no" %}}
Meta contains information for the importer about this package.
{{% /property-list-item %}}

{{% property-list-item name="packageImportAliases" type="map[string]" required="No" %}}
Map from package name -> package alias (e.g. `{ "github.com/pulumi/pulumi-kubernetes/sdk/go/kubernetes/flowcontrol/v1alpha1": "flowcontrolv1alpha1" }`).
{{% /property-list-item %}}

{{% /property-list %}}
```

... would get you a list that that looked like 

image
  
What do you think? I could push a commit to add the shortcodes, or maybe a separate PR first.

lukehoban:
> As a follow-up change, it would be useful to show some examples of how to specify more complex types like arrays, maps, and enums, and how to reference resources/types (internal and external).

Good suggestion - I replaced the example in the intro with a slightly more "real" one which uses most of these to some limited degree.  It might still be nice to have examples that go into more detail in each individual section - but I think this is good for now.

> @lukehoban not at present, but even so, forcing the text to wrap would still make some of these pretty hard to read.

Got it.  I went ahead and just manually inserted spaces in the two places that were problematic, and that avoids the overflow, and is sufficiently readable for now.

> Since these are all property lists, though, what about using the same lists we use in the resource docs?

Love it.  I don't think I'll block on this - but if we do add shortcodes to reuse this UX, we could follow-up by migrating this over.  Note though that this content is manually maintained, so not sure how "usable" the shortcodes will be for manually maintained content at this scale (vs. a table which is relatively simple).  If we can make this (nearly) as easy as managing a table (or can eventually autogenerate this content from the Go struct types it was manually ported from), then this would definitely be the best way to go.

PR #291, opened 2021-06-04
Add defaultTags option

frezbo:


themes/default/content/docs/intro/cloud-providers/aws/_index.md
```diff
@@ -96,6 +96,8 @@ The AWS provider accepts the following configuration settings.  These can be pro
   * `transitiveTagKeys`: (Optional) Set of assume role session tag keys to pass to any subsequent sessions.
 * `dynamodbEndpoint`: (Optional) Use this to override the default endpoint URL constructed from the `region`. It’s typically used to connect to dynamodb-local.
 * `forbiddenAccountIds`: (Optional) List of forbidden AWS account IDs to prevent you from mistakenly using the wrong one (and potentially end up destroying a live environment). Conflicts with `allowedAccountIds`.
+* `defaultTags`: (Optional) A JSON block with resource tag settings to apply across all resources handled by this provider. The `defaultTags` settings has one nested property. Additional tags can be added/overridden at a per resource level.
```
leezen, line 99: Thanks for the suggestion to add this to the provider configuration page. I don't think this should read as a "JSON block" (though it's consistent w/ `ignoreTags` below, which we should also fix) since the type of this argument depends on the language SDK being utilized.

frezbo: @leezen any suggestions on how this should be worded would be helpful, thanks. I'm not sure I completely understand the reference to the language SDK. I set this option as:

`pulumi config set aws:defaultTags {"tags": {"key": "value"}`

leezen: @frezbo You know what, I was thinking in a totally different context around provider config from within code and not outside of it. The change is good as-is :)

leezen (approved):

PR #292, opened 2021-06-04
Update GitLab go provider version to v4

frezbo:


stack72 (approved):
Thanks so much for this @frezbo!

PR #299, opened 2021-06-07
Remove query params

cnunciato:
These are no longer needed. Related to https://github.com/pulumi/pulumi-service/issues/6676.


zchase (approved):
LGTM. Thank you for doing this!!

PR #306, opened 2021-06-09
Adjust docs & blog layout for tablets

cnunciato:
This uses the mobile layout up to iPad width, which makes browsing the docs and blog a lot nicer on a tablet. (This used to be fine before the redesign, so I suspect between the redesign itself and the post-redesign upgrade to Tailwind 2, something must've shifted.)

Before: 

image

After:

image

### DIFF
--- a/themes/default/layouts/blog/list.html
+++ b/themes/default/layouts/blog/list.html
@@ -1,12 +1,12 @@
 {{ define "main" }}
     <div class="container mx-auto px-4 pb-8 mt-8">
-        <div class="md:flex">
+        <div class="lg:flex">
 
-            <div class="md:w-3/12 pr-8">
+            <div class="lg:w-3/12 pr-8">
                 {{ partial "blog/sidebar.html" . }}
             </div>
 
-            <div class="md:w-7/12">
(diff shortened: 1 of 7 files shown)


susanev:
i dont think we should use the hamburger icon again, since we already have that for the top nav. let me think about this a bit and figure out if another icon can work here.

cnunciato:
Ok. This is the same one we've been using for this for a few years now -- can we merge this and come back to fix the icon in a follow-up to unbreak? (Unless you think you'll have something soon.)

susanev:
if i dont come up with something today you can merge

cnunciato:
One of y'all mind giving this a 👍 so I can merge, assuming no other issues?

susanev:
oh i was hoping to switch to a label instead of an icon, its still on my list for today to add the details. do you need to merge it now, I suppose I can open a separate issue if so.

cnunciato:
That'd be fine to do as a quick follow-up, yeah, considering this PR doesn't change anything other than breakpoints.

susanev (approved):

cnunciato:
@susanev thank you!

PR #309, opened 2021-06-10
fix nav toggle button

susanev:
## whats the problem
we present 2 hamburger nav buttons on the blog page for small screens
fixes comment in https://github.com/pulumi/pulumi-hugo/pull/306#issuecomment-858064052

i also removed the link to the blog from the blog because I did not understand why that was there... pls let me know if its serving a purpose I'm overlooking

## proposed fix
use a text label instead of an icon

#### before
blog-nav-before

#### after
blog-nav-after


cnunciato (approved):

susanev:
@cnunciato yikes, ya that is not good. ill open a separate issue for that.

PR #312, opened 2021-06-10
Fix heading wrapping on Solutions page

### LINKED ISSUE #169: Wrapping issues on the Solutions page
The headings on the Solutions page are getting clipped on mobile — for example:

5CA4DD92-9A41-49BB-9CF2-15F0CF4B624C

cnunciato:
Fixes #169.

image

### DIFF
--- a/themes/default/layouts/page/solutions.html
+++ b/themes/default/layouts/page/solutions.html
@@ -155,14 +155,27 @@
                 </div>
             </pulumi-choosable>
         </pulumi-chooser>
-        <h2 class="mb-2 lg:mb-8 order-first">
-            {{ .Params.personas.title }} <span class="rainbow-text">
-                <span class="developers hidden">for Developers</span>
-                <span class="devops hidden">for DevOps & Infra Teams</span>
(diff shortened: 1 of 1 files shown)


susanev (commented):
i cannot repro the original issue, the text gets smaller for me as expected. i think this change looks kinda strange to be honest, with Pulumi on its own line.

heading

cnunciato:
Oh, that's not intended -- lemme fix.

cnunciato:
@susanev PTAL -- this should fix the wrapping you noticed on desktop. 

It'll still wrap on mobile shown in the screenshot, but that's intentional in terms of what I'm submitting here; if I messed with this more, I _might_ be able to figure out how to coax the browser into positioning the the rainbowed "for" up on the same line as "Pulumi" (it has to do with how we're using the `background-clip` property somehow), but I stopped short of trying to figure that out, since this fixes what's broken and replaces it with something that's maybe not perfect, but definitely not broken. 😇

susanev (approved):

zchase (approved):

PR #327, opened 2021-06-14
fix logos in cicd guide

### LINKED ISSUE #326: logo size issue on ci/cd user guide
left over from the rebrand, we are using the smaller size of some of these logos
they should all be the same size

varrying-logo-sizes

susanev:
## overview
fixes #326
changes to `/docs/guides/continuous-delivery/`

the ci/cd images are used in a few spots
- `/docs/guides/continuous-delivery/`
- `/`
- `https://www.pulumi.com/docs/guides/crosswalk/kubernetes/#pulumi-kubernetes-operator`
- `https://www.pulumi.com/docs/intro/cloud-providers/kubernetes/#pulumi-kubernetes-operator`

this fix should not change any of the visuals for the home page, or either of the kubernetes pages

### before
before

### after
after


zchase (approved):
Looks fine to me. We use these logos on the home page slot machine as well. The ones you changed look _slightly_ bigger now I think, but I'll leave it you (@susanev ) to decide if they look fine or need to resized.

PR #335, opened 2021-06-16
Fix CI/CD logos, errant relrefs

cnunciato:
Fixes the Google Code Build logo and adjusts the relative sizing of grid items across mobile, tablet and desktop. Also fixes the `relref`s on this page (and the same ones referenced elsewhere) to make them properly root-relative.

https://user-images.githubusercontent.com/274700/122151405-5f642b00-ce14-11eb-8de1-6277254cfa0c.mov


susanev (approved):
thank you!!!

i dont have an opinion on this but just want to call out that this shifts travis to the first column instead of centered

PR #342, opened 2021-06-17
Provider content consolidation

davidwrede:
To whomever takes this over, this PR is to bring together the overview, setup, tutorials, examples, and reference content for each provider into one location in the docs (sans the getting started material).

See https://github.com/pulumi/home/issues/1399 for more details.

#446 is not about finishing the work or merge this pr, and is instead to evaluate the level of work required to finish the work

Work still remaining:

- [ ] Redirects needed for all content that's been moved
- [ ] Related to ☝️ , relref links should also be updated to refer to new locations
- [ ] Need to decide on whether to label providers as core and other or native and bridged
- [ ] add an `API Docs` item to the left nav for each package

## Related issues

https://github.com/pulumi/docs/pull/6154
https://github.com/pulumi/pulumi/pull/7334


susanev:
im a little bit worried that the API docs are a bit buried in this work. from what i understand they are only available as a link in the intro of the main page. wondering if we can get them added to the left nav as well.

aws

if im missing something obvious here pls do let me know.

cc/ @infin8x

infin8x:
@susanev 100% agree with adding them to the left nav for each provider

susanev:
@infin8x do you make this decision: ` Need to decide on whether to label providers as core and other or native and bridged`? or if not, who is responsible for making that decision.

infin8x:
I will make that decision as part of categorization

susanev:
@infin8x i believe yall decided not to continue this work, should we close the pr?

Issue #318, opened 2021-03-14
Blog should have search box


headconnect:
There's just so much in the blog that's absolutely great - and vital to get up and running in a productive way. Unfortunately, it's hard to find stuff in there sometimes, and it would be great to search.

I'm aware that the search function also searches blog content - but there is no searchbar in the blog itself.  Could this be added?

jonelleb:
# Design Goals
It is sometimes hard to find items in the blog. It would be a better user experience if there was a search bar that searched the blog itself, this would cut down the search parameters and make it easier to find particular blog articles.

# [Figma File](https://www.figma.com/file/8kz7uSX7MD430hCwl7q8Ll/Jonelle's-Miscellaneous-Work?node-id=49%3A899)

### Blog Design with New Search Bar

**Blog without scroll**

<img width="614" alt="Screen Shot 2022-03-24 at 8 08 58 AM" src="https://user-images.githubusercontent.com/100616532/159924741-c303c6ee-ab2b-4b8e-988e-2e7773b60e07.png">

**Blog with scroll**

<img width="616" alt="Screen Shot 2022-03-24 at 8 09 09 AM" src="https://user-images.githubusercontent.com/100616532/159924797-93b33f48-1b10-4001-8e5f-24231fcb83f7.png">

# Design Changes
- New search bar added in left nav of blog page right under 'Pulumi Blog' header.
- Please follow the same style, size, and search behavior as search bar on current Docs page.
- On scroll, Pulumi nav is pinned to the top and middle page content scrolls while the left nav with the search bar stays in the same place while the user scrolls (this is the same scroll behavior that the current Docs' page has - please follow that).

# Engineer Notes
- Responsive design should follow the same behavior as current Docs search bar does when that page is responsive (basically the responsive design for Blog stays the same as is now but the search bar is added)
- Comments have been added in Figma but please feel free to reach out with any questions or concerns

Thanks!

PR #350, opened 2021-06-19
Filter blog tags to Top 40, and add back some metadata

joeduffy:
In #215, [I had suggested](https://github.com/pulumi/pulumi-hugo/pull/215#issuecomment-840917671) that instead of physically deleting tags we didn't want to show, we compute it algorithmically, by only showing the "Top N" tags. It wasn't clear if this was possible, but alas, it is! This commit introduces said functionality.

This has a few advantages:

* Preserves old metadata (the authors added the tags because they felt they were meaningful and captured information about the posts).

* Enables us to surface those tags differently in the future (who knows, maybe someday we'll want to run a "spinnaker" campaign).

* Notably, also keeps the tag index pages, which Google has indexed.

* Enables us to add a "View More ..." link at the bottom of the page if folks want to see the entire list.

* Perhaps most importantly, protects against future bloat. For example, since this tag cleanup happened, we have added top-level tags for "aliases", "app-runner", "iam", "open-source", and "refactoring", each of which has only a single post.

I chose 40 as the N in Top N, because that's how many we show today. I could see an argument for filtering this based on post count instead (e.g., only those with >3 posts).

The change also adds back some of the previous tags now that we filter out the unpopular ones. I preserved all of the great cleanup that happened in that change and elided many that were clearly superfluous and/or unhelpful.

### DIFF
--- a/themes/default/content/blog/amazon-ecr-public/index.md
+++ b/themes/default/content/blog/amazon-ecr-public/index.md
@@ -4,7 +4,7 @@ date: "2020-12-01"
 meta_desc: "Pulumi container images now available on Amazon ECR Public"
 meta_image: "pulumi-images-ecr.png"
 authors: ["paul-stack"]
-tags: ["aws", "containers"]
+tags: ["aws", "containers", "ecr"]
 ---
 
 At re:Invent, the AWS team unveiled the new Amazon Elastic Container Registry Public (Amazon ECR Public), creating a new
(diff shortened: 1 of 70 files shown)


susanev:
at some point it would be really great if we can get search added to the blog as well, which will also help folks find what they are looking for. currently tracked in #318.

infin8x (approved):

PR #351, opened 2021-06-19
Make the Pulumi SDK reference docs more visible

joeduffy:
This change adds the Pulumi SDK to the left-hand nav and
cross-links in a few more places (the language guides), while
also making them a little more visually appealing.

Fixes pulumi/docs#4523.

image

### DIFF
--- a/config/_default/menus.yml
+++ b/config/_default/menus.yml
@@ -91,3 +91,23 @@ reference:
     parent: cli
     url: /docs/reference/cli/pulumi_whoami/
     weight: 18
+
+  - name: Node.js
+    parent: Pulumi SDK
+    url: /docs/reference/pkg/nodejs/pulumi/pulumi
+    weight: 1
+
+  - name: Python
+    parent: Pulumi SDK
+    url: /docs/reference/pkg/python/pulumi
+    weight: 2
+
+  - name: Go
+    parent: Pulumi SDK
+    url: https://pkg.go.dev/github.com/pulumi/pulumi/sdk/v3/go/pulumi
+    weight: 3
+
+  - name: .NET Core
+    parent: Pulumi SDK
+    url: /docs/reference/pkg/dotnet/Pulumi/Pulumi.html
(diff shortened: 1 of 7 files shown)


susanev:
what would yall think about making the left-nav menu expand like the CLI one and have the language choice there as well?
if we could also get external link icons added to the page (and the left nav items) that would also be rad.

joeduffy:
@cnunciato Gonna need your help with @susanev's suggestion. I do like the idea of the links in the left-hand nav, but unless I'm mistaken, that's not possible with our current navigation. If it's easy, I'd love to do it, if it's hard, probably will need to defer.

cnunciato:
@joeduffy This should work, yeah -- I think if you just add these lines (assuming they look okay):

```
- name: Node.js
  parent: Pulumi SDK
  url: /docs/reference/pkg/nodejs/pulumi/pulumi
  weight: 1

- name: Python
  parent: Pulumi SDK
  url: /docs/reference/pkg/python/pulumi
  weight: 2

- name: Go
  parent: Pulumi SDK
  url: https://pkg.go.dev/github.com/pulumi/pulumi/sdk/v3/go/pulumi
  weight: 3

- name: .NET Core
  parent: Pulumi SDK
  url: /docs/reference/pkg/dotnet/Pulumi/Pulumi.html
  weight: 4
```

.. to `config/_default/menus.yml`, you should get the collapsible submenu:

image

joeduffy:
TIL, this is great! @susanev if this looks good to you, I’ll roll with it and get this added and merged.

susanev:
yup this looks good, thank you both! i think later on we should come back and add external icons where appropriate but this is an issue across our docs, so no need to tackle that now.

joeduffy:
@cnunciato @susanev Looks like I need one last CR before I can merge, since I added those new links. Thanks.

susanev (approved):
looks good, we should double-check everything works correctly once it gets to docs

PR #352, opened 2021-06-19
Fix poor Python API docs formatting

joeduffy:
It appears we made a change at some point that made it so that our
CSS selector no longer selected the Python API docs as intended, leading
to virtually illegible rendering of those docs. Unless I'm mistaken, we
only use these docs for the Pulumi SDK now-a-days, and not the other
providers, so the fix just targets the styles to that section.

This isn't the first time this has happened (see
https://github.com/pulumi/docs/pull/1397), so I'd love to know if there
are any ideas on what we can do to structurally avoid this breaking again!

# Example
## Before
image
## After
image

### DIFF
--- a/themes/default/assets/sass/_api-python.scss
+++ b/themes/default/assets/sass/_api-python.scss
@@ -22,8 +22,8 @@ div.section[id^="pulumi"] {
     }
 }
 
-// We rely on the Python docs that begin with `<div class="section" id="foo">` to target these styles.
-div.section[id]:not([id=""]) {
+// This selector is for the Pulumi Python SDK page.
+section[id="pulumi-python-sdk"] {
     // Don't show the header link - there's an anchorjs link on this header that does the
     // job just as well.
     h1[id$="\B6"],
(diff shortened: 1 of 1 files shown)


cnunciato (approved):
Thanks @joeduffy! I wasn't aware this had changed.

One thing we can do to prevent this from happening again is add a test [like this one](https://github.com/pulumi/docs/blob/92cb7a2002083242041656e07d3cc3c11d1da44a/cypress/integration/site_spec.js#L9-L15) that checks computed CSS values to ensure they've been applied. (Looks like the comment's a bit out of date -- I updated the test when we did the rebrand, but apparently looked right past the comment 🤦.) We could also explore visual diffing tools, but so far, this simpler approach has been working out pretty well.

joeduffy:
Thanks @cnunciato. The best approach I could come up with here was to add a custom CSS attribute, and then check for its presence in the Cypress tests: https://github.com/pulumi/docs/pull/6158. Does this seem reasonable? How do you run those tests? `make test` didn't seem to do it for me?

PR #353, opened 2021-06-20
Fix the EKS networking example

### LINKED ISSUE #325: Example does not pass TypeScript type checking
The example [Configuring Your EKS Cluster’s Networking](https://www.pulumi.com/docs/guides/crosswalk/aws/eks/#configuring-your-eks-clusters-networking) does not pass TS checking.

[themes/default/content/docs/guides/crosswalk/aws/eks.md](https://www.pulumi.com/docs/guides/crosswalk/aws/eks/)

joeduffy:
This example doesn't compile. Unfortunately, the AWSX VPC constructor
appears to require an args bag, even though all properties of it are
optional. Also, the way you pass subnets seems to have changed in the
EKS class -- or maybe this was always a bug -- but the examples I
found in the EKS repo itself suggested this is the right way to do it.

Fixes #325.

### DIFF
--- a/themes/default/content/docs/guides/crosswalk/aws/eks.md
+++ b/themes/default/content/docs/guides/crosswalk/aws/eks.md
@@ -259,13 +259,13 @@ import * as awsx from "@pulumi/awsx";
 import * as eks from "@pulumi/eks";
 
 // Create a VPC for our cluster.
-const vpc = new awsx.ec2.Vpc("my-vpc");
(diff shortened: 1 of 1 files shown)


lukehoban (approved):
LGTM.

I opened https://github.com/pulumi/pulumi-awsx/pull/678 to (re-)allow constructing a VPC with no args bag.  This apparently regressed in https://github.com/pulumi/pulumi-awsx/pull/470.

joeduffy:
Ahhh, I thought I was going crazy thinking that this used to work 😄

PR #361, opened 2021-06-22
Fix typo in NGINX livestream abstract

zchase:
Just as the title states.

Fixes: https://github.com/pulumi/marketing/issues/44


isaac-pulumi (commented):
Could we do a quick edit to drop the live stream link and change the time to 1 hour instead of 2?

isaac-pulumi (approved):

PR #370, opened 2021-06-28
Add Service API reference page.

clstokes:
- [ ] Finish remaining TODOs

### DIFF
--- a/themes/default/content/docs/reference/service-api/_index.md
+++ b/themes/default/content/docs/reference/service-api/_index.md
@@ -876,6 +876,66 @@ Status: 200 OK
 
 ## Organizations
 
+### List Users
+
+TODO: explain `githubLogin` (synonomous with username), `knownToPulumi`, and `virtualAdmin`
+
+```
+GET https://api.pulumi.com/api/orgs/{organization}/members?type=frontend
+```
+
+#### Parameters
+
+| Parameter | Type | In | Description |
+| --------- | ---------- | ---------- | ---------- |
+| `organization` | string | path | organization name |
+| `type` | string | query | the response format to return - possible values are `frontend` or `backend` (TODO: explain the difference) |
+
+#### Example
+
+```bash
+curl \
+  -H 'Accept: application/vnd.pulumi+3' \
+  -H 'Content-Type: application/json' \
+  -H 'Authorization: token pul-abcdefghijklmnopqrstuvwxyz' \
+  https://api.pulumi.com/api/orgs/{organization}/members?type=frontend
+```
+
+#### Default response
+
+```
+Status: 200 OK
+```
+
+```
+{
+  "members": [
+    {
+      "role": "member",
+      "user": {
+        "name": "First Last",
+        "githubLogin": "user2",
+        "avatarUrl": "https://en.gravatar.com/userimage/17756222/cabc55626abae89ebe2d8ae946521e15.png?size=300",
+        "email": "user@example.com"
+      },
+      "knownToPulumi": true,
+      "virtualAdmin": false
+    },
+    {
+      "role": "admin",
+      "user": {
+        "name": "First Last",
+        "githubLogin": "user1",
+        "avatarUrl": "https://en.gravatar.com/userimage/17756222/cabc55626abae89ebe2d8ae946521e15.png?size=300",
+        "email": "admin@example.com"
+      },
+      "knownToPulumi": true,
+      "virtualAdmin": false
+    }
+  ]
+}
+```
+
 ### Add User to Organization
 
 Users are not just-in-time provisioned. They must sign in to the Pulumi console first (they will receive an error that they're not a member) and then they can be added to the organization.
--- a/themes/default/content/docs/reference/service-api/_index.md
+++ b/themes/default/content/docs/reference/service-api/_index.md
@@ -1,0 +1,1155 @@
+---
+title: "Pulumi Service API"
+meta_desc: An overview of the Pulumi Service API for querying Organization, Stack, State, etc. information.
+---
+
+The Pulumi Service API is used by the Pulumi CLI and the Pulumi Console to query and interact with state information, history, stack tags, etc. This API is available for end users to integrate into their own automation use cases.
+
+{{% notes "info" %}}
+While the Pulumi Service API is used by the Pulumi CLI and Pulumi Console, Pulumi makes no guarantees about changes that might affect backwards compatibility or cause breaking changes.
+{{% /notes %}}
+
+## Authentication
+
+All requests must be authenticated using a token via the `Authorization` HTTP header. If the token is missing or invalid, the Pulumi Service API will return a 4xx status code.
+
+The `Authorization` header must be in the form below with the literal string `token`, then a space, then your access token value.
+
+```
+Authorization: token {token}
+```
+
+{{% notes "info" %}}
+As of June 28, 2019, Pulumi access tokens contain the prefix `pul-`, which should be included as part of the token value in the header.
+{{% /notes %}}
+
+To view your access tokens, or create a new one, view the <a href="https://app.pulumi.com/account/tokens" target="_blank">Access Tokens</a> page. You will see a list of past tokens, when they were last used, and have the ability to revoke them.
+
+## Required Request Headers
+
+TODO: Add more info.
+
+```
+Accept: application/vnd.pulumi+3
+Content-Type: application/json
+```
+
+<!-- ###################################################################### -->
+
+## Stacks
+
+### List Stacks
+
+```
+GET https://api.pulumi.com/api/user/stacks
+```
+
+#### Parameters
+
+| Parameter | Type | In | Description |
+| --------- | ---------- | ---------- | ---------- |
+| `organization` | string | query | **Optional.** organization name to filter stacks by |
+| `project` | string | query | **Optional.** organization name to filter stacks by |
+| `tagName` | string | query | **Optional.** tag name to filter stacks by |
+| `tagValue` | string | query | **Optional.** tag value to filter stacks by |
+
+#### Example
+
+```bash
+curl \
+  -H 'Accept: application/vnd.pulumi+3' \
+  -H 'Content-Type: application/json' \
+  -H 'Authorization: token pul-abcdefghijklmnopqrstuvwxyz' \
+  https://api.pulumi.com/api/user/stacks?organization={organization}&tagName={tagName}&tagValue={tagValue}
+```
+
+#### Default response
+
+```
+Status: 200 OK
+```
+
+```
+{
+  "stacks": [
+    {
+      "orgName": "org1",
+      "projectName": "project1",
+      "stackName": "dev",
+      "lastUpdate": 1612481387,
+      "resourceCount": 0
+    },
+    {
+      "orgName": "org2",
+      "projectName": "project2",
+      "stackName": "prod",
+      "lastUpdate": 1605474464,
+      "resourceCount": 0
+    }
+  ]
+}
+```
+
+### Get Stack
+
+```
+GET https://api.pulumi.com/api/stacks/{organization}/{project}/{stack}
+```
+
+#### Parameters
+
+| Parameter | Type | In | Description |
+| --------- | ---------- | ---------- | ---------- |
+| `organization` | string | path | organization name |
+| `project` | string | path | project name |
+| `stack` | string | path | stack name |
+
+#### Example
+
+```bash
+curl \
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/reference/service-api/_index.md
```diff
@@ -0,0 +1,1465 @@
...
+
+The Pulumi Service API is used by the Pulumi CLI and the Pulumi Console to query and interact with state information, history, stack tags, etc. This API is available for end users to integrate into their own automation use cases.
+
+{{% notes "info" %}}
+While the Pulumi Service API is used by the Pulumi CLI and Pulumi Console, Pulumi makes no guarantees about changes that might affect backwards compatibility or cause breaking changes.
+{{% /notes %}}
+
+## Authentication
+
+All requests must be authenticated using a token via the `Authorization` HTTP header. If the token is missing or invalid, the Pulumi Service API will return a 4xx status code.
```
chrsmith, line 14: If the access token is invalid or missing, we will _always_ return a 401 status code. So we don't need to be vague and say 4xx.

themes/default/content/docs/reference/service-api/_index.md
```diff
@@ -0,0 +1,1465 @@
...
+
+The `Authorization` header must be in the form below with the literal string `token`, then a space, then your access token value.
```
chrsmith, line 16: This is fine as-is, but we may be able to rephrase this to sound less like we just made this up and point to the docs for the Authorization header and it's ~standard.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization

So the "authorization type" is `token`. But again, this is fine as-is.

themes/default/content/docs/reference/service-api/_index.md
```diff
@@ -0,0 +1,1465 @@
...
+
+```
+Authorization: token {token}
+```
+
+{{% notes "info" %}}
```
chrsmith, line 22: Given that we added the `pul-` a while ago, it might be OK to just remove this. Or phrase it like "Recently created Pulumi access tokens have a `"pul-"` prefix...

Also, in case you were wondering, you _can_ omit the `pul-` prefix and it will work just fine. (But we do want to encourage people to keep that prefix, so that automated scanners can look for tokens accidentally checked into source control and inform us so that we can automatically revoke them. e.g. GitHub does this.)

themes/default/content/docs/reference/service-api/_index.md
```diff
@@ -0,0 +1,1465 @@
...
+As of June 28, 2019, Pulumi access tokens contain the prefix `pul-`, which should be included as part of the token value in the header.
+{{% /notes %}}
+
+To view your access tokens, or create a new one, view the <a href="https://app.pulumi.com/account/tokens" target="_blank">Access Tokens</a> page. You will see a list of past tokens, when they were last used, and have the ability to revoke them.
+
+## Required Request Headers
+
+TODO: Add more info.
```
chrsmith, line 30: FYI, the MDN documentation is great source for info on this.

For context, the `Content-Type` is kinda assumed, but good to mention. The `Accept` header is used to mark the version of the Pulumi API you are using. e.g. whenever we make a breaking change to a CLI endpoint we bump the `Accept` header newer CLIs send. (But will maintain backwards compatibility if we see an older `vhd.pulumi+{version}` number.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept

Perhaps not something you should mention, but technically you can omit the `Accept` header, and we will default to a ~recent version number. But it isn't recommended, as omitting the `Accept` header can lead to breaking changes (if we change the "default" version number used).

themes/default/content/docs/reference/service-api/_index.md
```diff
@@ -0,0 +1,1465 @@
...
+```
+
+```
+{
+  "stacks": [
+    {
+      "orgName": "org1",
+      "projectName": "project1",
+      "stackName": "dev",
+      "lastUpdate": 1612481387,
```
chrsmith, line 79: In case it wasn't clear, this is a Unix timestamp of when the stack was last updated.

themes/default/content/docs/reference/service-api/_index.md
```diff
@@ -0,0 +1,1465 @@
...
+    "pulumi:secrets_provider": "service",
+    "vcs:kind": "github.com",
+    "vcs:owner": "pulumi",
+    "vcs:repo": "customer-engineering"
+  },
+  "version": 91
+}
+```
+
+### Get Stack State
```
chrsmith, line 149: "Stack State" might not be the best way to phrase this. (Though I'm not 100% sure how we refer to this elsewhere in our documentation.)

This returns the "checkpoint file" of the Pulumi Stack. So "stack state" is perfectly clear and accurate, but just maybe not the most precise/technical term to use.

chrsmith: If you ever do get questions about the specific schema/format, we want to make clear that it isn't meant to be a "human-friendly" representation of the resource state. And is instead, our own, internal view of the resources. Generally it's good enough for direct consumption, e.g. checking `output` properties and what not.

But there are no guarantees about what the `id` field means, e.g., as it is dependent on the implementation of the actual resource provider. So for AWS it is _usually_ the ARN, but with no guarantees. Also, when it comes to things like "stack references" or other oddities, they are flagged by having having a GUID as a property name which would be very confusing to people looking at the data...

Anyways, we just want to be a little vague that we are exposing the state information in a specific file format and not a general purpose/always-super-friendly way.

themes/default/content/docs/reference/service-api/_index.md
```diff
@@ -0,0 +1,1465 @@
...
+
+#### Example
+
+```bash
+curl \
+  -H 'Accept: application/vnd.pulumi+3' \
+  -H 'Content-Type: application/json' \
+  -H 'Authorization: token pul-abcdefghijklmnopqrstuvwxyz' \
+  --request POST \
+  --data '{"name":"{tagName}","value":"{tagValue}"}' \
```
chrsmith, line 299: Having the "Parameters" table makes this a little bit confusing.

As there are URL _route parameters_ (e.g. parts of the REST-ful URL). And there is the _POST request body_, which contains the JSON blob with a `name` and `value` field. Perhaps we'd want to call that out somehow, rather than putting it all into the same "Parameters" table? Also, in some places we accept URL-query parameters as well, which is a third class of "ways you pass data when making a REST API call".

themes/default/content/docs/reference/service-api/_index.md
```diff
@@ -0,0 +1,1465 @@
...
+## Stack Updates
+
+### List Stack Updates
+
+```
+GET https://api.pulumi.com/api/stacks/{organization}/{project}/{stack}/updates
+```
+
+By default the results are not paginated. You can specify `page` and `pageSize` query parameters to paginate the results.
+> `?pageSize=1&page=1` can be used to return only the most recent update.
```
chrsmith, line 360: If you want to get fancy, you can just use `"/{stackName}/updates/latest` instead. As that will do the same thing (but return just the latest update, not an array of updates with one element).

themes/default/content/docs/reference/service-api/_index.md
```diff
@@ -0,0 +1,1465 @@
...
+#### Parameters
+
+| Parameter | Type | In | Description |
+| --------- | ---------- | ---------- | ---------- |
+| `organization` | string | path | organization name |
+| `project` | string | path | project name |
+| `stack` | string | path | stack name |
+| `page` | number | query | **Optional.** page of the results to return |
+| `pageSize` | number | query | **Optional.** number of results per page |
+| `output-type` | number | query | **Optional.** the response format to return - possible values are `service` or `cli` (default) |
```
chrsmith, line 371: Ugh, yeah, this is a lame implementation detail bleeding through. Essentially the difference is that the "service" output type includes more data, whereas the CLI does not. But off the top of my head I don't recall the exact details.

chrsmith (commented):
Thank you so much for putting this together! Left a bunch of minor comments.

themes/default/content/docs/reference/service-api/_index.md
clstokes: I think I saw this note elsewhere in our docs or maybe a blog post, but since that was 2 years ago (from yesterday!) I'll remove.

themes/default/content/docs/reference/service-api/_index.md
clstokes: Going to leave as-is per our conversation.

themes/default/content/docs/reference/service-api/_index.md
clstokes: Going to leave this as-is per our convo.

themes/default/content/docs/reference/service-api/_index.md
clstokes: Given we refer to this as the "state" in many other places in the docs, I feel that introducing the term "checkpoint" would be confusing.

MitchellGerdisch:
The auditlogs API can be used without `/export` to return the audit logs in JSON instead of a zipped CSV. 
It's probably worth stating this in the note section or as an additional example.

Also, is github the only case where you cannot modify team membership? What about SAML/SCIM scenarios?

lukehoban:
It looks like this is not yet connected up to the ToC?  I'd suggest adding to the "Reference" section after "Pulumi SDK".  I think we could also link to this from http://pulumi-hugo-origin-pr-370-7d18cbea.s3-website.us-west-2.amazonaws.com/docs/intro/console/, http://pulumi-hugo-origin-pr-370-7d18cbea.s3-website.us-west-2.amazonaws.com/docs/intro/console/webhooks/, http://pulumi-hugo-origin-pr-370-7d18cbea.s3-website.us-west-2.amazonaws.com/docs/intro/console/audit-logs/ and other places where features that will have meaningful overlap with use of the REST API exist in the docs.

 Also - should we call this "Pulumi Service REST API" to be more explicit on what it is?

clstokes:
@MitchellGerdisch @chrsmith I updated the "Note" on `Update Team Membership` to try to explain the GitHub/SCIM nuances. Let me know what you think.

https://github.com/pulumi/pulumi-hugo/blob/7601ed812c2bad643caa7a7e95e8149cde5e0576/themes/default/content/docs/reference/service-api/_index.md#update-team-membership

MitchellGerdisch:
Typo in the SSO/SCIM text:
```
For SCIM managed teams, this operation cannot be used as membership as membership is managed via the SSO provider.
```
One too many "as memberships" in there.

clstokes:
> One too many "as memberships" in there.

Fixed.

clstokes:
> It looks like this is not yet connected up to the ToC? I'd suggest adding to the "Reference" section after "Pulumi SDK". I think we could also link to this from http://pulumi-hugo-origin-pr-370-7d18cbea.s3-website.us-west-2.amazonaws.com/docs/intro/console/, http://pulumi-hugo-origin-pr-370-7d18cbea.s3-website.us-west-2.amazonaws.com/docs/intro/console/webhooks/, http://pulumi-hugo-origin-pr-370-7d18cbea.s3-website.us-west-2.amazonaws.com/docs/intro/console/audit-logs/ and other places where features that will have meaningful overlap with use of the REST API exist in the docs.
> 
> Also - should we call this "Pulumi Service REST API" to be more explicit on what it is?

@lukehoban crosslinked on those pages and renamed per your suggestion.

@lukehoban @chrsmith can I get an official review on the PR so this can be merged? I think it's ready to go.

clstokes:
> Your site preview for commit [8e1a2a1](https://github.com/pulumi/pulumi-hugo/commit/8e1a2a1a02b0f62942520f60c58483f6616a3c4f) is ready! 🎉
> 
> http://pulumi-hugo-origin-pr-370-8e1a2a1a.s3-website.us-west-2.amazonaws.com.

Specific pages to look at based on this bucket:
- Main page - http://pulumi-hugo-origin-pr-370-8e1a2a1a.s3-website.us-west-2.amazonaws.com/docs/reference/service-rest-api/
- Console http://pulumi-hugo-origin-pr-370-8e1a2a1a.s3-website.us-west-2.amazonaws.com/docs/intro/console/#integrations-and-extensions
- Webhooks - http://pulumi-hugo-origin-pr-370-8e1a2a1a.s3-website.us-west-2.amazonaws.com/docs/intro/console/webhooks/#additional-resources
- Audit Logs - http://pulumi-hugo-origin-pr-370-8e1a2a1a.s3-website.us-west-2.amazonaws.com/docs/intro/console/audit-logs/#exporting-audit-logs-using-the-api

MitchellGerdisch (approved):
LGTM

PR #373, opened 2021-06-29
Add more language coverage for Crosswalk/EKS docs

joeduffy:
The Crosswalk/EKS docs are very good and are ranking highly on SEO.
However, we only had TypeScript examples, despite recently broadening
language coverage for our EKS component. This change adds language
choosers and equivalent examples in JavaScript, Python, Go, and C#.

Part of https://github.com/pulumi/docs/issues/3171.

### DIFF
--- a/themes/default/content/docs/guides/crosswalk/aws/eks.md
+++ b/themes/default/content/docs/guides/crosswalk/aws/eks.md
@@ -61,6 +61,25 @@ These are not required but are recommended if you plan on interacting with your 
 
 To create a new EKS cluster, allocate an instance of an `eks.Cluster` class in your Pulumi program:
 
+{{< chooser language "javascript,typescript,python,go,csharp" / >}}
+
+{{% choosable language javascript %}}
+
+```javascript
+const aws = require("@pulumi/eks");
+
+// Create an EKS cluster with the default configuration.
+const cluster = new eks.Cluster("my-cluster");
+
+// Export the cluster's kubeconfig.
+module.exports = {
+    kubeconfig: cluster.kubeconfig,
+};
+```
+
+{{% /choosable %}}
+{{% choosable language typescript %}}
+
 ```typescript
 import * as eks from "@pulumi/eks";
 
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/guides/crosswalk/aws/eks.md
```diff
@@ -243,6 +631,100 @@ const cluster = new eks.Cluster("my-cluster", {
...
+using Pulumi.Eks;
+
+class MyStack : Stack
+{
+    public MyStack()
+    {
+        // Create an EKS cluster with the default configuration.
+        var cluster = new Cluster("my-cluster", new ClusterArgs {
+            DesiredCapacity = 5,
+            MinSize =3,
```
lukehoban, line 708:
```suggestion
            MinSize = 3,
```

themes/default/content/docs/guides/crosswalk/aws/eks.md
```diff
@@ -147,6 +231,57 @@ Any existing tools will work here, including `kubectl`, Helm, and other CI/CD pr
 to define Kubernetes application-level objects and configuration in code too. For instance, we can deploy a canary
 to our EKS cluster in the same program if we want to test that it is working as part of `pulumi up`:
 
+{{< chooser language "javascript,typescript,python,go,csharp" / >}}
+
+{{% choosable language javascript %}}
```
lukehoban, line 236: I'm not sure we have a hard and fast policy on this yet - but my own guidance has been that we should move toward not having both JavaScript and TypeScript examples in content like this.  The incremental value is fairly low, we are not consistent on it in many other places (most notably resource provider examples) so it sets users who do want it up for disappointment, and the maintenance burden of yet another manually managed copy is quite high.  I suggest we actually formalize this as our policy for mutli-language content.  Fine to leave these in here pending that decision if you'd prefer - but this was a reminder that we've avoided doing similar work in other places recently.

themes/default/content/docs/guides/crosswalk/aws/eks.md
```diff
@@ -190,6 +325,232 @@ export const url = service.status.loadBalancer.ingress[0].hostname;
...
+    spec = k8s.core.v1.ServiceSpecArgs(
+        type = 'LoadBalancer',
+        selector = app_labels,
+        ports = [ k8s.core.v1.ServicePortArgs(port = 80) ],
+    ), opts = pulumi.ResourceOptions(provider = cluster.provider)
+)
+
+# Export the URL for the load balanced service.
+pulumi.export('url', service.status.apply(
+    lambda status: status.load_balancer.ingress[0].hostname))
```
lukehoban, line 364: Perhaps it's not desirable here for some reason - but I would have expected `service.status.load_balancer.ingress[0].hostname` to work.  Quite possible completion lists aren't available for this though.

lukehoban (approved):
I did a spot check, but have not tried to manually verify that all of these examples actually work correctly in all languages.  

This is a reminder that we almost certainly need to take large and complex examples like these and check them in and run tests against them to ensure they work and continue working.  I'm not sure if some subset of these are already examples in the examples repo, available in all languages.  But if not, we likely will want to go add them there (or in a test suite in this repo).  

As noted in https://github.com/pulumi/pulumi-hugo/issues/343, it would be useful in general if all the CrossWalk docs linked to complete working examples in GitHub along with the large snippets of inline code.

joeduffy:
> I did a spot check, but have not tried to manually verify that all of these examples actually work correctly in all languages.

I did manually verify them. I agree we need a better approach to embedding working code samples, ideally by linking to code checked in somewhere running during tests. Two questions

1) Is that work on the roadmap?
2) In its absence, do you prefer that I do something differently than what's in this pull request? It felt like the redundancy of maintaining it in multiple places would be a burden, but I'm happy to do that if you prefer.

themes/default/content/docs/guides/crosswalk/aws/eks.md
joeduffy: For what it's worth, I actually assumed we'd go in the opposite direction, and maintain both, since TypeScript != JavaScript, and the latter has a larger community than the former. I agree for most examples the delta is tiny (although when it's material, it can be quite jarring to the uninitiated (generics and whatnot)) -- to me, it's honestly as much of a marketing thing, than being about the differences.

themes/default/content/docs/guides/crosswalk/aws/eks.md
joeduffy: Good catch, that does indeed work! Fixed.

PR #381, opened 2021-07-01
Update unit.md

jandom:
The links are pointing to an old commit, which contains code that no longer works – it'd be more helpful to just point to something newer, here I picked master because it's the simplest

### DIFF
--- a/themes/default/content/docs/guides/testing/unit.md
+++ b/themes/default/content/docs/guides/testing/unit.md
@@ -960,28 +960,28 @@ All the tests passed!
 
 {{% choosable language "typescript" %}}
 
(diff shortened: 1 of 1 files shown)


stack72 (approved):
thanks so much for this @jandom - sorry that was a broken link for you

jandom:
wow that's a fast merge, many thanks @stack72 – will submit some further changes for using pytest over unittest. It's my understanding that pytest is superior/more popular than unittest in the python community, I think this would be useful

stack72:
> wow that's a fast merge, many thanks @stack72 – will submit some further changes for using pytest over unittest. It's my understanding that pytest is superior/more popular than unittest in the python community, I think this would be useful

We welcome the changes :)

PR #396, opened 2021-07-07
Add Cloud Engineering Summit to header submenu

zchase:
This PR adds the Cloud Engineering Summit landing page as submenu item to Cloud Engineering. The motivation for the change is to expose a link to the summit hype page that is discoverable. In the Cloud Engineering submenu felt like the natural place to put this link but open to other suggestions.

### DIFF
--- a/themes/default/layouts/partials/header.html
+++ b/themes/default/layouts/partials/header.html
@@ -42,6 +42,11 @@
                                     </a>
                                 </li>
                                 <li>
+                                    <a href="{{ relref . "/cloud-engineering-summit" }}">
+                                        2021 Summit
+                                    </a>
+                                </li>
+                                <li>
                                     <a href="{{ relref . "/solutions" }}">
                                         Solutions
                                     </a>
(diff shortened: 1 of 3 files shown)


themes/default/layouts/partials/header.html
```diff
@@ -51,6 +51,11 @@
                                         Case Studies
                                     </a>
                                 </li>
+                                <li>
+                                    <a href="{{ relref . "/cloud-engineering-summit" }}">
+                                        Cloud Engineering Summit
```
susanev, line 56:
```suggestion
                                        2021 Summit
```

not sure if this is any better, but thought this might make it a bit more clear. id also advocate that we make it the 2nd item in the list below the overview.

susanev (approved):

cnunciato (approved):
Seems fine to me, defer to y'all on the fine-tunings.

themes/default/layouts/partials/header.html
zchase: Seems reasonable to me --- makes it more clear this is an upcoming event so let's go with it!

PR #404, opened 2021-07-12
Fix syntax error and remove semicolons

1oglop1:
The example python code contains a syntax error and semicolons that are not usually used in python.


stack72 (approved):
Thank you so much for this!

PR #412, opened 2021-07-15
Fixed csharp test examples

pierskarsenbarg:
They are currently using a static method that isn't static.

Plus in one place we were getting userdata when it should have been tags﻿


mikhailshilkov (approved):
TY!

Issue #422, opened 2021-07-21
Missing Configuring Secrets Encryption section


jimmyp:
[themes/default/content/docs/intro/concepts/config.md](https://www.pulumi.com/docs/intro/concepts/config/)

Multiple other document pages, such as https://www.pulumi.com/docs/intro/concepts/state/#secrets link to this page pointing to a Configuring Secrets Encryption section which no longer seems to exist. The link is https://www.pulumi.com/docs/intro/concepts/config/#configuring-secrets-encryption

jimmyp:
I suspect it is supposed to link to https://www.pulumi.com/docs/intro/concepts/secrets/

stack72:
Closer via #424

PR #424, opened 2021-07-21
fixes link from state to secrets

jimmyp:
to address issue #422


stack72 (approved):
Thanks so much for this @jimmyp

Issue #414, opened 2021-07-15
update cloud eng logo


susanev:
## overivew
update the old cloud eng logo to the new one
use it in the header and remove the logo on the page

logo: https://drive.google.com/file/d/1r2Reugm6fox6nZh71s_10ov3UtSJkrJv/view?usp=sharing

zchase:
The logo was updated in this PR: https://github.com/pulumi/pulumi-hugo/pull/562

Going to close this out but feel free to re-open if something was missed.

PR #431, opened 2021-07-22
Update cloud engineering summit logo

zchase:
Fixes: https://github.com/pulumi/pulumi-hugo/issues/414

Updating the logo as per the issue and placed it into the hero. The only issue is that using the image in hero removes the `h1` from the page and we need that for SEO. I added the `h1` back in lower on the page but very open to other suggestions on how to get the `h1` tag into the layout now that the hero uses an `.svg`.


susanev:
@zchase can you add an h1 to the page with some css to visually hide it instead?

zchase:
@susanev Hiding the `h1` is a no go SEO wise: https://developers.google.com/search/docs/advanced/guidelines/hidden-text-links?hl=en&visit_id=637626547601287758-1073258535&rd=1

If we visually hide the `h1` we are risk to take an SEO penalty so we really need it on the page and visible.

susanev:
okay cool thx, let me think on this just a bit and then ill reach out to you on slack.

susanev (approved):

Issue #387, opened 2021-07-02
reduce the rows of dots on pages other than the homepage


susanev:
## overview
to prevent hiding information on small screens, lets reduce the rows of dots on pages other than the home page

## design details

- [ ] reduce the rows of dots both above and below the headings to ~4 rows above and below
    - [ ] if the heading ends up wrapping, still maintain ~4 rows above and below, if possible. (i.e., the header on that page will but a different size)

susanev:
no urgency on this, but getting it fixed sometime in july would be rad

PR #433, opened 2021-07-22
Reduce size of non home page hero sections

zchase:
Fixes: https://github.com/pulumi/pulumi-hugo/issues/387

This PR reduces the size of the non home page hero sections.


susanev (approved):
yay!! you are the best, thank you!!

PR #454, opened 2021-07-28
Learn Pulumi with a module and some topics

cnunciato:
This example takes the [Introduction to Pulumi on AWS tutorial](https://github.com/pulumi/tutorials/tree/main/introduction_to_pulumi) and ports it to a Learn Pulumi "module" (which we will soon rename to "tutorial") to give us a sense of what we're currently working with.


mattstratton:
also just in case people are looking for it

https://github.com/pulumi/pulumi-hugo/blob/master/LEARN.md

PR #457, opened 2021-07-28
Learn Pulumi style adjustments

cnunciato:


themes/default/assets/sass/_marketing.scss
```diff
@@ -4,7 +4,7 @@
     }
 }
 
-body:not(.section-docs):not(.section-blog):not(.section-authors):not(.section-tags):not(.section-legal) {
+body:not(.section-docs):not(.section-blog):not(.section-learn):not(.section-authors):not(.section-tags):not(.section-legal) {
```
zchase, line 7: Not for this PR, but the approach I took here is not great (`body:not(...)`). These rules override everything so we can't actually override these styles using utility classes. Just something to think about.

zchase (approved):
LGTM. Thank you for doing this!!

themes/default/assets/sass/_marketing.scss
cnunciato: Yep, agree we should come back to this. I’ve been meaning to do a more holistic assessment of all this post-rebrand, but still haven’t managed to get to it. Thanks for bringing it back up — will see if I can make some time soon.

PR #460, opened 2021-07-29
Update comment in blog post archetype for clarity

mattstratton:
The linter task will flag/block if this field is omitted, so it's not optional. This provides clarity for folks doing blogs for the first time.


cnunciato (approved):
👍 Thanks!

PR #464, opened 2021-07-30
Move azure blog series to draft

mattstratton:
We would like to schedule this out for next week, so marking them as draft mode for now


aaronkao (approved):
Reapproving what was previously already merged.

PR #468, opened 2021-08-02
[Blog] cloud engineering on the rise

spara:


gthuang (changes requested):
could you add the logos to each case study?

gthuang (approved):

PR #472, opened 2021-08-03
Update Kenshoo branding to Skai

gthuang:
Kenshoo rebranded to Skai, so we need to update the copy. We also need to update the Kenshoo logo to Skai logo

The changes to the blog post are in this PR: https://github.com/pulumi/pulumi-hugo/pull/473


gthuang:
@susanev we need to replace the Kenshoo logo in here with the Skai logo. do you have a file we could use here?

susanev (approved):

PR #484, opened 2021-08-06
adjust top nav to use a stronger font-weight instead of opacity on hover

### LINKED ISSUE #427: redesign marketing top-navigation menus
This is a super minor piece of feedback, however, I find it odd that we grey out the top nav links when you hover over them. Typically in most websites I use, the hover behavior is to make the target link look _more_ activated rather than _less_, and to me greying something out typically indicates it is unimportant, not clickable, etc.

For instance:

image

Notice when I hover over Cloud Engineering, I get:

image

I assume this was intentional (I see in the original website design it was present also), and perhaps I'm the only person with this reaction, however, it strikes me as odd every time I see it so I figured I would mention.

susanev:
## overview
moving from applying opacity to menu items on interaction to making them have a stronger weight for usability improvements.
fixes #427

### DIFF
--- a/themes/default/assets/sass/_header.scss
+++ b/themes/default/assets/sass/_header.scss
@@ -96,6 +96,7 @@
 
                 a {
                     @apply block px-4 my-1 transition-all;
+                    transition: font-weight .1s ease-in-out;
                 }
 
                 .dropdown-title {
@@ -123,6 +124,7 @@
 
                             a {
                                 @apply font-normal;
+                                transition: font-weight .1s ease-in-out;
 
                                 &:hover {
                                     @apply font-bold;
--- a/themes/default/assets/sass/_header.scss
+++ b/themes/default/assets/sass/_header.scss
@@ -123,9 +123,7 @@
 
                             a {
                                 @apply font-normal;
-                            }
 
-                            a {
                                 &:hover {
                                     @apply font-bold;
                                 }
--- a/themes/default/assets/sass/_header.scss
+++ b/themes/default/assets/sass/_header.scss
@@ -94,18 +94,8 @@
             li:not(.header-dropdown-item) {
                 @apply flex items-center ;
 
-                a::before, .dropdown-title::before {
-                    display: block;
-                    content: attr(data-content);
-                    @apply font-bold;    
-                    height: 0;
-                    overflow: hidden;
-                    visibility: hidden;
-                }
-
                 a {
                     @apply block px-4 my-1 transition-all;
-                    transition: font-weight .1s ease-in-out;
                 }
 
                 .dropdown-title {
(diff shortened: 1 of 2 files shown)


susanev:
@kimberleyamackenzie is gonna take over this work

zchase (commented):
I think we are going to want to adjust the CSS so that the other menu items don't move when the bold text is applied. Thoughts @cnunciato ?

kimberleyamackenzie:
@zchase Yeah, susan and I chatted about this yesterday and I'm going to adopt the PR + move to adjusting text shadow instead of font weight to avoid the shifting and make the transition a little less abrupt.  I have a few other things on my plate, but am aiming to get to this this afternoon.

cnunciato:
Agree with this improvement for sure. We could also just darken the text color on hover.

kimberleyamackenzie:
Adventures in CSS!  I pushed a change that uses the bold on hover strategy, but adds an invisible bold pseudo-element to ensure that the other elements don't experience that shifting.  I landed here after trying a few different things, all with some drawbacks - let me know what y'all think, or if you like one of the below better. @cnunciato @zchase 

Other ideas I tried:

Following your suggestion, @cnunciato, and doing black text on hover.  The contrast didn't feel enough to me:
Screen Shot 2021-08-11 at 11 13 49 AM

Using text shadow to emulate font weight - this one felt a little fuzzy to me around the edges because of the blur radius:
Screen Shot 2021-08-11 at 11 15 07 AM

Using text stroke - I liked this the best, but text stroke is technically non-standard:
Screen Shot 2021-08-11 at 11 14 45 AM

cnunciato:
@kimberleyamackenzie Do you happen to have a commit you could push that shows the text-stroke approach? I see that it's technically non-standard, but it also looks pretty well supported. (The jump between normal and bold looks a bit abrupt to me, so would just like to see what the others look like in comparison.)

kimberleyamackenzie:
@cnunciato Here's a gif demo, can definitely push a commit if you want to play around
Aug-11-2021 11-55-06

cnunciato:
Yeah I definitely prefer that, too. I'd be 👍 on going that route for now, since it looks good, is currently well-supported, and is easily updated if support for text-shadow suddenly goes away.

NOW my question is ... is it `transition`able? 😇

kimberleyamackenzie:
@cnunciato Pushed the version with the text stroke.  Weirdly (i guess this is where "non-standard" comes in), only the stroke color is `transition`able, not the stroke width - so adding the transition just looks kind of bizarre, in this case, in my opinion.

susanev:
@cnunciato did you have more thoughts you wanted to share about this?

cnunciato:
@susanev @kimberleyamackenzie and I did some back-and-forth on this after that last commit, and neither of us was able to come up with something that we felt quite worked, so we set it aside and haven't come back to it yet.

PR #489, opened 2021-08-09
adjust messaging on old pricing blog

susanev:
folks are starting to get confused by this old pricing blog and the new pricing model
this drafts the old blog so folks don't continue to use the information in it

[community slack convo](https://pulumi-community.slack.com/archives/C01PF3E1B8V/p1628510974034900)
[internal slack convo](https://pulumi.slack.com/archives/C01ADF2C8KC/p1628514904003900)
[internal community chat convo](https://pulumi.slack.com/archives/CG326KPTM/p1628511070033200)

### DIFF
--- a/themes/default/content/blog/announcing-per-user-pricing-and-unlimited-stacks-for-teams/index.md
+++ b/themes/default/content/blog/announcing-per-user-pricing-and-unlimited-stacks-for-teams/index.md
@@ -1,17 +1,19 @@
 ---
 title: "Announcing Per User Pricing and Unlimited Stacks for Teams"
 date: "2019-04-19"
-draft: true
+draft: false
 meta_desc: "Today we are announcing Pulumi's new pricing tier, with three paid editions: Team Starter Edition, Team Pro Edition, and Enterprise Edition."
 authors: ["joe-duffy"]
 tags:
   - features
 ---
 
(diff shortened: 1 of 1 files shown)


cnunciato:
Alternatively, we could de-list the post rather than just remove it, since it was true at one point (and I think we have several posts that have similarly gone out of date). We could do that now with `block_external_search_index: true`, but we could also go a step further by adding general support for flagging outdated posts like this one to remove them from Google, add the on-page note, remove them from tag/author pages, Swiftype, etc.

susanev:
Can you say more about why you think it is important to keep around? What value does it bring?

I am concerned about this particular one because it is confusing, and it's easy to overlook the date when you are quickly scanning for information. This pr was motivated by a customer finding and using this blog and being confused. I linked to the discussions in the pr description.

susanev:
chatted with @lukehoban about this, we are going to try linking to the new blog, and using the warning style instead of the note style for now. and if we get more feedback about this we can revisit it then.

susanev:
we should update the color we use for warning, so it has more contrast on white. ill do that later today.

susanev:
note styles will be fixed in https://github.com/pulumi/pulumi-hugo/pull/559

cnunciato (approved):

PR #507, opened 2021-08-12
Revise pricing FAQ "how many credits will I need?"

gthuang:
The answer to this FAQ is confusing some customers per feedback from the CE team. People do not know what "production" means, the number of environments chosen in each scenario seems to cause confusion, and the wording around "deployment" also seems to throw people off. 

I revised the answer to make it less ambiguous and more consistent.


pierskarsenbarg:
Looks a lot clearer

pierskarsenbarg (approved):

PR #510, opened 2021-08-12
Fix some broken links

cnunciato:


stack72 (approved):

cnunciato:
@spara This makes a slight (hopefully okay) tweak to your post -- the link checker fails with `/generate` because that page requires the GitHub user be signed in (which the link checker isn't). So I added a note to click the `Use this template` button -- not quite as slick as what you had before, but hopefully okay. (If not, we'll need to add an exclusion in pulumi/docs for it.)

PR #518, opened 2021-08-17
[Blog] Top 5 Things an Azure Developer Needs to Know: Kubernetes Applications

spara:


pierskarsenbarg:
If I were reading a blog post on how to deploy all this stuff using Pulumi and the first half of it was how to deploy without Pulumi, I'd probably give up and go elsewhere. Do we need all the `az` commands and yaml?

spara:
This series is less about Pulumi and more about Azure. The reasoning is
that there are more people looking for information about Azure and that
we're presenting Pulumi as an alternative means to do the same thing but
with code.

pierskarsenbarg (approved):
Fair enough. Looks good then

PR #530, opened 2021-08-19
Add `pulumi refresh` tip.

gregwebs:
`pulumi refresh` worked for me.


cnunciato (approved):
LGTM, but would love to get @lblackstone's perspective also.

lblackstone (approved):
LGTM

I believe this is caused by https://github.com/pulumi/pulumi-kubernetes/issues/1649, so this shouldn't be required once that gets fixed.

susanev:
@cnunciato @lblackstone should this be merged?

PR #532, opened 2021-08-20
Add ian to the website.

iwahbe:


justinvp (approved):
Looks like the image needs to have a `.jpg` extension rather than `.jpeg`.

https://github.com/pulumi/pulumi-hugo/blob/37d95cc500f16a279cbfaa3f388b990148ef2e50/themes/default/layouts/page/about.html#L28

Otherwise, LGTM!

PR #533, opened 2021-08-20
Fix header not to cover search modal

SDuck4:
When search in docs, the header container covers search modal.
So I moved `sticky` class of `.header-container` to `&.is-pinned`.

Before:
Screen Shot 2021-08-20 at 17 04 58

After:
Screen Shot 2021-08-20 at 17 05 25


cnunciato (approved):
Nice find! Thanks for the fix. ⭐

PR #537, opened 2021-08-21
Update the examples Stencil component to support non-h3 based toggles

praneetloke:
Also added some styles related to package docs.

Related to https://github.com/pulumi/pulumi-service/issues/7229.


praneetloke:
Closing in favor of the styles and components living as part of the `pulumi/theme` repo.

PR #551, opened 2021-08-30
improve messaging related to orgs and subscriptions

susanev:
## overview
its not super clear to folks that when they create a new organization it means a new subscription (unless they have a custom invoiced plan that accounts for it)

some convo about this in this [internal slack thread](https://pulumi.slack.com/archives/C8A7QG1CZ/p1630289756004200)

### DIFF
--- a/themes/default/content/docs/intro/console/organizations.md
+++ b/themes/default/content/docs/intro/console/organizations.md
@@ -15,7 +15,6 @@ aliases:
 ---
 
 Organizations are a space for you to collaborate on shared projects and stacks.
-New organizations have their own subscription.
 
 ## Organization Pages
 
@@ -30,7 +29,9 @@ New organizations have their own subscription.
 
 ## Creating an Organization
 
-Every organization has it's own subscription.
+Creating an organization will start a free trial that has access to all features.
+At the end of the trial, you can choose to move to Team or Enterprise.
+Learn more about the edition features on the [Pricing Page]({{< relref "/pricing" >}}).
 
 To create an organization:
 
--- a/themes/default/content/docs/intro/console/organizations.md
+++ b/themes/default/content/docs/intro/console/organizations.md
@@ -15,6 +15,7 @@ aliases:
 ---
 
 Organizations are a space for you to collaborate on shared projects and stacks.
+New organizations have their own subscription.
 
 ## Organization Pages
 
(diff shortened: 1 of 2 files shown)


susanev:
@aureq idk what im doing wrong, but i cant seem to add you as a reviewer

themes/default/content/docs/intro/console/organizations.md
```diff
@@ -15,6 +15,7 @@ aliases:
 ---
 
 Organizations are a space for you to collaborate on shared projects and stacks.
+New organizations have their own subscription.
```
lukehoban, line 18: This doesn't feel quite right in this header section - might leave out here - but flesh out more below.

themes/default/content/docs/intro/console/organizations.md
```diff
@@ -29,6 +30,8 @@ Organizations are a space for you to collaborate on shared projects and stacks.
 
 ## Creating an Organization
 
+Every organization has it's own subscription.
```
lukehoban, line 33: I feel like we probably could put this in-context in the bullet list below as well or instead. 

I'm also not sure "subscription" is a term we use anywhere, so it may not be clear what this means.

I expect we want to describe a little more fully as part of the "start a free trial" note below that each organization will start a trial of Enterprise (linking to pricing page or product details page), and can choose to move to Team or Enterprise before their trial completes.  Then in that context, noting that each organization is on a separate plan.

themes/default/layouts/page/pricing.html
```diff
@@ -638,7 +638,7 @@ <h2 class="text-center">Compare editions</h2>
                         <span>1</span>
                     </div>
                     <div class="compare-plans-table-row-item">
-                        <span>Unlimited</span>
+                        <span>Custom</span>
```
lukehoban, line 641: I think the original text here is more what was and is intended for the product as we actually offer it today - I'm not sure I'd change this yet until we've gotten a bit clearer on whether the cases that might motivate a change are actually intentional or not.

themes/default/content/docs/intro/console/organizations.md
susanev: we still use subscription in the console, but its been mostly removed from the website and replaced with edition, which is kinda funky to use in a sentence, but ill give it a try.

susanev: @lukehoban i tried again, let me know what you think. id prefer to keep the list of instructional steps short and not too complicated. but if you see a way to add it in there pls let me know.

zchase (approved):
LGTM.

PR #559, opened 2021-09-02
make note styles accessible and align them to the new brand colors

### LINKED ISSUE #558: make note styles accessible and align them to our new brand
## overview
our note styles are inaccessible currently and are a bit off from our new brand colors.
for now, thats info and warning. there seemed to be styles for deprecated as well but they are the same as warning.

#### warning
Screen Shot 2021-09-02 at 9 25 45 AM

#### info
Screen Shot 2021-09-02 at 8 49 58 AM

## design details
- [x] change info background color to `#4d5bd9`
- [x] change warning background color to `#bd4c85`
- [x] change heading text color to `#fff` (note body text color should remain the same)

#### designs
note-styles

#### info in context
Screen Shot 2021-09-02 at 9 19 37 AM

#### warning in context
Screen Shot 2021-09-02 at 9 21 49 AM

susanev:
fixes #558

### DIFF
--- a/themes/default/assets/sass/_notes.scss
+++ b/themes/default/assets/sass/_notes.scss
@@ -28,26 +28,26 @@
     }
 
     &.note-info {
-        @apply border-blue-100;
+        @apply border-blue-600;
 
         .note-heading {
-            @apply bg-blue-100 text-blue-900;
+            @apply bg-blue-600 text-white;
         }
     }
 
     &.note-warning {
-        @apply border-orange-100;
+        @apply border-fuchsia-600;
 
         .note-heading {
-            @apply bg-orange-100 text-orange-900;
+            @apply bg-fuchsia-600 text-white;
         }
     }
 
     &.note-deprecated {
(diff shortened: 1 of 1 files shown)


infin8x (approved):
Nice! 🚀

cnunciato (approved):
This looks great for the notes themselves, but do you think we should use the treatment as our primary buttons for these? E.g., just as a side-by-side:

 

If so, 👍, but figured I'd ask just in case it made sense to consider something slightly less visually prominent for them, particularly notes.

susanev:
its a good callout. our palette is fairly limited and that is the first blue in our palette that is accessible on white. so if i were to pick a different blue it would be darker, and possibly communicate a stronger affordance than the primary buttons. alternatively, we could use a different color entirely, but in my experience blue is very commonly associated with blue.

im good with this for now, until we could do more extensive work with our brand pallette that allows for more range with a specific focus on accessibility.

PR #562, opened 2021-09-03
Add registration for 2021 Cloud Engineering Summit

zchase:
This PR adds registration for the Cloud Engineering Summit, adds a speaker wall, and puts the logo into the hero. To ensure we still have an `h1` tag on the page I put it above the "Featured Speakers" text but open to other suggestions.


susanev (approved):
this looks good to me.

a few comments, neither blocking.

- [x] those large blocks of text for manage, deploy and build are more than likely not going to be read by anyone cause its so dense in text. id really encourage us to limit them to 3 or less lines in general. alternatively, breaking them up a bit into multiple paragraphs may help folks scan it for information relevant to them.
- [ ] id drop the bullet points from the lists also in that section, and bump up the space between individual lines just a tad

PR #565, opened 2021-09-07
use descriptive page names instead of _other_ in the docs

### LINKED ISSUE #560: use descriptive page names instead of _other_ in the docs
## overview
i noticed we use _other_ to describe a few pages, which isn't very helpful to folks scanning the nav because its not descriptive. id like to propose we update this page names and URLs and add redirects.

## design details
- [x] change o[ther in continuous delivery](https://www.pulumi.com/docs/guides/continuous-delivery/other/) to `Create New CI/CD Integration`
- [x] change [other in crosswalk for aws](https://www.pulumi.com/docs/guides/crosswalk/aws/other/) to `AWS Index of Services`
- [x] change [pulumi vs other ](https://www.pulumi.com/docs/intro/vs/)to `Compare Pulumi to...` h/t @infin8x

susanev:
fixes #560

### DIFF
--- a/themes/default/content/docs/guides/continuous-delivery/create-new-cicd-integration.md
+++ b/themes/default/content/docs/guides/continuous-delivery/create-new-cicd-integration.md
@@ -1,15 +1,15 @@
 ---
-title: Other
+title: Create New CI/CD Integration
 meta_desc: This page walks-through how you can modify the Pulumi CLI to support your own
            CI system or perhaps even update an existing one.
 menu:
     userguides:
         parent: cont_delivery
         weight: 2
-
 aliases:
 - /docs/reference/cd-supporting-new-ci/
 - /docs/console/continuous-delivery/other/
+- /docs/guides/continuous-delivery/other/
 ---
 
 If Pulumi doesn't already support your CI/CD system, this guide
--- a/themes/default/content/docs/intro/compare-pulumi-to/terraform.md
+++ b/themes/default/content/docs/intro/compare-pulumi-to/terraform.md
@@ -1,13 +1,15 @@
 ---
-title: Pulumi vs. Terraform
+title: Compare Pulumi to Terraform
 meta_desc: Pulumi is like Terraform—create, deploy, and manage infrastructure as code on any cloud. But unlike Terraform you can use familiar languages and tools.
 linktitle: Terraform
 menu:
   intro:
-    parent: vs
+    parent: compare-pulumi-to
     weight: 1
 
-aliases: ["/docs/reference/vs/terraform/"]
+aliases:
+- /docs/reference/vs/terraform/
+- /docs/intro/vs/terraform/
 ---
 
 Terraform and Pulumi hold a lot of similarities, but they differ in a few key ways. This page helps provide a rundown of the differences. First, Pulumi is like Terraform, in that you create, deploy, and manage [infrastructure as code]({{< relref "/what-is/what-is-infrastructure-as-code" >}}) on any cloud. But where
--- a/themes/default/content/docs/intro/vs/chef-puppet-etc.md
+++ b/themes/default/content/docs/intro/vs/chef-puppet-etc.md
@@ -1,13 +1,12 @@
 ---
-title: Compare Pulumi to Chef, Puppet, etc.
+title: Pulumi vs. to Chef, Puppet, etc.
 meta_desc: This page gives an overview of the major differences between Pulumi and
            configuration management tools (Chef, Puppet, Ansible, Salt, etc.).
 linktitle: Chef, Puppet, etc.
 menu:
   intro:
-    parent: compare-pulumi-to
+    parent: vs
     weight: 7
-
 aliases:
 - /docs/reference/vs/chef_puppet_etc/
 - /docs/intro/vs/chef_puppet_etc/
--- a/themes/default/content/docs/intro/vs/chef-puppet-etc.md
+++ b/themes/default/content/docs/intro/vs/chef-puppet-etc.md
@@ -1,5 +1,5 @@
 ---
-title: Pulumi vs. to Chef, Puppet, etc.
+title: Pulumi vs. Chef, Puppet, etc.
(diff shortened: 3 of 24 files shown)


susanev:
@cnunciato can you remind me how to run the link checker locally, so i don't end up with issues once it gets to docs?

cnunciato:
@susanev The link checker runs against a fully built site, so if you wanted to run it against a PR like this one, you'd need to clone the docs repo, update the local pulumi-hugo reference to point to your PR branch, and run a full build:

```
cd docs
git pull origin master
hugo mod get github.com/pulumi/pulumi-hugo@se/fix-other-page-names
make clean ensure build
```

That said, a full build also uses `relref`s, which will fail on bad links, so if your build passes _this_ step, you're good to go in terms of mergeability; you don't have to run the link checker explicitly (and indeed we don't do that for docs-site PRs, because it just takes a long time to complete). 

To save you the time of actually doing this, I went ahead and ran the build step for ya:

```
                   |  EN
-------------------+--------
  Pages            | 12678
  Paginator pages  |    71
  Non-page files   |   726
  Static files     |   889
  Processed images |     0
  Aliases          |   440
  Sitemaps         |     1
  Cleaned          |     0

Total in 238663 ms
Done!
``` 

... so you're good to go as of now in terms of links working and mergeability.

themes/default/content/docs/intro/compare-pulumi-to/terraform.md
```diff
@@ -1,13 +1,15 @@
 ---
-title: Pulumi vs. Terraform
+title: Compare Pulumi to Terraform
```
zchase, line 2: We are at least going to want to leave this page as-is because we rank really well for the query "pulumi vs terraform" and we are going to want to maintain that ranking.

If we are really motivated to change the this to be "compare" instead of "vs" then we should keep the "vs" version of the page around so it will continue to rank well SEO wise. That means we would maintain two versions of the same page, but it would be safer SEO wise.

zchase (changes requested):

themes/default/content/docs/intro/compare-pulumi-to/terraform.md
susanev: @zchase i can leave all the page titles/urls as vs. is it okay to change the category name to `Compare Pulumi to...` or will that have a negative effect?

themes/default/content/docs/guides/continuous-delivery/create-new-cicd-integration.md
```diff
@@ -1,15 +1,15 @@
 ---
-title: Other
+title: Create New CI/CD Integration
```
praneetloke, line 2: This title might be a bit misleading since you are not actually "creating" a new CI/CD integration. This doc basically describes an escape-hatch mechanism for a user to set some custom env vars for Pulumi to associate an update with in case their CI/CD is not automatically picked up by the Pulumi CLI. I do like the idea of having a different title than "Other" but just that "Create New CI/CD Integration" means something else to me.

susanev: @praneetloke do you have any suggestions for other titles that might work better?

praneetloke: My possibly awful suggestion is "Custom CI/CD Detection"? 😁

themes/default/content/docs/intro/compare-pulumi-to/terraform.md
susanev: @zchase i changed them all back to keep their original titles and page names, minus the ones that were using underscores in the URLs

do you think its still okay that the parent name changes to the compare phrase? or will that also have a negative effect?

cc/ @infin8x

themes/default/content/docs/guides/continuous-delivery/create-new-cicd-integration.md
susanev: @praneetloke okay i read the page, instead of just the intro paragraph which is where i got the create language.. heh. ive updated that, and also tried out the name `Add Support for CI Systems`. what do you think? still misleading? if so, i will use your suggestion directly, though i think someone should maybe update the first couple paragraphs to match that.

praneetloke: Yeah sounds good. I'll connect with you tomorrow. I think we should rephrase the intro as you suggest.

themes/default/content/docs/intro/vs/chef-puppet-etc.md
```diff
@@ -1,14 +1,15 @@
 ---
-title: Pulumi vs. Chef, Puppet, etc.
+title: Pulumi vs. to Chef, Puppet, etc.
```
susanev, line 2:
```suggestion
title: Pulumi vs. Chef, Puppet, etc.
```

themes/default/content/docs/intro/vs/cloud-templates.md
```diff
@@ -1,13 +1,14 @@
 ---
-title: Pulumi vs. AWS CloudFormation and Other Templates
+title: Pulumi vs. to AWS CloudFormation and Other Templates
```
susanev, line 2:
```suggestion
title: Pulumi vs. AWS CloudFormation and Other Templates
```

themes/default/content/docs/intro/compare-pulumi-to/terraform.md
zchase: The page title will still need to be "Pulumi vs X" but the H1 tag could use the "compare" phrasing. What is important SEO wise is that our keywords are in the Title and URL (as well as the content). The H1 is more a "vanity" thing and varying the Title and H1 should have very little impact to SEO.

Changing the parent page should be fine as the SEO juice is really on a per-page basis.

zchase (approved):
LGTM --- just looks like there is that one lint error to fix.

susanev:
@cnunciato did you want to take another look at this?

praneetloke (approved):

PR #580, opened 2021-09-13
build(sass): update to dart-sass to enable m1 local builds

nimbinatus:
Requesting to update to dart-sass to enable m1 local builds (node-sass won't build on arm64 because it doesn't support the platform). It should be a simple lift-and-shift with no additional work needed according to https://sass-lang.com/blog/libsass-is-deprecated.

Marked as `do-not-merge` so there's time to chat if needed.

cc: @mattstratton @katcosgrove

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   "dependencies": {
     "concurrently": "^6.0.0",
     "js-yaml": "^3.13.1",
-    "markdownlint": "^0.17.2"
+    "markdownlint": "^0.17.2",
+    "sass":"^1.39.2"
   }
 }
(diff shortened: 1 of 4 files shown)


cnunciato:
Thanks! Will take a look today. I haven't spent much focused time with dart-sass yet, but so far, it's been good; the work we're doing on a new theme module (which uses the UI library I mentioned this morning) is using dart-sass as well, and all's gone well, so yeah, I think you're right that it should be an easy switch. (I believe the only thing I've seen so far deprecation notices around math functions, but would be good to do a little actual diffing just to be safe.)

cnunciato (approved):
Rendered diff looks good to me (only very minor changes). 👍  Removing the do-not-merge label; feel free to merge when you likd.

nimbinatus:
Yay! I'll double-check that folks are ready on the DA team, and then merge it tomorrow. Thanks so much!

mattstratton:
lgtm

PR #596, opened 2021-09-15
Migrate pulumi-hugo docs to registry

djgrove:
PR Draft so I can easily see the registry migration changeset


susanev:
@djgrove only double checking.. we arent going to migrate anything until release right? and right now ur working on copying the content over for folks to edit?

PR #600, opened 2021-09-16
adds overview blog for ces build track

katcosgrove:
Signed-off-by: Kat Cosgrove <kat.cosgrove@gmail.com>


nimbinatus (approved):
LGTM; only significant question would be whether the links need to be relrefs, but I'm guessing it's not a big deal right now.

Issue #602, opened 2021-09-16
Doc search is broken (Safari)


iwahbe:
## Steps to reproduce

1. Open Safari
2. Type something into the search box
3. Hit enter

Expected: A search window to appear center screen, and the rest of the screen greyed out. 
Actual: Only the side bar greyed out. Search didn't show up. I think they do show up in some way, cause I can click on them (hence the mouse icon change), but they are not visible. 

https://user-images.githubusercontent.com/22222529/133693093-09011d21-98ec-485d-b8eb-751f70341ae6.mov

lukehoban:
Related, in Chrome on Mac, the nav doesn't dim correctly when the search modal comes up.   There was an attempt to fix this in https://github.com/pulumi/pulumi-hugo/pull/533, but it was reverted in https://github.com/pulumi/pulumi-hugo/pull/542.

<img width="1792" alt="Screen Shot 2021-09-16 at 3 12 31 PM" src="https://user-images.githubusercontent.com/223467/133704224-b0665ad9-2e1c-4e05-a5ef-a0dfb06a3ea4.png">

susanev:
https://github.com/pulumi/pulumi-hugo/issues/602#issuecomment-921348103 is fixed in https://github.com/pulumi/pulumi-hugo/pull/603

djgrove:
Not fixed until the theme is updated in Hugo

djgrove:
Hugo uses the latest version of the Theme, so I will consider this fixed.

PR #603, opened 2021-09-17
fix docs search results and nav issue

susanev:
## overview
a community member opened up a pr to fix the issue where the top nav is still displayed when the docs search results dialog is open but it was later discovered that this prevented the menus from appearing on top.

ref: https://github.com/pulumi/pulumi-hugo/pull/533

im adding that back in, and also adding a z-index to the nav menus

@cnunciato @zchase pls let me know if this isn't the correct way to fix this

### before
before

### after
after
after-2

### DIFF
--- a/themes/default/assets/sass/_header.scss
+++ b/themes/default/assets/sass/_header.scss
@@ -43,7 +43,7 @@
 }
 
 .header-container {
-    @apply sticky z-50 bg-white py-6;
+    @apply z-50 bg-white py-6;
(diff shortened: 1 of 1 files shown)


cnunciato (approved):
This looks okay -- it still leaves what's reported in #602, and the overlay doesn't cover the Cloud Eng CTA (at least not in FF), but it's a step in the right direction, so 👍! (I can address those two with #602.)

image

image

PR #605, opened 2021-09-17
use svg for akamai logo instead of png

susanev:
## overview
i came across this when reviewing the registry work, we have an svg for this that will resize well, so updating to use svg instead pf png

i believe this is the only page that was using this logo `/docs/intro/cloud-providers/akamai/`


susanev:
@cnunciato this logo is also in use on this page https://www.pulumi.com/docs/reference/pkg/#api-reference
i cant figure out if i need to update it somewhere else, since that's not part of the preview
could you let me know? help me out?

cnunciato (approved):
Looks great. 👍 

Also yes, you're right, this logo is used on the API reference page. I reviewed and it looks fine there, too (local vs. www):

image

The way I check this is just by pasting some content from docs into the corresponding location in pulumi-hugo. So for this, I just copied the docs/reference/pkg/akamai folder, pasted it into themes/default/content/docs/reference/pkg/ folder in pulumi-hugo, and reloaded.

PR #622, opened 2021-09-22
Fix paths for all outdated refs so they will not point to redirect

djgrove:
This takes care of the other half of work for https://github.com/pulumi/pulumi-service/issues/7347.

(please see that private issue for full context)


praneetloke:
@djgrove shall we add the `do-not-merge` label to this until we are actually ready to merge this?

djgrove:
@praneetloke I've just got this set up to merge to the `reg-staging` branch which would serve as our mainline for this work. I can open a PR for that branch to master w/ the do-not-merge tag... is that what you have in mind?

cnunciato (approved):
LGTM, and since we have the `aliases` for these providers set up [here](https://github.com/pulumi/registry/tree/master/themes/default/content/registry/packages), we should be good in terms of 301s for this batch. Right?

praneetloke:
@djgrove please ignore what I said. I see that you are merging into `reg-staging`. I looked at the changes but had missed looking at the target branch. This is fine to merge.

djgrove:
> LGTM, and since we have the `aliases` for these providers set up [here](https://github.com/pulumi/registry/tree/master/themes/default/content/registry/packages), we should be good in terms of 301s for this batch. Right?

Yep! Because these are refs--and thus Hugo controlled--rather than basic hyperlinks, we cannot actually build the docs site without first fixing these or else the build will fail with path not found errors.

Issue #617, opened 2021-09-21
nav changes for learn pulumi release


susanev:
_note this issue is only for the learn changes not the registry changes and is part of #592_

## overview
we need to make a few changes to our top navigation
this was a collaborative effort with @infin8x jay and @susanev 

## design details
- [ ] change the learn menu to be `Blog`, `Learn Pulumi`, `Tutortials`, `Events & Workshops`
    - [ ] link to the new learn platform in the learn pulumi item
    - [ ] rename `Resources` to `Events & Workshops`
    - [ ] change the page name of `Resources` to `Events & Workshops`
    - [ ] optional but would be nice to update the learn platform URL to `learn-pulumi`
    - [ ] note tutorials stays for now, will get removed in #592
- [ ] note the desktop and mobile navs are defined in separate code pieces, pls update both versions of the nav

### screenshot
<img width="228" alt="learn-nav" src="https://user-images.githubusercontent.com/5489125/134218725-070f91c7-e0c5-4d3c-8502-51b81ada60a3.png">

PR #625, opened 2021-09-22
Move resources to events-workshops

praneetloke:
You should be able to review each of these in the individual commits:

* Added the alias for `/resources` in `events-workshops/_index.md`
* Updated references to the marketing `/resources` (not the `/docs/intros/concepts/resources`) to `/events-workshops`
* Updated the marketing nav template (`header.html` partial) to point to `/events-workshops` instead of `/resources`


susanev:
it was intended to make this change when we released the learn platform as defined in https://github.com/pulumi/pulumi-hugo/issues/617 though i don't feel super strongly about holding. @infin8x do you have opinions?

praneetloke:
@susanev yeah I misunderstood the timing of the change. I'll defer to you all on whether you want to combine this PR with something else.

infin8x:
Ah, yeah, I'd expected to see this roll out alongside the Learn Pulumi launch. But, I don't see any harm in rolling it out now - means that the Learn Pulumi gets a little simpler since we only need to add a new item, not also make these changes. If @susanev and @praneetloke are good with these changes going out now, I am too 🚀.

praneetloke:
@infin8x I am good with it. To be clear, I worked on this with a different frame of mind and wasn't aware of the Learn Pulumi stuff. I was just breaking down another piece of work and thought this would be good to do separately than club with another thing I am working on.

infin8x:
Sounds good - let's ship it if @zchase is 👍🏻

zchase (commented):
Do the resource content files need `aliases`? We have links out in the wild that point to these pages and I don't think I saw anything in the PR that would correctly redirect those links to right place.

praneetloke:
> Do the resource content files need `aliases`? We have links out in the wild that point to these pages and I don't think I saw anything in the PR that would correctly redirect those links to right place.

Hmm I hadn't thought of that. I came across two that already had aliases for typos in the title but yeah the content would need aliases with the `/resources` prefix. I'll update the PR.

praneetloke:
@zchase could you please have a look at the commit e775c92851db7e644652e2d646326e56ef8831c5 for the redirects and let me know if it looks good? I also added a silly little tool to help with adding aliases for content in bulk.

praneetloke:
Ping

susanev (commented):
looks good to me, but id like Zack to review before merging

one thing i noticed is that the all filter when you navigate to events and resources is not highlighted so maybe the link should be to `events-workshops/#all` .. this is my fault since i said events-workshops without the anchor, this is my fault

themes/default/content/events-workshops/pulumi-update-2020-04-29/index.md
```diff
@@ -8,10 +8,11 @@ meta_image: "/images/webinar/pulumi-update-twitter-card.png"
...
 aliases:
-    - /webinars/pulumi-v2-virtual-launch
-    - /events/launch-v2-seattle-2020-04-15
-    - /webinars/pulumi-update-2020-04-22
-    - /webinars/pulumi-update-2020-04-29
+  - /resources/pulumi-update-2020-04-29
+  - /webinars/pulumi-v2-virtual-launch
+  - /events/launch-v2-seattle-2020-04-15
+  - /webinars/pulumi-update-2020-04-22
+  - /webinars/pulumi-update-2020-04-29
```
praneetloke, line 15: I changed the indentation here to match what we are using for lists. Calling this out to not miss the addition of the alias `/resources/pulumi-update-2020-04-29` as the first item here.

praneetloke:
@susanev 

> one thing i noticed is that the all filter when you navigate to events and workshops is not highlighted so maybe the link should be to `events-workshops/#all` .. this is my fault since i said events-workshops without the anchor, sry about that

Good catch! I have updated the links in the `header.html` partial in the latest commit to reflect that.

zchase (approved):
LGTM

PR #647, opened 2021-09-29
Add date to meta info for CES pages

mattstratton:
I noticed that in open graph previews (slack, etc) the date of the event isn't in text, and thought it might be helpful to add that.


zchase (commented):
The meta description with the current title is 161 characters which won't pass the SEO linter. It is one character over as meta_desc are capped at 160 characters.

mattstratton:
I thought I fixed it; the linter stopped it on the first push but then I trimmed it and it passed. 

Let me WC it again.

mattstratton:
it's 160 characters according to my count

image

zchase (approved):

Issue #648, opened 2021-09-29
Add Cloud Engineering Summit CTA to homepage


lindsaymarolich:
Can we please add a more prominent CTA for the upcoming Cloud Engineering Summit to the home page? I've added a [suggested style here](https://drive.google.com/file/d/18voNf2dS7oxI-yScsyQ8VsMYCS-RM_BM/view?usp=sharing), but definitely open to other options. We may want to do a banner takeover in the week or two leading up to the summit as well.

susanev:
this looks fine to me!

PR #650, opened 2021-09-29
Add secondary CTA for summit to home hero

zchase:
Fixes: https://github.com/pulumi/pulumi-hugo/issues/648


susanev:
@zchase did the side-by-side not work for some reason?

isaac-pulumi (approved):
This addresses most of my concern -- the agenda CTA on wider screens still might siphon off clicks but the mobile version should work the way we need it to.

PR #655, opened 2021-09-30
Center the CES button

cnunciato:
Happened to notice this while fixing another bug -- little thing, but my eyes couldn't look past it. 😂

### DIFF
--- a/themes/default/layouts/index.html
+++ b/themes/default/layouts/index.html
@@ -15,7 +15,7 @@
                     <a class="home-hero-btn-primary" href="{{ relref . "/docs/get-started" }}">{{ .Params.hero.cta_text }}</a>
                 </div>
                 <div>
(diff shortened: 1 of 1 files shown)


susanev:
@zchase i had a question on this work before it merged that i think got overlooked. @lindsaymarolich's designs had these side-by-side not stacked, was there a reason we couldn't do that?

nimbinatus (commented):
Other than @susanev's note (hence comment v approve), LGTM

zchase (approved):

zchase:
@susanev the primary button in the hero section has some custom CSS attached and a secondary button side by side looked awkward. I opted to stack them as that didn't require a CSS change as that was the easiest path forward and we needed to add the button before today. Happy to revisit.

PR #654, opened 2021-09-30
Add a Snowflake logo for the cloud-provider page

cnunciato:
Link checker caught this (missing image).

PR #656, opened 2021-09-30
fix snowflake logo

susanev:
follow-on to #654 

fix the snowflake logo so it does not have whitespace built into the svg
@cnunciato im not sure what is going on with these logos or how you are creating them, but we really need to strip any whitespace around them so they will work correctly in registry


praneetloke (approved):

cnunciato:
Excellent, thanks folks!

PR #657, opened 2021-09-30
add notes re: local teams

MitchellGerdisch:


chrsmith (approved):
Thanks for the update here @MitchellGerdisch . Not sure if I like the terminology "Pulumi-local" teams. But then again, I can't think of a better way to describe them as separate from "GitHub-based teams", like we have on https://www.pulumi.com/docs/intro/console/teams/.

PR #660, opened 2021-09-30
Update newsroom for AWS Native

gthuang:
Add press release for AWS Native

### DIFF
--- a/themes/default/data/newsroom.toml
+++ b/themes/default/data/newsroom.toml
@@ -1,4 +1,10 @@
 [[releases]]
+date = "30th September 2021"
+title = "Pulumi AWS Native Provider Speeds Users Into the Future of Modern Cloud Application Development on AWS"
+summary = "Pulumi now provides same-day support for new AWS Cloud Control API-supported services and features, allowing customers to harness modern cloud architectures using industry standard languages, software tools, and cloud engineering best practices"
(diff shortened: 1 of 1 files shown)


cnunciato (approved):
Looks good, but a couple of little things worth mentioning: 

The PR itself has a page title that looks like it goes with something else:

image

Other thing I happened to notice is that the the Recent News section of the Newsroom could probably use some refreshing:

image

These are all under `[[ coverage ]]` (underneath `[[ releases ]]`), starting here: https://github.com/pulumi/pulumi-hugo/blob/8af5768fe82bbad15b39f927ba1b0e9911cba0f2/themes/default/data/newsroom.toml#L91

gthuang:
> Looks good, but a couple of little things worth mentioning:
> 
> The PR itself has a page title that looks like it goes with something else:
> 
> image
> 
> Other thing I happened to notice is that the the Recent News section of the Newsroom could probably use some refreshing:
> 
> image
> 
> These are all under `[[ coverage ]]` (underneath `[[ releases ]]`), starting here:
> 
> https://github.com/pulumi/pulumi-hugo/blob/8af5768fe82bbad15b39f927ba1b0e9911cba0f2/themes/default/data/newsroom.toml#L91

thanks for calling that out christian, we'll fix

PR #671, opened 2021-10-01
Revert "Move resources to events-workshops (#625)"

infin8x:
This reverts commit cc7cf448295940b960a657590efb3722623a2579.

PR #679, opened 2021-10-04
Update the top nav with Learn Pulumi link

zchase:
Part of https://github.com/pulumi/pulumi-hugo/issues/617

This PR updates the top nav as defined in the above issue. What this PR doesn't do is update the actual URLs from `/resources/*` -> `/events-and-workshops/*`. This is intentional as the last attempt at the updating the URLs ran into an issue and had to be [reverted](https://github.com/pulumi/pulumi-hugo/pull/671#issuecomment-932447881).

I broke out the cosmetic changes from the URL changes so that we will reduce the risk of hitting an issue when all this merges. I think updating the URLs can happen right after we ship the main chunk of the Learn Pulumi stuff. That way we can reduce this risk of hitting a deployment issue during the actual release of the Learn content.


susanev:
@zchase something went weird with this pr, there are a whole bunch of extra commits in it

zchase:
@susanev Alright I fixed up the branch. Git is hard 😂

susanev:
@zchase do you know why this never got a preview link?

susanev:
so i guess its cause ur not merging into main, ill try it out locally

themes/default/layouts/partials/header.html
```diff
@@ -123,9 +123,9 @@
 
                         <li class="mobile-menu-item"><a href="#">Learn</a></li>
                         <li class="mobile-menu-subitem"><a href="{{ relref . "/blog" }}">Blog</a></li>
-                        <li class="mobile-menu-subitem"><a href="{{ relref . "/resources" }}">Resources</a></li>
+                        <li class="mobile-menu-subitem"><a href="{{ relref . "/learn" }}">Learn Pulumi</a></li>
                         <li class="mobile-menu-subitem"><a href="{{ relref . "/docs/tutorials" }}">Tutorials</a></li>
-                        <li class="mobile-menu-subitem"><a href="{{ relref . "/resources#upcoming" }}">Workshops</a></li>
+                        <li class="mobile-menu-subitem"><a href="{{ relref . "/resources/#upcoming" }}">Events & Workshops</a></li>
```
susanev, line 128: do we want to change this to all? or do you think upcoming still makes sense?

susanev (approved):

themes/default/layouts/partials/header.html
zchase: I think upcoming still makes sense for now as those are where we want people to convert. 

I do want to follow up with some layout formats to the Events & Workshops page as I think we can do a lot of things we can do to improve the experience there.

susanev: k, sounds good

PR #681, opened 2021-10-04
Update GCP examples to match current API

aureq:
Update 2 GCP examples as the property `registry.repositoryUrl`  doesn't exist anymore. Update the code according to
`gcp.container.getRegistryRepository()` [example](https://www.pulumi.com/docs/reference/pkg/gcp/container/getregistryrepository/).

This was also mentioned to me via this SO [question](https://stackoverflow.com/questions/69413119/deploying-helloworld-docker-images-to-gke-with-pulumi/69420339#69420339).

### DIFF
--- a/themes/default/content/docs/guides/crosswalk/kubernetes/apps.md
+++ b/themes/default/content/docs/guides/crosswalk/kubernetes/apps.md
@@ -315,11 +315,14 @@ import * as pulumi from "@pulumi/pulumi";
 // Get the GCP project registry repository.
 const registry = gcp.container.getRegistryRepository();
 
+// Get the repository URL
+const repositoryUrl = registry.then(_r => _r.repositoryUrl);
+
 // Build a Docker image from a local Dockerfile context in the
 // './node-app' directory, and push it to the registry.
 const customImage = "node-app";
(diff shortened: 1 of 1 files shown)


komalali (commented):
I find this very surprising. You should be able to use `registry.repositoryUrl` directly in the pulumi.interpolate and if you're not being able to then that is a bug. Have you tested this yourself?

aureq:
Using the `interpolate`, this is the error I'm getting
```
error: Running program '/root/work/pulumi-demos/demos/gcp-ts-gcr' failed with an unhandled exception:
    TSError: ⨯ Unable to compile TypeScript:
    index.ts(12,50): error TS2339: Property 'repositoryUrl' does not exist on type 'Promise<GetRegistryRepositoryResult>'.
```

komalali:
Ah right okay, this is because of https://github.com/pulumi/pulumi/issues/5758 - I expected the function to return an `Output` but it returns a `Promise`. I also expected `pulumi.interpolate` to be able to unwrap promises. Which it appears that it does not. 🤔

komalali (approved):

aureq:
@susanev Good to go. Thank you.

PR #697, opened 2021-10-07
add audit logs export to pricing page

susanev:
adding the audit logs export to pricing page
it is available to select enterprise customers, so we are listing it as available


susanev:
there's some discussion of whether we want this, so added the do-not-merge label

PR #698, opened 2021-10-07
Add package icons

infin8x:


djgrove (commented):
Sounds like we want to adjust whitespace https://pulumi.slack.com/archives/C01QV7KBZN0/p1633650444340300

PR #704, opened 2021-10-08
Remove CSS and JS, as it's been moved to pulumi/theme

cnunciato:
Since [pulumi/theme is the new home of all things CSS and JS for the website](https://github.com/pulumi/theme), we can remove this code, as it's no longer used.


cnunciato:
I'll make sure any recent pulumi-hugo changes (e.g, the gradient banner, the darker-blue notes treatment) are accounted for in the theme repo as well.

praneetloke:
> Since pulumi/theme is the new home of all things CSS and JS for the website, we can remove this code, as it's no longer used.

Is it no longer used in `master` as well?

cnunciato:
@susanev Yep, these changes just add the CSS and JS that supports the newly updated banner; it doesn't change any of the templates that show (or don't show) that batter. So should be okay.

PR #705, opened 2021-10-09
Add theme for local dev, use the theme-assets partial, fix cli font size

cnunciato:
Intended to accompany https://github.com/pulumi/theme/pull/23, this change sets up this repository to consume pulumi/theme for its CSS and JS assets, and to use that module's `assets` partial to load the CSS and JS bundles into the `head` partial that's ultimately used to build the website. Also removes the `scripts` partial as it's no longer necessary.


themes/default/layouts/partials/head.html
```diff
@@ -3,17 +3,16 @@
...
+    {{ partial "assets" }}
+
     <!-- Google Site Verification for Google Webmaster tools and YouTube. -->
     <meta name="google-site-verification" content="N-ezSTIu4P3bSc4TqidV4wWCkMzFiMN269ZgDYArGkk" />
 
     {{ if .Params.redirect_to }}
         <meta http-equiv="refresh" content="0; url={{ .Params.redirect_to }}" />
     {{ end }}
 
-    {{ if eq .Params.block_external_search_index true }}
```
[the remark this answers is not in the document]
cnunciato: Oh, that was a mistake! Good catch.

themes/default/layouts/partials/head.html
```diff
@@ -144,9 +144,6 @@
         </script>
     {{ end }}
 
-    <!-- Preload the component JavaScript to help our Google Page Speed score. -->
-    <link rel="preload" href="/js/components/components.esm.js" as="script" crossorigin="anonymous">
```
praneetloke, line 148: I don't see this in the `assets` partial in https://github.com/pulumi/theme/pull/23. Is this also an accidental deletion?

praneetloke: Note that this is a preload, so maybe this is not needed anymore with the new structure?

cnunciato: No this is intentional — we no longer load Stencil separately this way; it gets bundled into theme/assets/js/bundle.js.

praneetloke (approved):

PR #708, opened 2021-10-11
Update robots.txt layout to allow Swiftbot in non-prod envs

praneetloke:
Also noticed that the `theme` module pseudo-version was pointing to a non-existent one (maybe it was valid at one point). So I updated it to the latest.

Note this is being merged into `reg-staging`.


themes/default/layouts/robots.txt
```diff
@@ -1,3 +1,11 @@
+{{- if ne hugo.Environment "production" }}
```
cnunciato, line 1: @praneetloke What is the implication of allowing non-prod sites to be crawled like this, though? Are we creating a new index (or "domain") just for this new content? If not, can we ensure the base URLs will be right (i.e., "https://www.pulumi.com/" as opposed to "https://some-other-host/") and not point to our temporary staging URLs? Here's what I see when I look at the rendered Swiftype results, for example:

image

Mainly just want to be sure we won't be spilling links to pre-prod environments into prod-facing results.

PR #719, opened 2021-10-12
Include C# examples in methods blog post

justinvp:
While we don't yet support authoring multi-language components in .NET, we do support using them from .NET. Since the blog post mentions being able to use methods from all supported Pulumi languages, I wanted to show usage from .NET.

This change adds back the C# chooser tabs for usage examples, and adds authoring snippets for C# so that the choosers on the page work correctly. The C# authoring snippets just mention that C# authoring isn't supported yet, with a link to the GH issue for upvoting.


emiliza (approved):
I like the change, and added reference to an issue the community can track & upvote. Thanks for correcting this, @justinvp !

PR #721, opened 2021-10-12
Add another meta tag to add a title prefix for other types of pages

praneetloke:


themes/default/layouts/partials/head.html
```diff
@@ -81,6 +81,8 @@
         <!-- For blog posts that come up in Swiftype result sets, denote that the links are blog posts by adding a "Blog" prefix to the title. -->
         {{ if and (eq .Type "blog") .IsPage }}
             <meta class="swiftype" name="title" data-type="string" content="Blog: {{ .Title }}" />
+        {{ else if and (eq .Type "registry") .IsPage }}
```
praneetloke, line 84: Per Hugo's docs, the default `.Type` value for a page is its parent folder that is directly under `content`. https://gohugo.io/content-management/types/.

> Hugo resolves the content type from either the type in front matter or, if not set, the first directory in the file path.

cnunciato (approved):
Sorry this slipped by me earlier. :shipit:

PR #742, opened 2021-10-15
Add Registry blog post

infin8x:


themes/default/content/blog/introducing-pulumi-registry/index.md
```diff
@@ -0,0 +1,333 @@
+---
+title: "Introducing Pulumi Registry: your window to the cloud"
+date: 2021-10-18T00:00:00-07:00
```
cnunciato, line 3: I think this should be okay, since we'll build the preview the day before (which should show it, since we include future-dated posts in docs PR builds), and then by the time we merge, it'll be past midnight Pacific time Monday. Sound right?

infin8x: Ah, technically this shouldn't go live until after the embargo. Perhaps it's better for me to combine this with @susanev's PR (https://github.com/pulumi/pulumi-hugo/pull/743) against `master`, in that case? Then both the PR and the blog post can go live at the same time (at/near 5am Pacific)?

susanev: let me dm you about this

infin8x: Actually, sounds like we're OK for this to go out Sunday night/early Monday. So, merging.

PR #786, opened 2021-10-21
Add company values to the about page

joeduffy:
We've been discussing this for a while now but understandably haven't been able to carve out the time due to lots of other priorities. My design skills are quite poor but I think it's an okay start.

image

I also slightly editorialized the values so they fit more cleanly on the page (we probably want to edit the internal doc down a little bit too, as it's a tad more verbose than it needs to be.)

### DIFF
--- a/themes/default/layouts/page/about.html
+++ b/themes/default/layouts/page/about.html
@@ -1,25 +1,173 @@
 {{ define "hero" }}
-    {{ partial "hero" (dict "title" "About Us") }}
+    {{ partial "hero" (dict "title" "Program the Cloud") }}
 {{ end }}
 
 {{ define "main" }}
     <section class="max-w-4xl container mx-auto py-8 px-4 text-center text-lg">
         <p>
-            Pulumi is a venture backed startup in Seattle whose mission is to enable every
-            person to harness the power of the cloud. Our team is a diverse and talented group from all
-            corners of our industry, and our products sit uniquely at the intersection of
-            developer tools and cloud operations.
-        </p>
-        <p>
-            Get your code to the cloud. Faster. Together.
+            We help developers and infrastructure teams work better together to build
+            modern cloud software.
         </p>
         <p class="text-3xl">
             <a href="{{ relref . "/careers" }}">We are hiring!</a>
         </p>
     </section>
 
+    <section id="values" class="px-4 pb-8">
+        <div class="container mx-auto">
+            <h2 class="mb-8 text-center">Our Values</h2>
+            <div class="lg:flex ml-16 mr-16">
+                <div class="lg:w-1/4 p-5">
+                    <div class="card bg-white max-w-md mx-auto">
+                        <div class="px-8 py-8">
+                            <div class="icon-section pb-4">
+                                <i class="fas fa-users text-violet-600 text-3xl"></i>
+                            </div>
+                            <div>
+                                <h6>Work as a Team</h6>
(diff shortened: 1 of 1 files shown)


themes/default/layouts/page/about.html
```diff
@@ -1,25 +1,173 @@
...
+                        <div class="px-8 py-8">
+                            <div class="icon-section pb-4">
+                                <i class="fas fa-users text-violet-600 text-3xl"></i>
+                            </div>
+                            <div>
+                                <h6>Work as a Team</h6>
+                            </div>
+                            <div class="description">
+                                <p class="text-sm">
+                                    We believe that working together is critical to our success
```
susanev, line 31: i think this should be rephrased into what we do instead of what we believe. something like "we work together..."

themes/default/layouts/page/about.html
```diff
@@ -1,25 +1,173 @@
...
+                            <div class="icon-section pb-4">
+                                <i class="fas fa-heart text-salmon-600 text-3xl"></i>
+                            </div>
+                            <div>
+                                <h6>Be Inclusive</h6>
+                            </div>
+                            <div class="description">
+                                <p class="text-sm">
+                                    We are a global team and believe an inclusive and diverse
+                                    set of perspectives is essential to accomplishing our mission.
```
susanev, line 50:
```suggestion
                                    We are a global and inclusive team with a diverse
                                    set of perspectives.
```

themes/default/layouts/page/about.html
```diff
@@ -1,25 +1,173 @@
...
+                            <div class="icon-section pb-4">
+                                <i class="fas fa-running text-blue-600 text-3xl"></i>
+                            </div>
+                            <div>
+                                <h6>Bias for Action</h6>
+                            </div>
+                            <div class="description">
+                                <p class="text-sm">
+                                    We say yes more than we say no. Good today is better than
+                                    perfect tomorrow. When in doubt, experiment and fail fast.
```
susanev, line 106: this is maybe a separate convo, but fail fast has a negative connotation now with some folks so we may want to figure out different phrasing. i do think this work though just ending on experiment though.

susanev (approved):
i dont feel strongly about my suggested edits to the values, so feel free to only use what you agree with.

a few recommendations...

- [ ] we have been trying to standardize the nav items matching the page titles; this helps folks confirm they navigated to the place they intended, so id rec keeping the current title of about us
- [ ] the 4th piece of content in the second row looks really out of place, id recommend adding it as a subhead to the title instead
- [ ] id love to use icons that family better with the visual design on our website, no need to block though, i can follow up with improvements there in a separate pr

joeduffy:
@susanev I made some changes, including incorporating your suggestions to the descriptions, and making the title "About Us." I don't have a ton of time to work on other design elements but this feels like strictly a much better place to end up for now. Are you okay merging as-is and we can come back and do an overhaul when you and Casie are able to spend a bit more effort on overall design and flow? I know we can do better, but I would love to lead with our values on this page as soon as we can! :-)

susanev:
yes absolutely pls feel free to merge, thats why i approved when i left the comments!

joeduffy:
Oh, gotcha, sounds good! :+1: I always feel bad merging without addressing all feedback, but I'm now realizing almost two weeks later I might have bitten off more than I can chew. Appreciate the support merging this in 😄

PR #791, opened 2021-10-25
Add team stack access apis

MitchellGerdisch:
Added API docs for Granting and removing team stack access.


susanev:
@MitchellGerdisch are we good to merge this? or were you wanting to add more?

PR #800, opened 2021-10-27
Fixed Swiftype Search Modal overlay on Safari

### LINKED ISSUE #602: Doc search is broken (Safari)

djgrove:
Safari appears to have slightly different interpretations about [z-index stacking](https://philipwalton.com/articles/what-no-one-told-you-about-z-index/) from Chromium based browsers. I'm not exactly sure what that distinction is (probably comes into play because the positioning of modals is typically done with `absolute` or `relative` positioning), but the implication is that if we render a modal that we want displayed _appearing_ at "root" level, it needs to be at that place in the DOM tree for Safari to render it how we want.

This bug occurred because creating a div with an ID of `#search-results` will override Swiftype default behavior which manages injecting the search results modal when you submit your search. By default, Swiftype injects at the first level below the `body` tag in the DOM tree, which is what we want.

Since we weren't leveraging anything, I removed the div that was overriding Swiftype and this fixed the issue in Safari (things still work in Chrome and Firefox)

## Issue
Fixes #602

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.16
 
 require (
 	github.com/pulumi/pulumi-hugo/themes/default v0.0.0-20211008162151-6e65a2068c3b // indirect
-	github.com/pulumi/theme v0.0.0-20211025172707-d60f9ad59328 // indirect
+	github.com/pulumi/theme v0.0.0-20211101174505-358f8e2823f5 // indirect
 )
 
 replace github.com/pulumi/pulumi-hugo/themes/default => ./themes/default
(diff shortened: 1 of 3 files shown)


susanev:
@djgrove the get started button and the chat widget are appearing on top of the overlay when the search dialog is open

docs-search-z-index-issues

djgrove:
Updated the theme, as this is the fix we really needed. @susanev this appears to have resolved the issue you called out. Hopefully the demo site will be built by the time you see this, does this look good to you?

cnunciato:
The theme change has been pulled into https://github.com/pulumi/docs/pull/6732 as well, so it can be previewed there, too. (Which I think means we can close this PR without merging, since this repo should always pick up the latest from pulumi/theme.)

PR #804, opened 2021-11-02
Update min Go version

dtamai:
I go an error with the message

> embed: package embed is not in GOROOT

when using Pulumi with Golang version 1.15.

Since `embed` was introduced in version 1.16 I believe this is the minimum version, updating to this version did solve the problem with embed.


susanev:
thank you for submitting the pr @dtamai!

@stack72 i added you as a reviewer since that's what github suggested, but pls remove urself if that's not correct, thank you!

PR #805, opened 2021-11-02
Update the Team page

cnunciato:
For an auto-assigned offboarding task.


susanev (approved):
not that i want to, but should we remove Komal as well?

PR #807, opened 2021-11-02
remove quote thats not a quote from shortcode

susanev:
## overview
- we should not be using quote styling for things that are not quotes, this removes it and uses the note styling instead
- note that we are repeating the note definition because you cannot use a shortcode inside of another shortcode definition

note to reviewers: i don't believe you will be able to preview this anywhere until it gets to the docs

### DIFF
--- a/themes/default/layouts/shortcodes/aws-resource-note.html
+++ b/themes/default/layouts/shortcodes/aws-resource-note.html
@@ -1,7 +1,11 @@
-<blockquote>
-    <p>
-        If you do not already have an AWS account, you can <a href="https://aws.amazon.com/free/">create a free account</a>.  Most
+<div class="note note-info" role="alert">
+    <div class="note-heading">
+        <i class="fas fa-info-circle mr-1"></i>
+        <span>Note</span>
+    </div>
+    <div class="note-body">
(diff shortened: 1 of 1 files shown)


susanev:
@djgrove would you be up for adopting this pr and trying it out?

cnunciato:
Yep, best of my knowledge, that's right @djgrove. You can do this:

In `some-markdown-file.md`:
```
...
{{< some-shortcode >}}
    {{< some-other-shortcode >}}
{{< /some-shortcode >}}
...
```

.. but you can't do this:

In `some-shortcode-html`:
```
...
{{< some-other-shortcode >}}
...
```

So IMO, what susan has here is 👍 .

cnunciato:
> i don't believe you will be able to preview this anywhere until it gets to the docs

@susanev I thought about maybe adding a reference to this repo's config to make it possible to see registry content in it, but just never got around to it. Would that be helpful? If so, it'd be a pretty quick thing to add.

cnunciato (approved):
LGTM! Screenshot of what I see locally:

image

susanev:
@cnunciato it would be useful to me, but we should maybe wait until it becomes useful to more than just me

cnunciato:
Good to know. It'd actually be useful to me too -- I always grumble when I navigate to /registry and don't see anything useful. So maybe I'll throw something together for it. Thanks!

PR #811, opened 2021-11-02
Add Registry for local dev and PR previews

cnunciato:
It's a bit cumbersome not to have at least some Registry content available while developing in this module. So this change just makes the Registry module available for development and PR previews.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -27,9 +27,10 @@ jobs:
       - name: Check out branch
         uses: actions/checkout@v2
 
-      - name: Fetch the latest pulumi/theme
+      - name: Fetch the latest pulumi/theme and pulumi/registry
         run: |
-          hugo mod get github.com/pulumi/theme
+          hugo mod get github.com/pulumi/theme@release
+          hugo mod get github.com/pulumi/registry/themes/default
           hugo mod tidy
 
       - name: Commit any changes to go.mod/go.sum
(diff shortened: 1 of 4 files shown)


susanev:
@cnunciato who would be the right person to review this?

.github/workflows/push.yml
```diff
@@ -27,9 +27,10 @@ jobs:
       - name: Check out branch
         uses: actions/checkout@v2
 
-      - name: Fetch the latest pulumi/theme
+      - name: Fetch the latest pulumi/theme and pulumi/registry
         run: |
-          hugo mod get github.com/pulumi/theme
+          hugo mod get github.com/pulumi/theme@release
```
cnunciato, line 32: This should use the `release` branch also, since that's what we use elsewhere.

config/_default/config.yml
```diff
@@ -13,6 +13,18 @@ module:
           target: static
         - source: layouts
           target: layouts
+    - path: github.com/pulumi/registry/themes/default
```
praneetloke, line 16: Will this cause a problem for the docs repo during a full site build? I ask because registry also this repo as a dep and both of these modules join during the full site build. If you've verified this won't cause a problem, then please ignore this comment.

praneetloke (approved):

config/_default/config.yml
cnunciato: Nope -- these Hugo `config` files only apply where the `hugo` command is run, so in this case, only in local and PR builds of this repo. Module consumers like pulumi/docs will never see them.

PR #818, opened 2021-11-03
Fix code in kubernetes/identity.md that does not compile

t0yv0:
Stumbled upon this one under Azure tab in:

https://www.pulumi.com/docs/guides/crosswalk/kubernetes/identity/

The guide does not pin the dependencies and I believe something might have bit-rotted. I've tested the change compiles against the latest dependencies but I've not done the due diligence here to check the entire guide and deploy it.


iwahbe (approved):
This is still the right direction. I would merge it in. Is there a framework to check that our examples compile?

pgavlin (approved):

PR #823, opened 2021-11-04
Update _index.md

MitchellGerdisch:
The Crossguard core concepts page is buried in the getting started flow.
So I added the link early in the flow. 
Also, I didn't think we should keep calling the capability "new".


susanev (approved):
yeeesss, i agree with new and worry so much when we use it in docs in general or when we say something is coming soon, :wee:. thank you for fixing this!!

PR #833, opened 2021-11-07
Include note about Go 1.16 minimum

indrora:
New users who are on older versions of Go may encounter an error with `embed` being required by newer versions of Pulumi. 

Ran into this while I was getting started with the tool -- Debian is shipping 1.15 and Ubuntu LTS ships with 1.13. This won't matter for too long -- Debian will be shipping 1.17 in `testing` soon and the next Ubuntu LTS will likely ship with an up to date version.

### DIFF
--- a/themes/default/layouts/shortcodes/install-go.html
+++ b/themes/default/layouts/shortcodes/install-go.html
@@ -11,11 +11,7 @@
         <p>
             Pulumi requires Go 1.16 or later. If you're using Linux, your distribution may not
             provide an up to date version of the Go compiler. To check what version of Go you have
-            installed, <code>go version</code> will tell you:
+            installed, use:<code>go version</code>.
         </p>
-<pre><code>
-$ go version
-go version go1.17.1 darwin/arm64
-</code></pre>
     </div>
 </div>
--- a/themes/default/layouts/shortcodes/install-go.html
+++ b/themes/default/layouts/shortcodes/install-go.html
@@ -1,3 +1,21 @@
 <p class="mt-4">
     Install <strong><a href="https://golang.org/doc/install" target="_blank">Go</a></strong>. <br/>
 </p>
+
+<div class="note note-info" role="alert">
+    <div class="note-heading">
+        <i class="fas fa-info-circle mr-1"></i>
+        <span>Note</span>
+    </div>
+    <div class="note-body">
+        <p>
+            Pulumi requires Go 1.16 or later. If you're using Linux, your distribution may not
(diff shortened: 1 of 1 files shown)


leezen:
Thanks for the suggestion. I'm not sure this change is necessary as the Go language installation page defaults to the latest stable version and includes instructions on determining the version you have installed. I worry that the note will clutter the getting started guides and we already document the supported language versions on the individual language pages.

indrora:
This change brings the Go content more in line with the Node and Python
guidance already in place.

leezen:
@indrora Thanks for the clarification.

leezen:
Let me tag a couple other folks to get their opinions as from my perspective, I'd prefer to keep these notes relatively short and not try to clutter the initial experience. If anything, I'd want to reduce the Python verbiage and keep the details to the language-based pages.

leezen:
Curious on your thoughts @susanev, @cnunciato, and @zchase

susanev:
to make sure we are all on the same page, this is going from ...
before

to
after

i agree with @indrora that this aligns go with the other languages. i hear the point about the amount of content though, i think that we should tackle that problem separately. i also in general worry about sending folks out of the getting started flow in order to get setup.

zchase:
I am fine with this change, but if we want to reduce content I think we should just drop the `code` block and just have the two lines of text. Which is what I think @susanev alluded to above as well?

Seems reasonable that when someone is picking a language during the Getting Started flow that we should let them know the minimum version needed.

themes/default/layouts/shortcodes/install-go.html
```diff
@@ -1,3 +1,21 @@
...
+    <div class="note-body">
+        <p>
+            Pulumi requires Go 1.16 or later. If you're using Linux, your distribution may not
+            provide an up to date version of the Go compiler. To check what version of Go you have
+            installed, <code>go version</code> will tell you:
+        </p>
+<pre><code>
+$ go version
+go version go1.17.1 darwin/arm64
+</code></pre>
```
susanev, line 19:
```suggestion
            provide an up to date version of the Go compiler. To check what version of Go you have
            installed, use:<code>go version</code>.
        </p>
```

susanev (approved):

themes/default/layouts/shortcodes/install-go.html
susanev: @indrora what would you think of removing the code snippet?

indrora: This looks good

PR #843, opened 2021-11-09
Add link to public roadmap blog on /product/

gthuang:
Surface the public roadmap in the "open source" section of our product page

### DIFF
--- a/themes/default/content/product/_index.md
+++ b/themes/default/content/product/_index.md
@@ -17,70 +17,70 @@ key_features:
     items:
         - description: Build [Infrastructure as Code](/what-is/what-is-infrastructure-as-code/) in familiar languages
           features:
-            - title: Languages you love
-              icon: code
-              icon_color: yellow
-              description: |
-                Use TypeScript, JavaScript, Python, Go, and .NET to model cloud infrastructure
-                by leveraging the features of each language.
+              - title: Languages you love
+                icon: code
+                icon_color: yellow
+                description: |
+                    Use TypeScript, JavaScript, Python, Go, and .NET to model cloud infrastructure
+                    by leveraging the features of each language.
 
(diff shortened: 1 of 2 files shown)


gthuang:
@cnunciato I'm having trouble getting the linking to work in here.

cnunciato:
@gthuang Ah, I see -- we just need to make a little tweak to the template to honor that Markdown linking syntax. Mind if I push a commit onto your branch?

gthuang:
@cnunciato yep please do

susanev:
im glad we are linking to this here, should we possibly consider linking to it in the footer as well?

gthuang:
@susanev where exactly do you suggest adding it?

susanev:
in the footer.

two options that i think would work well...

- third column, between contact us and support
- first column at the bottom

gthuang:
@susanev I think at the bottom of the first column makes a lot of sense to me.

cnunciato:
@gthuang Ok, should be all set. (And sorry for the spacing noise; my editor applied it.)

gthuang:
thanks @cnunciato !

gthuang:
@cnunciato so the link works but the text looks just like regular text so you can't tell there's a link there. do you know how to make it so that it's underlined (just like how it is in the rest of the page)?

cnunciato:
Ah, I see what you mean -- it's blue, which is the default. I'll take a peek at how the others on the page are defined and see if I can emulate that.

cnunciato:
Ok, just pushed up a change that should do the trick. Let me know what you think!

cnunciato (approved):
Adding an approval here so this can be merged whenever you like @gthuang.

gthuang:
@cnunciato thanks looks great!

PR #845, opened 2021-11-10
update footer link and link text; apis -> registry

susanev:
## overview

we have an apis link in the footer that went to the docs API reg list page which we will prolly be removing at some point in the future; i updated the link to send you to registry and use that link text; if anyone feels strongly that it should be `Pulumi Registry` instead of just registry pls let me know, happy to update


susanev:
note: the preview shows that weird registries page, but it should work once it gets to docs. the current registry link does the same.

infin8x (approved):
LGTM, nice catch

PR #848, opened 2021-11-12
Refresh homepage banner with Registry announcement

gthuang:
Update the banner from cloud engineering summit to the Registry.


infin8x (approved):
🚀

PR #866, opened 2021-11-17
Fix import reference to aws to be eks in eks example

bit-cloner:
sample code imports pulumi eks package as aws. I think it should be imported as eks instead. 
as later in the code eks.cluster is used


leezen:
Nice catch. Thanks for the PR!

PR #870, opened 2021-11-18
remove all caps from button text

susanev:
randomly noticed these buttons that still had all caps, this fixes it.

### DIFF
--- a/themes/default/content/docs/guides/continuous-delivery/github-actions.md
+++ b/themes/default/content/docs/guides/continuous-delivery/github-actions.md
@@ -396,11 +396,8 @@ Once the Pulumi GitHub App is installed, when your GitHub Actions run Pulumi, a 
 any resource changes will be left on the Pull Request, as well as links to the Pulumi
 Console for more detailed information.
 
-You can install the Pulumi GitHub App now, by visiting
-[github.com/apps/pulumi](https://github.com/apps/pulumi).
-
 <a class="btn" href="https://github.com/apps/pulumi" target="_blank">
-    Install
+    Install Pulumi GitHub App
 </a>
 
 Example comment when using the Pulumi GitHub App:
--- a/themes/default/content/docs/guides/continuous-delivery/github-actions.md
+++ b/themes/default/content/docs/guides/continuous-delivery/github-actions.md
@@ -397,10 +397,10 @@ any resource changes will be left on the Pull Request, as well as links to the P
 Console for more detailed information.
 
 You can install the Pulumi GitHub App now, by visiting
-[github.com/apps/pulumi](https://github.com/apps/pulumi) or clicking the button below.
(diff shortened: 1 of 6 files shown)


cnunciato (approved):

themes/default/content/docs/guides/continuous-delivery/github-actions.md
```diff
@@ -397,10 +397,10 @@ any resource changes will be left on the Pull Request, as well as links to the P
 Console for more detailed information.
 
 You can install the Pulumi GitHub App now, by visiting
```
cnunciato, line 399:
```suggestion
You can install the Pulumi GitHub App now by visiting
```

themes/default/content/docs/guides/continuous-delivery/github-app.md
```diff
@@ -35,10 +35,10 @@ This provides even more detail about any resource changes, including the full up
 ## Installation and Configuration
 
 Pulumi's GitHub workflow integration is a GitHub application you can install by visiting
-[github.com/apps/pulumi](https://github.com/apps/pulumi) or clicking the button below.
+[github.com/apps/pulumi](https://github.com/apps/pulumi).
 
 <a class="btn" href="https://github.com/apps/pulumi" target="_blank">
-    INSTALL
+    InstalL
```
cnunciato, line 41:
```suggestion
    Install
```

themes/default/content/docs/guides/continuous-delivery/github-actions.md
```diff
@@ -397,10 +397,10 @@ any resource changes will be left on the Pull Request, as well as links to the P
...
-[github.com/apps/pulumi](https://github.com/apps/pulumi) or clicking the button below.
```
cnunciato, line 400: Actually just noticed this -- and I think I get why you're removing it ('cause directional reference), but do you think it's strange to say "install the GitHub app by going here" and then placing a button right below that that reads "Install"? I'm thinking people may be slightly confused by this (since "click the button below", directionality notwithstanding, would've at least explained that).

image

susanev: @cnunciato heh i actually had originally updated the button text to be more specific, e.g, Install Pulumi GitHub App. would you support that instead? but then i backed out.. i can do that though.

and if i do that, would you support removing the you can.. language, with direct link?

cnunciato: Yep, I was just about to suggest something just like that. Sounds perfect. 👍

themes/default/content/docs/guides/continuous-delivery/github-actions.md
susanev: this was removed

PR #875, opened 2021-11-18
[Learn] Abstraction and Encapsulation pathway

nimbinatus:
Add an intermediate pathway for abstraction, encapsulation, and component resources.


ericrudder:
nice to see the pathway!

maybe suggest "reusability" a bit more than "abstraction/encapsulation?"

i didn't suggest any direct edits, however, i think that's a really nice
benefit that would be cool to highlight even more than the current mentions
... maybe even in the title.

also worth considering a mention that it's not just for "you", it's great
if you have a platform team enabling others!

nimbinatus:
The for loop was a mistake and makes no sense in context. Please ignore it; it will get fixed tomorrow

nimbinatus:
Feedback so far:
- [ ] needs an example showing how to use the component resource from another program. like closing the loop
- [ ] encapsulation section needs a clearer explanation ("it took me awhile to understand the Encapsulation section, but i think i get it now.")

themes/default/content/learn/abstraction-encapsulation/component-resources/index.md
```diff
@@ -0,0 +1,113 @@
...
+
+We've actually already started creating a component resource in the last part
+when we made a new class that built up an `s3.Bucket` and an `s3.BucketPolicy`
+with a `define_policy` function. Let's now turn that logical grouping into an
+actual component resource.
+
+## Converting to a component resource
+
+When we're converting to a component resource, we're subclassing (or composing
+from an anonymous field, if you're into Go) the `ComponentResource` so that our
```
jkodroff, line 52: I think it's ok to ignore Go and just focus on Python since the implementation is pretty different in each language IMO.

themes/default/content/learn/abstraction-encapsulation/component-resources/index.md
```diff
@@ -0,0 +1,113 @@
...
+            policy=bucket.arn.apply(lambda arn: define_policy(policy_name)),
+            opts=pulumi.ResourceOptions(parent=bucket)
+        )
+        self.register_outputs({
+            "bucket_name": bucket.bucket_name
+        })
+```
+
+Within `super()`'s init, we pass in a name for the resource, which we recommend
+being of the form `<package>:<module>:<type>` to avoid type conflicts since it's
```
jkodroff, line 93: Suggest using more prescriptive language, e.g. "which should be"

themes/default/content/learn/abstraction-encapsulation/component-resources/index.md
```diff
@@ -0,0 +1,113 @@
...
+being registered alongside other resources like the Bucket resource we're
+calling (`aws:s3:Bucket`).
+
+That last call, `self.register_outputs({})`, passes Pulumi the expected outputs
+so Pulumi can read the results of the creation or update of a component resource
+just like any other resource, so don't forget it! You can register default
```
jkodroff, line 99:
```suggestion
just like any other resource, so be sure not to omit this step! You can register default
```

themes/default/content/learn/abstraction-encapsulation/encapsulation/index.md
```diff
@@ -0,0 +1,150 @@
...
+example.
+
+## Breaking apart Pulumi code
+
+Let's take a fairly uncomplicated piece of Pulumi code: the definition of an s3
+bucket:
+
+```python
+import pulumi
+import pulumi_aws_native as aws
```
jkodroff, line 34: Mmmmmmmaybe import `as aws_native` to make it glaringly obvious which provider is being used?

themes/default/content/learn/abstraction-encapsulation/encapsulation/index.md
```diff
@@ -0,0 +1,150 @@
...
+
+bucket = aws.s3.bucket("my-bucket")
+
+pulumi.export("bucket", bucket.bucket_name)
+```
+
+Here, we're creating one resource and exporting the output from that resource.
+Generally, however, there's a lot more to building up infrastructure with a
```
jkodroff, line 42:
```suggestion
In practice however, production systems typically have more requirements than a
```

themes/default/content/learn/abstraction-encapsulation/encapsulation/index.md
```diff
@@ -0,0 +1,150 @@
...
+                        f"{arn}/*"
+                    ]
+                }]
+            })
+        ),
+        opts=pulumi.ResourceOptions(parent=bucket)
+    )
+```
+
+But that JSON list of the policy is an object, too, that we can tuck away,
```
jkodroff, line 112:
```suggestion
But that JSON blob of the policy is also an object which we can encapsulate,
```

themes/default/content/learn/abstraction-encapsulation/encapsulation/index.md
```diff
@@ -0,0 +1,150 @@
...
+class OurBucketClass(self, name_me, policy_name):
+
+    policy_list = {
+        'default': '{...}',
+        'locked' : '{...}',
+        'permissive': '{...}'
+    }
+
+    def define_policy(policy_name, bucket_id):
+        try:
```
jkodroff, line 131: This `try` should be removed as it has no effect since the `except` block has no error handling.

themes/default/content/learn/abstraction-encapsulation/encapsulation/index.md
jkodroff: Also consider just using AWS classic as referenced below to keep everything within a single provider, since AWS Native isn't going to be RTM until 2022.

themes/default/content/learn/abstraction-encapsulation/abstraction/index.md
```diff
@@ -0,0 +1,68 @@
...
+Let's imagine the Pulumipus brand is expanding to all kinds of systems, and so
+we don't know which infrastructure each team will need. However, each team will
+need some kind of storage location that has an access policy, no matter whether
+they're doing a serverless build or a compute-based build.
+
+Mentally, we're defining a sort of abstraction like this:
+
+![A diagram showing a user doing an action that touches the storage through a policy](./abstraction.png)
+
+Seems pretty clean, doesn't it? Abstraction is thinking through what we
```
cnunciato, line 63: To be honest, I don't understand what this diagram is telling me. Are "user" and "action" _part_ of the abstraction somehow? Or are they stand-ins for the consumers of the `policy + storage` abstraction that we develop in the pages that follow? I also feel, as I reach the end of this page, like I still don't quite grok what "abstraction" means in this context, or what problem it solves; the words all make sense, and the points you've made all ring true, but somehow, I still seem to be missing the take-away.

One thing that might help would be to consider explaining encapsulation first, since that's often what you reach for before you start thinking about higher-level abstractions. When you realize you've copy-pasted the same block of code to provision a bucket and policy a dozen times, it's sometimes enough just wrap it up into a function -- `makeBucket("my-bucket-name")`, say -- to reduce all that duplication. That qualifies as encapsulation -- tucking away the repetition and exposing only the parts that need to vary -- but not really abstraction, since the concept of "bucket" still exists in the mind of the person calling the `makeBucket()` API.

From there, you could introduce abstraction to solve the problem of APIs that are too low-level. If the reason your users are creating all these buckets is to serve static websites, for example, and those websites also require provisioning additional things like domain names or other stuff, it'd probably might be much just nicer to give them a higher-level API, like `new Website("something.com")`, than to expect them to call lots of functions to create buckets, A records, TLS certs, and so on.

Does this make sense? And then finally, bring in components to solve the problem of conveying how collections of resources are related to one another -- how all these dozens of policies and things all ultimately fall under the `Website` component, for example. Anyway, let me know what you think -- I'm not married to any of this or anything, I'm just thinking of ways to help bring these points home is all.

themes/default/content/learn/abstraction-encapsulation/component-resources/index.md
```diff
@@ -0,0 +1,113 @@
...
+outputs using this call, as well. It's not hard to imagine we will always want
+the bucket name for our use case, so we pass that in as an always-given output
+for our component resource.
```
cnunciato, line 102: It'd be _really_ nice to see how this looks in an output or in the Console somehow, vs. how things look without it.

cnunciato (approved):
Nice work! Lots of little suggestions for you to consider, but nothing need block this from getting out into the world and continuing to iterate on it.

themes/default/content/learn/abstraction-encapsulation/_index.md
```diff
@@ -0,0 +1,60 @@
...
+date: 2021-11-17
+draft: true
+description: |
+    Explore abstraction and encapsulation with Pulumi programs using component
+    resources.
+meta_desc: |
+    Explore abstraction and encapsulation with Pulumi programs using component
+    resources.
+index: 6
+# meta_image: meta.png
```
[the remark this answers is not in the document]
nimbinatus: Ah, this was a holdover from when I was exploring the stuff in #911. I'll add it back in. Thanks for catching it!

themes/default/content/learn/abstraction-encapsulation/_index.md
```diff
@@ -0,0 +1,60 @@
...
+    - learn
+    - components
+providers:
+    - aws
+block_external_search_index: true
+---
+
+## Time
+
+How long this tutorial will take depends on your internet connection, reading
```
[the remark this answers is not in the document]
nimbinatus: Sorry, old holdover from another company's markdown policies; it was easier for editing at the time and we didn't have it in a style guide here (though I see that's been added since this PR went in).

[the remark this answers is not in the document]
nimbinatus: Not applying this change; this is standard language.

themes/default/content/learn/abstraction-encapsulation/_index.md
```diff
@@ -0,0 +1,60 @@
...
+## About this pathway
+
+In this pathway, we will explore how you could use Pulumi as a tool to help you
+abstract and encapsulate your resource definitions as reusable models for others
+to use.
+
+Note that this pathway will not have as many code samples to use as we will be
+defining patterns, exploring architecture, and discussing resources.
+
+<!-- This pathway is considered an advanced pathway. -->
```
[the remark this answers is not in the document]
nimbinatus: It was a holdover from when I was trying to identify levels for pathways. Thanks for catching it!

themes/default/content/learn/abstraction-encapsulation/abstraction/index.md
```diff
@@ -0,0 +1,68 @@
...
+Just like any application code, Pulumi infrastructure code can be abstracted,
+enabling us to work with models or objects, depending on which language we use
+to define that code. If we use an object-oriented langauge such as Javascript,
+Python, or Typescript, we can create and instantiate classes. For languages like
+Go, we can build up interfaces. Either way, we're thinking in terms of taking
+code that we've written and making it reusable in some form.
+
+Breaking apart Pulumi code is essential to keeping a clean, well-organized, and
+easily maintained code base that enables teams to collaborate to enhance current
+systems. As the owners of the infrastructure for the Pulumipus brand, we've got
```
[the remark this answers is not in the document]
nimbinatus: It's a fictional company that we've been using for all of the pathways. I'll fix it.

themes/default/content/learn/abstraction-encapsulation/abstraction/index.md
```diff
@@ -0,0 +1,68 @@
...
+meta_image: meta.png
+authors:
+    - laura-santamaria
+tags:
+    - learn
+block_external_search_index: false
+---
+
+Just like any application code, Pulumi infrastructure code can be abstracted,
+enabling us to work with models or objects, depending on which language we use
```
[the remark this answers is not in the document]
nimbinatus: I was going for conceptual models/modeling, specifically the idea of models in mathematics when talking about abstraction of a problem set. I realized that with `object` meaning so much more in CS, I had to find a different term. I'll find a way to reword this (and technically, `models` are the abstract concept underneath `objects` in Python 😄  Objects --> data abstractions --> conceptual models of data used to define and understand general patterns in structures. Python was originally built by someone with a masters in mathematics; a lot of the concepts are underpinned by mathematical theories, which might be why I like it so much 😆  ).

themes/default/content/learn/abstraction-encapsulation/abstraction/index.md
```diff
@@ -0,0 +1,68 @@
...
+all of the programs we maintain.
+
+Let's first explore abstraction.
+
+## Thinking in abstraction
+
+When we use our code for our Pulumi program, we're generally starting out with a
+declarative thought process. We write code that declares we want three resources
+of a specific type, in that order, and run the code to create those three
+resources in order.
```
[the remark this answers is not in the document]
nimbinatus: Good point; I wasn't thinking straight when I wrote that, clearly. I'll adjust it to talk more to the "sure, we could write code in a single file with no abstraction; that's a likely case for a proof-of-concept" idea to set us up for why they should care...

themes/default/content/learn/abstraction-encapsulation/encapsulation/index.md
nimbinatus: Good choice on switching it to `aws_native`. I've been just using `aws` in workshops and such, but I'm there to keep reminding the attendee of the import naming choice. Fixing in the next commit.

In regard to the question around the provider, we're trying to using the native provider whenever possible so folks get familiar with it. While I can see the argument, I think I'll stick with it as we have it. This part of the API is pretty stable as it's the one they're using to show off everything they're doing.

themes/default/content/learn/abstraction-encapsulation/abstraction/index.md
nimbinatus: No, this makes sense! I'm used to thinking in broad strokes and then drilling down, but that's how my brain works since I'm often trying to find patterns so I can pull out more accessible analogies (e.g., talking about state from physics to explain state in infrastructure) to explain the concepts to more people. I should have thought more about that flip when putting this together. That will also make the diagram make more sense (I was trying to diagram the S3 bucket example in the most generic form possible; it probably didn't help that I didn't use a standard architecture program to make the diagram 🤔  ). Let me rework this.

themes/default/content/learn/abstraction-encapsulation/_index.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Prerequisites
+
+You will need the following tools to complete this pathway:
+
+- A [Pulumi account and
+  token]({{< relref "/docs/intro/console/accounts#access-tokens" >}})
+  - If you don't have an account, go to the [signup
+    page](https://app.pulumi.com/signup).
+- An AWS account (the free version should be fine)
```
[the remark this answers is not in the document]
nimbinatus: I'll move it to an optional prerequisites section; some folks may still want to try things out.

themes/default/content/learn/abstraction-encapsulation/encapsulation/index.md
nimbinatus: whoops, forgot that; I'm fixing the block with better handling in the next commit

nimbinatus:
Note for future upgrades: Probably should address transformations (https://www.pulumi.com/docs/intro/concepts/resources/options/transformations/)

themes/default/content/learn/abstraction-encapsulation/component-resources/index.md
nimbinatus: I think I'm going to get this out with the next round when we've got the TS support in. I'm just not sure I'm doing it right, and there's not a lot of Python examples :/

themes/default/content/learn/abstraction-encapsulation/component-resources/index.md
```diff
@@ -0,0 +1,94 @@
...
+
+That last call in the init, `self.register_outputs({})`, passes Pulumi the expected outputs so Pulumi can read the results of the creation or update of a component resource just like any other resource, so don't forget that call! You can register default outputs using this call, as well. It's not hard to imagine we will always want the bucket name for our use case, so we pass that in as an always-given output for our component resource.
+
+From here, you can deploy it and get your custom resource appearing in the resource tree in your terminal! You can also share it with others so they can import the resource and use it without ever needing to understand all of the underlying needs of a standard storage system on AWS.
+---
+
+Congratulations! You've now finished this pathway on abstraction and encapsulation in Pulumi programs! In this pathway, you've learned about thinking of code in abstract forms, wrapping up logical groupings of code to make reuse easier, and building with component resources to make those logical groupings something that Pulumi recognizes.
+
+Go build new things, and watch this space for more learning experiences with Pulumi!
```
cnunciato, line 94: A link to more example of what this looks like would give those who want to learn more on this somewhere to go from here. What do you think?

themes/default/content/learn/abstraction-encapsulation/encapsulation/index.md
```diff
@@ -0,0 +1,150 @@
...
+layout: topic
+date: 2021-11-17
+draft: false
+description: |
+    Explore encapsulation of cloud resources with Pulumi.
+meta_desc: |
+    Explore encapsulation of cloud resources with Pulumi.
+index: 0
+estimated_time: 10
+meta_image: meta.png
```
cnunciato, line 12: Lil' reminder to update this.

cnunciato (approved):
Nice! This is an awesome improvement and feels much more understandable to me now. A few things to consider before merging, and it'd be great to add some TypeScript in there, too, but we could also just do add the additional-language stuff as a follow-up.

I will say that while there's a bunch of useful info in this pathway/lesson, it does feel like we could develop this one a _lot_ more, since it's such a rich topic (both conceptually as well as technically). I don't think we need to hold this pathway up or anything, but I do think we should consider what to do for those readers who reach the end of it and think, "Wait -- that's it?" since there really a lot more to be learned. So if possible, it'd be great to add some sort of a "Further Reading" section (maybe just a few bullets for now) that gives them somewhere to go from here.

themes/default/content/learn/abstraction-encapsulation/_index.md
```diff
@@ -0,0 +1,43 @@
+---
+title: "Abstraction and Encapsulation"
+layout: module
+date: 2021-11-17
+draft: true
```
cnunciato, line 5:
```suggestion
draft: false
```

or better, just remove all of these:

```suggestion
```

themes/default/content/learn/abstraction-encapsulation/component-resources/index.md
```diff
@@ -0,0 +1,94 @@
...
+
+
+
+```
+
+Within `super()`'s init, we pass in a name for the resource, which we recommend being of the form `<package>:<module>:<type>` to avoid type conflicts since it's being registered alongside other resources like the Bucket resource we're calling (`aws:s3:Bucket`).
+
+That last call in the init, `self.register_outputs({})`, passes Pulumi the expected outputs so Pulumi can read the results of the creation or update of a component resource just like any other resource, so don't forget that call! You can register default outputs using this call, as well. It's not hard to imagine we will always want the bucket name for our use case, so we pass that in as an always-given output for our component resource.
+
+From here, you can deploy it and get your custom resource appearing in the resource tree in your terminal! You can also share it with others so they can import the resource and use it without ever needing to understand all of the underlying needs of a standard storage system on AWS.
```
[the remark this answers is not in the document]
nimbinatus: Oh weird, that didn't happen locally 🤔  I'll fix it. Thanks!

themes/default/content/learn/abstraction-encapsulation/component-resources/index.md
nimbinatus: :100: Good idea

themes/default/content/learn/abstraction-encapsulation/component-resources/index.md
```diff
@@ -0,0 +1,94 @@
...
+            policy=self.define_policy(),
+            opts=pulumi.ResourceOptions(parent=self.bucket)
+        )
+        return bucket_policy
+
+
+
+```
+
+Within `super()`'s init, we pass in a name for the resource, which we recommend being of the form `<package>:<module>:<type>` to avoid type conflicts since it's being registered alongside other resources like the Bucket resource we're calling (`aws:s3:Bucket`).
```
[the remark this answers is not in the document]
nimbinatus: :100:

themes/default/content/learn/abstraction-encapsulation/component-resources/index.md
```diff
@@ -0,0 +1,94 @@
...
+
+That last call in the init, `self.register_outputs({})`, passes Pulumi the expected outputs so Pulumi can read the results of the creation or update of a component resource just like any other resource, so don't forget that call! You can register default outputs using this call, as well. It's not hard to imagine we will always want the bucket name for our use case, so we pass that in as an always-given output for our component resource.
```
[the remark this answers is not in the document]
nimbinatus: 💯

themes/default/content/learn/abstraction-encapsulation/encapsulation/index.md
nimbinatus: ✅  done

PR #878, opened 2021-11-19
transform docs nav item into a menu

susanev:
## overview
this is transforming the docs nav item into a menu that contains docs, registry, and sdk reference. this is to help folks get easier access to these things and is intended as a small fix before we can do more extensive work to the docs navigation and potentially marketing top navigation.

there are a few general ux concerns about this change that ill mention
- label of docs repeated
- registry in the nav, but once you go to registry this nav is not represented there
- both docs and sdk reference go to docs, just a specific section in docs which may be confusing

im good with ignoring all of these concerns, to try out and see if this helps folks get easier access to things before we can make larger changes.

this work was motivated by feedback shared in community slack.

_note: registry will go to the weird registries page until this gets to docs_

#### mobile
mobile

#### desktop
desktop

### DIFF
--- a/themes/default/layouts/partials/header.html
+++ b/themes/default/layouts/partials/header.html
@@ -70,8 +70,28 @@
                     <li>
                         <a href="{{ relref . "/pricing" }}">Pricing</a>
                     </li>
-                    <li>
-                        <a href="{{ relref . "/docs" }}">Docs</a>
+                    <li class="header-nav-item-drop-down">
+                        <span class="dropdown-title">Docs</span>
+                        <span class="dropdown-menu-caret"></span>
+                        <div class="header-nav-drop-down-menu">
+                            <ul>
+                                <li>
+                                    <a href="{{ relref . "/docs" }}">
+                                        Docs
+                                    </a>
+                                </li>
+                                <li>
+                                    <a href="{{ relref . "/registry" }}">
+                                        Registry
+                                    </a>
+                                </li>
+                                <li>
+                                    <a href="{{ relref . "/docs/reference/pulumi-sdk" }}">
+                                        SDK Reference
+                                    </a>
+                                </li>
+                            </ul>
+                        </div>
                     </li>
                     <li class="header-nav-item-drop-down">
                         <span class="dropdown-title">Learn</span>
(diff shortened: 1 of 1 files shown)


susanev:
@infin8x @cnunciato super open to feedback on the items added or their link text; let me know if you have other ideas

cnunciato:
@susanev Do you have an issue or link you can share that explains the motivation for this change? I'm not against it or anything, just want to better understand what we're solving for before just emitting feedback. 😄

susanev:
@cnunciato i do not, its from a slack convo that i should have linked, sry about that.
https://pulumi.slack.com/archives/C85BS3LJZ/p1637252998094600

but if you don't support this change im happy to close, just let me know.

cnunciato:
Ok cool, thanks -- I see the motivation now, having read through that thread, and I've definitely had the same kinds of experiences this person describes. 

At the same time, this change doesn't feel quite right to me, in part for the same reasons you cite in the PR description, but also:

* There are already on-page links that point to Registry and Docs (two in the practitioner nav, which we presumably want to encourage people to use, and one in the main nav), so both `/registry` and `/docs` are already just one click away -- and it sounds like this particular user isn't looking for more ways to access to those pages, anyway

* The SDK Reference link is nice, but probably wouldn't be used often enough to call for a place in the main nav (though we should consult the data to be sure) -- and indeed might even frustrate users who might be looking, say, for the TypeScript docs for AWS -- as this user's feedback seems to suggest

So to me, it sounds like the need is more along the lines of:

* Make it easier to get to the API docs for a given provider
* Make it easier to get to the Pulumi SDK and CLI docs

Does that sound right? If so, it might make sense to make a few adjustments to the Docs home page to uplevel the visibility of these two things. Most of what's on that page is (a) pretty dusty and (b) probably not all that useful to frequent users, so if we could still make it the primary thing to "Get Started" but also give returning visitors a way to get to the docs for the major providers (with a pointer to Registry as @infin8x suggests) and to the SDK and CLI docs as well, we might end up in a much better place. What do you think?

susanev:
@cnunciato im going to go ahead and close this since Alex will be out for a bit. this was meant to be a small quick fix. when we are ready for larger changes we can incorporate your feedback.

PR #882, opened 2021-11-29
Update self-hosted docs for configuring the email identity availability

praneetloke:
Related to https://github.com/pulumi/pulumi-service/pull/7726.

### DIFF
--- a/themes/default/content/docs/guides/self-hosted/console.md
+++ b/themes/default/content/docs/guides/self-hosted/console.md
@@ -93,8 +93,8 @@ The following are the core environment variables that are required at a minimum.
 
 | Variable Name | Description |
 | ------------- | ----------- |
-| PULUMI_HIDE_EMAIL_LOGIN| When `true`, hides the email identity as a login option but does not disable the API handler. To disable the API handler for email logins set the API container environment variable `PULUMI_DISABLE_EMAIL_LOGIN`. Refer to the API container [environment variables]({{< relref "api#other-env-vars" >}}) for more information. |
-| PULUMI_HIDE_EMAIL_SIGNUP | When `true` hides the email identity as a signup option but does not disable the API handler. To disable the API handler for email signups set the API container environment variable `PULUMI_DISABLE_EMAIL_SIGNUP`. Refer to the API container [environment variables]({{< relref "api#other-env-vars" >}}) for more information. |
+| PULUMI_HIDE_EMAIL_LOGIN | When `true`, hides the email identity as a login option but does not disable the API handler. To disable the API handler for email logins set the API container environment variable `PULUMI_DISABLE_EMAIL_LOGIN`. Refer to the API container [environment variables]({{< relref "/docs/guides/self-hosted/api#other-env-vars" >}}) for more information. |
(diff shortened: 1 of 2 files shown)


chrsmith (approved):
The changes are fine, but it might be clearer to be explicit about "pairing" the console/API env vars.

It isn't super clear what is meant by "To hide the email signup option from the Console refer to []({{< relref "console#email-identity" >}})".

Instead, what do you think of a more imperative tone like: "This should only be set if the `PULUMI_DISABLE_EMAIL_SIGNUP` environment variable is set in the API, see [[configuring the api service]]."

Do you think that would be an improvement?

praneetloke:
> It isn't super clear what is meant by "To hide the email signup option from the Console refer to []({{< relref "console#email-identity" >}})".

You beat me to it. I made a mistake there and didn't fill out the link text initially. I have fixed that now.

> Instead, what do you think of a more imperative tone like: "This should only be set if the `PULUMI_DISABLE_EMAIL_SIGNUP` environment variable is set in the API, see [[configuring the api service]]."

Sure, yeah. I'll update the text and let you know.

themes/default/content/docs/guides/self-hosted/console.md
```diff
@@ -89,6 +89,13 @@ The following are the core environment variables that are required at a minimum.
 | BITBUCKET_OAUTH_ID| Atlassian Bitbucket OAuth consumer client ID. Used for Bitbucket OAuth signins. |
 | BITBUCKET_OAUTH_SECRET | |
 
+### Email and password {#email-identity}
+
+| Variable Name | Description |
+| ------------- | ----------- |
+| PULUMI_HIDE_EMAIL_LOGIN| When `true`, hides the email identity as a login option but does not disable the API handler. To disable the API handler for email logins set the API container environment variable `PULUMI_DISABLE_EMAIL_LOGIN`. Refer to the API container [environment variables]({{< relref "api#other-env-vars" >}}) for more information. |
```
cnunciato, line 96:
```suggestion
| PULUMI_HIDE_EMAIL_LOGIN| When `true`, hides the email identity as a login option but does not disable the API handler. To disable the API handler for email logins set the API container environment variable `PULUMI_DISABLE_EMAIL_LOGIN`. Refer to the API container [environment variables]({{< relref ""/docs/guides/self-hosted/api#other-env-vars" >}}) for more information. |
```

themes/default/content/docs/guides/self-hosted/console.md
```diff
@@ -89,6 +89,13 @@ The following are the core environment variables that are required at a minimum.
...
+| PULUMI_HIDE_EMAIL_SIGNUP | When `true` hides the email identity as a signup option but does not disable the API handler. To disable the API handler for email signups set the API container environment variable `PULUMI_DISABLE_EMAIL_SIGNUP`. Refer to the API container [environment variables]({{< relref "api#other-env-vars" >}}) for more information. |
```
cnunciato, line 97:
```suggestion
| PULUMI_HIDE_EMAIL_SIGNUP | When `true` hides the email identity as a signup option but does not disable the API handler. To disable the API handler for email signups set the API container environment variable `PULUMI_DISABLE_EMAIL_SIGNUP`. Refer to the API container [environment variables]({{< relref ""/docs/guides/self-hosted/api#other-env-vars" >}}) for more information. |
```

praneetloke:
@chrsmith I pushed an update. Can you please take another look? I didn't use the exact phrasing you had suggested. Instead, I made it explicit that simply hiding the button doesn't disable the API handler. That must be done separately.

cnunciato (approved):
Agree with Chris's feedback; otherwise just some `relref` fixups.

praneetloke:
Thanks, @cnunciato! I have fixed up the relref's.

Issue #639, opened 2021-09-28
Add additional customer logos to pulumi.com


gthuang:
Add additional customer logos to the bottom of the existing row using similar layout as below.

![image](https://user-images.githubusercontent.com/14854767/135126571-e19a4987-6e46-42db-b0b7-67aa492c589d.png)

gthuang:
@zchase can we please add [these](https://docs.google.com/document/d/1HTp36sCnilBUjdWED_m24bQJYzm98Jssj_EclohT3lY/edit) logos to the page?

PR #887, opened 2021-12-02
Update customer logos on the home page

zchase:
Fixes: https://github.com/pulumi/pulumi-hugo/issues/639

~~I wasn't able to find good logos for Singtel and CBOE. I was only able to find PNGs and can't to the do gray hover thing with the PNGs, I am going to need someone to provide me with SVGs to add those to the home page.~~


gthuang:
thanks zack! I think the logo layout isn't working as we originally intended. @susanev suggested doing 3 rows of medium-sized logos with equal spacing instead. originally, the two-tiered logo layout was meant to highlight the logos with case studies, but I now don't think is achieving the goal necessarily. 

instead, could we add a secondary (white colored) button next to the "Read Atlassian case study" button that says "See more case studies"?

also, the Atlassian logo color looks incorrect (I think it should be all blue)

susanev (approved):

PR #895, opened 2021-12-09
Update STYLE-GUIDE.md

jkodroff:
* Replace `-` for `<ul>`s with `*`, per Markdown spec.
* Touch up verbiage for neutrality, (hopefully) improved readability.
* Add link to Markdownlint.

### DIFF
--- a/STYLE-GUIDE.md
+++ b/STYLE-GUIDE.md
@@ -2,42 +2,40 @@
 
 This document defines some general styles we adhere to in the docs.
 
-## Words are important
+## Use Inclusive Language
 
-- Instead of _guys_ try _folks_, _yall_, or _everyone_
-- Instead of _crazy_ try _wild_
-- Instead of _click_ use _select_
-- Instead of _dummy_ use _placeholder_
-- Avoid using gendered language
-- Avoid using violent language (e.g., _kill_)
-- Avoid pop-culture references
+Words are important.  Pulumi strives to use language that is clear, harmonious, and friendly to all readers.  With these goals in mind, we use the following guidelines:
 
-Want to learn about why these words should be avoided?
-Check out [Google's inclusive documentation guide](https://developers.google.com/style/inclusive-documentation).
+* Avoid ableist language:
+  * Instead of _crazy_ try _wild_.
+  * Instead of _click_ use _select_.
+  * Instead of _dummy_ use _placeholder_.
+* Avoid unnecessarily gendered language: Instead of _guys_ try _folks_, _yall_, or _everyone_.
+* Avoid using violent language (e.g., _kill_)
+* Avoid pop-culture references as such references may not be familiar to all readers.
 
 ## Headings
 
-- You should be able to scan the headings of a document and get a good sense of the page
-- Every page should have exactly one h1
-- Don't skip heading levels in Markdown files, e.g., if your previous heading level was a 2, the next heading should be a 2 or a 3, but not a 4 or 5
+* Ensure that readers are able to scan the headings of a page and get an effective overview of the page's content.
+* Every page should have exactly one `h1`.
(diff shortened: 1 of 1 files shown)


STYLE-GUIDE.md
```diff
@@ -2,38 +2,36 @@
 
 This document defines some general styles we adhere to in the docs.
 
-## Words are important
+## Use Inclusive Language
```
susanev, line 5: could we have the words are important line as a line of text below the heading?

STYLE-GUIDE.md
```diff
@@ -2,38 +2,36 @@
...
 
 Want to learn about why these words should be avoided?
 Check out [Google's inclusive documentation guide](https://developers.google.com/style/inclusive-documentation).
 
 ## Headings
 
-- You should be able to scan the headings of a document and get a good sense of the page
-- Every page should have exactly one h1
-- Don't skip heading levels in Markdown files, e.g., if your previous heading level was a 2, the next heading should be a 2 or a 3, but not a 4 or 5
+* Readers should be able to scan the headings of a page and get an effective overview of the page's content.
```
susanev, line 19: i prefer this not be put on the readers and instead on the authors of the docs... can you say more about why _you_ is bad to use here?

STYLE-GUIDE.md
```diff
@@ -2,38 +2,36 @@
...
+* Every page should have exactly one h1.
+* Headings levels should only increment one level at a time.  E.g., if your previous heading level was an h2, the next heading should be a h2 or an h3, but not an h4 or h5.
 
 ## Links
 
-- Link text should be descriptive and be understandable outside of the context that it is in.
-- Avoid link text like _here_, _click here_, _see here_
+* Link text should be descriptive and have meaning outside of the surrounding context: Avoid link text like _here_, _click here_, _see here_ and instead link to the title of the linked page, e.g. "see [Pulumi Pricing](https://www.pulumi.com/pricing/)".  (This is recommended in order to accommodate visually impared users who use screen readers and often jump through the links of a document.)
```
susanev, line 25: sooo our page titles are not good enough right now to require this, so id prefer to come back and add that direction after they get improved

and this is not only for visually impaired folks, its for everyone!

STYLE-GUIDE.md
```diff
@@ -49,8 +47,12 @@ This bit of info is serious. If you missed it, bad things could happen.
...
 ## Blockquotes
 
-- Use blockquotes only when you are quoting content from another source.
+* Use blockquotes only when directly quoting content from another source.
 
 ### Example
 
 > This is something a person said.
+
+## Additional Resources
```
susanev, line 56: mind moving the link to the google doc down here as well?

susanev (approved):

STYLE-GUIDE.md
jkodroff: Yes - with an end to that sentence saying why.  Give me a minute to work on that text.

STYLE-GUIDE.md
jkodroff: I changed it as a stylistic thing - I was kinda trained not to use second-person singular in order to make the language more neutral, but my style may be outdated here, e.g.: https://docs.microsoft.com/en-us/style-guide/grammar/person

Let me do a little more research into what the Big Tech Companies recommend style-wise.

susanev: okay id prefer to keep it in this situation because we are describing things we want the docs authors to do. we could adjust this to _as an author_... if you prefer though. i don't want it to read as though its on the readers though.

jkodroff:
@susanev Pushed another revision.  LMK what you think.

susanev (approved):
:shipit: 

thank you!!

PR #911, opened 2021-12-16
[Learn] Proposed layout updates

nimbinatus:
Now that we have more pathways (and as we add more), I was thinking a hierarchy might make navigation a bit easier. Any thoughts on this proof-of-concept for a future iteration?

### DIFF
--- a/themes/default/content/learn/building-with-pulumi/_index.md
+++ b/themes/default/content/learn/building-with-pulumi/_index.md
@@ -6,6 +6,7 @@ draft: false
 description: Learn how to use Pulumi for more complex configurations with multiple environments.
 meta_desc: Learn how to use Pulumi for more complex configurations with multiple environments.
 index: 4
+level: novice
 meta_image: practice.png
 youll_learn:
     - Using stacks for unique configurations of different environments
--- a/themes/default/content/learn/pulumi-fundamentals/_index.md
+++ b/themes/default/content/learn/pulumi-fundamentals/_index.md
@@ -11,6 +11,7 @@ meta_desc: |
     application.
 index: 0
 meta_image: fundamentals.png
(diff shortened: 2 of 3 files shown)


susanev:
i really like the general idea of improving the navigation here. headings are def the lowest eng cost way to make that happen, but i wonder if filters and even search might better fulfill the user's needs and make it a bit more flexible.

have we been able to gather any user feedback? can i review that anywhere? or have we looked at the data for what lessons folks have tried and which they have finished? i ask because it would be great to use that data to inform the design choices here

and some more thoughts...

- i dont think we have enough lessons yet to want this, i think once we have more than 12 then something like filtering would be a great addition
    - breaking it up too early will make it appear that we have less than we do, and also make stuff slightly more difficult to get to because of the extra scrolling
- im a bit confused, i thought we were going to have pathways based on topics that went through the beginner, intermediate, advanced levels but what you have here makes it seem like our main separation will be level not topic, is that correct?
- instead of applying a single filter with headings id love for us to explore filers at the top of the page,  some that come to mind our: level, topic, cloud, user-type, etc. then we can have a mix of lessons at the top which could engage a larger variety of users beyond just the folks that want the beginner stuff

susanev:
@nimbinatus is this important to you to keep around in pr form?

nimbinatus:
Ah, sorry. No, we don't need to keep the PR. I've kept it in draft so I don't forget about it, but I can close it.

nimbinatus:
If we can leave the branch until the end of April, though, I'd appreciate it as I'm hoping to finish out a couple more pathways and have something to explore by then.

susanev:
branches are cool no worries!

nimbinatus:
Moved to #1331

PR #923, opened 2021-12-21
Update guidance on implicit templates for Deploy Button

leezen:
I re-ordered the explicit vs. implicit sections here and added guidance on not using implicit. I considered removing implicit completely, but felt like someone might choose to consider it knowing the drawbacks? Open to removing it completely though.


susanev:
@leezen id personally prefer to remove because id be worried about documenting something that we know will exclude a set of users. but maybe im missing something?

leezen:
@susanev Nope -- I think I was tending toward being conservative, but as I mentioned, pretty open to just ripping it out. Let me push that up.

susanev (approved):
thank you!!

PR #938, opened 2021-12-29
Break Resource page into separate pages for each core concept

lukehoban:
Even after breaking the Programming Model content down, this Resources page still contained a *lot* of core content.  This made a lot of important features feel very "hidden", like resource options, components, dynamic providers, and more.

It also led to suboptimal search results, both in Google and in our own internal site search, because searches for specific concepts (like "pulumi aliases") would return the "Resources" page, which didn't even rank particularly well as it doesn't "feel" like definitive content for this specific topic.

This change breaks this page into 18 separate pages.  It keeps the "Custom Resource" content in the main Resources page (since that content is nearly entirely actually not specific to Custom Resources anyway), pulls the Component Resource, Resource Providers and Dynamic Providers content into separate top level items, and pulls the Resource Options content into it's own index with a separate page per resource option.

Some of the resulting pages are fairly light in content, but this actually offers room to flesh these out more fully now that they have room to breathe.

Reading through this content as part of the refactor, there is quite a bit of cleanup we could still do here.  I took a few liberties to clean up some things as part of this, but mostly focused on the refactor instead of changes to prose.

This also introduces a set of client-side redirects to the resources page, so that each of the previous anchors which now lives elsewhere will get redirected in the client, following the same strategy used in the Programming Model page refactor.


susanev (approved):
this is wonderful, thank you for doing this work!!

for anyone else reviewing here is what the new nav structure looks like...
new-nav-structure

a few questions...
- should the children under resources be in alphabetical order?
- the dynamic providers page is still quite long, would it make sense to break it up even more?

themes/default/content/docs/intro/concepts/resources/_index.md
```diff
@@ -0,0 +1,425 @@
...
+{
+    Name = "my-role-001",
+});
+```
+
+{{% /choosable %}}
+
+{{< /chooser >}}
+
+If the `name` property is not available on a resource, consult the [Registry]({{< relref "/registry" >}}) for the specific resource you are creating. Some resources use a different property to override auto-naming. For instance, the `aws.s3.Bucket` type uses the property `bucket` instead of name. Other resources, such as `aws.kms.Key`, do not have physical names and use other auto-generated IDs to uniquely identify them.
```
t0yv0, line 224: TIL.. this is a significant complication it seems..

themes/default/content/docs/intro/concepts/resources/components.md
```diff
@@ -0,0 +1,287 @@
+---
+title: "Components"
+meta_desc: A component resource is a logical grouping of resources.
+menu:
+  intro:
+    parent: resources
+    weight: 3
+---
+
+A component resource is a logical grouping of resources. Components resources usually instantiate a set of related resources in their constructor, aggregate them as children, and create a larger, useful abstraction that encapsulates their implementation details.
```
t0yv0, line 10: "Components resources" --> "component resources"?

themes/default/content/docs/intro/concepts/resources/components.md
```diff
@@ -0,0 +1,287 @@
...
+});
+```
+
+{{% /choosable %}}
+
+{{< /chooser >}}
+
+The call to `registerOutputs` typically happens at the very end of the component resource’s constructor.
+
+The call to `registerOutputs` also tells Pulumi that the resource is done registering children and should be considered fully constructed, so—although it’s not enforced—the best practice is to call it in all components even if no outputs need to be registered.
```
t0yv0, line 216: This is an interesting bit. So registerOutputs call is truly optional, we support not calling it?

themes/default/content/docs/intro/concepts/resources/options/parent.md
```diff
@@ -0,0 +1,70 @@
...
+var parent = new MyResource("parent", new MyResourceArgs());
+var child = new MyResource("child", new MyResourceArgs(),
+    new CustomResourceOptions { Parent = parent });
+```
+
+{{% /choosable %}}
+
+{{< /chooser >}}
+
+Using parents can clarify causality or why a given resource was created in the first place. For example, this pulumi up output shows an AWS Virtual Private Cloud (VPC) with two subnets attached to it, and also shows that the VPC directly belongs to the implicit pulumi:pulumi:Stack resource:
```
t0yv0, line 60: Perhaps a quick note here on inheriting options such as `provider`, `providers`, `transformations`, and `protect` from parents.

t0yv0 (approved):
Thank you for following up here, I think this is a significant improvement!

joeduffy:
Any reason to stop at leaving the few topics still in the top-level resources doc? If we do keep the few topics there (creation, names, args, getters), it would be worth advertising that the other topics also exist. As far as I can tell, you'd need to pay close attention to the nav on the left hand side to know that other content exists. A section at the top and/or bottom, maybe?

themes/default/content/docs/intro/concepts/resources/components.md
lukehoban: Yes.  The only affect it has today is to update the `updating...` display in the CLI to `updated` on the component itself.  

As we've discussed recently, we should make a breaking change to require this, as an indication of when the component construction is complete so that dependencies on the component are processed at that point in time (and can see "all the children".  This is tracked in https://github.com/pulumi/pulumi/issues/2653, https://github.com/pulumi/pulumi/issues/2484 and https://github.com/pulumi/pulumi/issues/5941.

lukehoban:
> should the children under resources be in alphabetical order?

I wasn't sure about this at first, but after another look, I think this ends up with a natural order that is also nicely alphabetical, so have pushed a change that updates this to be alphabetical for now.

> the dynamic providers page is still quite long, would it make sense to break it up even more?

Although long, this one is more clearly logically "one topic".  I'll leave this as one page for now.

> Any reason to stop at leaving the few topics still in the top-level resources doc? 

I agree, I'm going to try one of these two options - splitting most of the remaining content out and/or adding more of an index of the sub-pages to the main resources page.

themes/default/content/docs/intro/concepts/resources/components.md
lukehoban: Fixed.

themes/default/content/docs/intro/concepts/resources/options/parent.md
lukehoban: Done.

joeduffy:
Gentle prod, @lukehoban 

I've had to link to these pages a lot lately, and I would _love love love_ for this change to be there.

By the way, will prior links "just work" after this change, including anchored links like https://www.pulumi.com/docs/intro/concepts/resources/#dependson?

lukehoban:
Yep - still planning to land this. I agree with your feedback above about doing a little more to split the core custom resource out of the overview page - so want to finish that up before merging. But expect to land this this week. 

> By the way, will prior links "just work" after this change, including anchored links like https://www.pulumi.com/docs/intro/concepts/resources/#dependson

Yes. This uses the same client-side redirect trick we have used in the past to ensure that every anchor that was on the old page will redirect to its new location. It’s slightly more “jagged” than a server side redirect, but from my testing is still a good experience. You can try it yourself in the preview link above.

themes/default/content/docs/intro/concepts/resources/options/import.md
```diff
@@ -0,0 +1,154 @@
...
+  intro:
+    parent: options
+    weight: 7
+---
+
+The `import` resource option imports an existing cloud resource so that Pulumi can manage it. Imported resources can have been provisioned by any other method, including manually in the cloud console or with the cloud CLI.
+
+To import a resource, first specify the `import` option with the resource’s ID. This ID is the same as would be returned by the id property for any resource created by Pulumi; the ID is resource-specific. Pulumi reads the current state of the resource with the given ID from the cloud provider. Next, you must specify all required arguments to the resource constructor so that it exactly matches the state to import. By doing this, you end up with a Pulumi program that will accurately generate a matching desired state.
+
+This example imports an existing EC2 security group with ID sg-04aeda9a214730248 and an EC2 instance with ID `i-06a1073de86f4adef`:
```
cnunciato, line 14:
```suggestion
This example imports an existing EC2 security group with ID `sg-04aeda9a214730248` and an EC2 instance with ID `i-06a1073de86f4adef`:
```

themes/default/content/docs/intro/concepts/resources/options/aliases.md
```diff
@@ -0,0 +1,116 @@
...
+    weight: 2
+aliases:
+  - /docs/intro/concepts/resources/#aliases
+---
+
+The `aliases` resource option provides a list of aliases for a resource or component resource. If you’re changing the name, type, or parent path of a resource or component resource, you can add the old name to the list of aliases for a resource to ensure that existing resources will be migrated to the new name instead of being deleted and replaced with the new named resource.
+
+Aliases are frequently used when refactoring Pulumi programs.
+
+For example, imagine we change a database resource’s name from `old-name-for-db` to `new-name-for-db`. By default, when we run pulumi up, we see that the old resource is deleted and the new one created. If we annotate that resource with the aliases option, however, the resource is updated in-place:
```
cnunciato, line 16:
```suggestion
For example, imagine we change a database resource’s name from `old-name-for-db` to `new-name-for-db`. By default, when we run `pulumi up`, we see that the old resource is deleted and the new one created. If we annotate that resource with the aliases option, however, the resource is updated in-place:
```

themes/default/content/docs/intro/concepts/resources/options/dependson.md
```diff
@@ -0,0 +1,60 @@
...
+  intro:
+    parent: options
+    weight: 5
+---
+
+The `dependsOn` resource option creates a list of explicit dependencies between resources.
+
+Pulumi automatically tracks dependencies between resources when you supply an input argument that came from another resource’s output properties. In some cases, however, you may need to explicitly specify additional dependencies that Pulumi doesn’t know about but must still respect. This might happen if a dependency is external to the infrastructure itself—such as an application dependency—or is implied due to an ordering or eventual consistency requirement. The `dependsOn` option ensures that resource creation, update, and deletion operations are done in the correct order.
+
+This example demonstrates how to make res2 dependent on res1, even if there is no property-level dependency:
```
cnunciato, line 14:
```suggestion
This example demonstrates how to make `res2` dependent on `res1`, even if there is no property-level dependency:
```

themes/default/content/docs/intro/concepts/resources/options/parent.md
```diff
@@ -0,0 +1,80 @@
...
```
cnunciato, line 60:
```suggestion
Using parents can clarify causality or why a given resource was created in the first place. For example, this pulumi up output shows an AWS Virtual Private Cloud (VPC) with two subnets attached to it, and also shows that the VPC directly belongs to the implicit `pulumi:pulumi:Stack` resource:
```

themes/default/content/docs/intro/concepts/resources/options/protect.md
```diff
@@ -0,0 +1,51 @@
+---
+title: "Protect"
+meta_desc: An in depth look at Pulumi resources and their usage.
+menu:
+  intro:
+    parent: options
+    weight: 9
+---
+
+The `protect` resource option marks a resource as protected. A protected resource cannot be deleted directly. Instead, you must first set `protect: false` and run `pulumi up`. Then you can delete the resource by removing the line of code or by running `pulumi destroy`. The default is to inherit this value from the parent resource, and `false` for resources without a parent.
```
cnunciato, line 10: Do you think it'd make sense to briefly mention/link to `pulumi state unprotect` here as well?

themes/default/content/docs/intro/concepts/resources/options/replaceonchanges.md
```diff
@@ -0,0 +1,100 @@
...
+        Input = "something",
+    }
+}, pulumi.ReplaceOnChanges([]string{"spec.input"}));
+```
+
+{{% /choosable %}}
+
+{{< /chooser >}}
+
+The property paths provided as input to `replaceOnChanges` can each describe a subset of the properties of the resource which should trigger a replacement on changes.  The `*` wildcard can be used in any part of a path.  A few examples of what changes will trigger replacement for a given property path string:
```
cnunciato, line 87:
```suggestion
The property paths provided as input to `replaceOnChanges` can each describe a subset of the properties of the resource which should trigger a replacement on changes.  The `*` wildcard can be used in any part of a path.  A few examples of what changes will trigger replacement for a given property path string are:
```

themes/default/content/docs/intro/concepts/resources/options/replaceonchanges.md
```diff
@@ -0,0 +1,100 @@
...
+- `*`: any property change
+- `spec`: any change to the `spec` property or any of its children
+- `spec[0]`: any change to the first element of the array in the `spec` property or any of its children
+- `spec[*].item`: any change to the `item` property of any element of the array in the `spec` prroperty or any of the `item` property's children
+
+{{% notes "info" %}}
+The property paths passed to `replaceOnChanges` should always be the "camelCase" version of the property name, as used in the core Pulumi resource model.
+{{% /notes %}}
+
+If there initialization errors on a resource (because the resource was created but failed to fully initialize correctly on a previous deployment), then this will typically cause the resource to be updated on the following Pulumi update, even if there are no other changes to the resources inputs.  If `*` is specified as a property path for `replaceOnChanges`, then initialization errors will trigger a replacement instead of an update.
```
cnunciato, line 98: This sentence seems like it needs fixing, but I'm not exactly sure what it's trying to say. Maybe this?

```suggestion
If there are initialization errors on a resource (because the resource was created but failed to fully initialize correctly on a previous deployment), then this will typically cause the resource to be updated on the following Pulumi update, even if there are no other changes to the resource's inputs.  If `*` is specified as a property path for `replaceOnChanges`, then initialization errors will trigger a replacement instead of an update.
```

cnunciato (approved):
Agree making the Resources page more of an index would make this even better, but this is a huge improvement -- thanks for doing it. Only a handful of little things I noticed while reading.

susanev:
@lukehoban there is a ref issue over in the docs pr for this..

```
ERROR 2022/01/18 02:30:21 [en] REF_NOT_FOUND: Ref "additional-secret-outputs": "/tmp/hugo_cache/modules/filecache/modules/pkg/mod/github.com/pulumi/pulumi-hugo/themes/default@v0.0.0-20220118014907-135686cccef6/content/docs/intro/concepts/resources/options/_index.md:13:29": page not found
```

PR #939, opened 2021-12-29
Expand `aliases` documentation to describe parent alias inheritance

lukehoban:


cnunciato:
@lukehoban Do you want to close this since it looks like it's been incorporated at https://www.pulumi.com/docs/intro/concepts/resources/options/aliases/?

lukehoban:
Indeed - closing now.

PR #951, opened 2022-01-04
Filter author and tag taxonomy templates to blog posts only

cnunciato:
Our author and tag listings are unintentionally showing Learn content (which also happens to have `author` and `tag` metadata), so this change add a filter to those pages to have them show only content from the blog. 

This is a less-than-ideal fix in that it _should_ be possible (one would think) to have Hugo generate section-specific `tag` and `author` taxonomy pages, but a quick look through the docs and some fiddling this morning didn't make that super clear to me. So we'll probably want to revisit this once we start thinking about supporting taxonomy pages outside of just the blog.

### DIFF
--- a/themes/default/content/learn/building-with-pulumi/secrets/index.md
+++ b/themes/default/content/learn/building-with-pulumi/secrets/index.md
@@ -11,7 +11,6 @@ meta_image: meta.png
 authors:
     - matt-stratton
 tags:
-    - learn
     - secrets
 block_external_search_index: false
 ---
--- a/themes/default/content/learn/building-with-pulumi/_index.md
+++ b/themes/default/content/learn/building-with-pulumi/_index.md
@@ -12,8 +12,6 @@ youll_learn:
     - Sharing values from one Pulumi program or project to another
     - Working with secrets inside of Pulumi
     - Testing your Pulumi programs
-tags:
-    - learn
 providers:
     - aws
 block_external_search_index: false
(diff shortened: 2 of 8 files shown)


susanev (approved):
thanks for the quick fix @cnunciato! and for reporting the issue @stack72 !!

themes/default/content/learn/building-with-pulumi/secrets/index.md
```diff
@@ -289,4 +288,4 @@ programmatically, review the
 From here, we're moving on to the last part of the Pulumi in Practice pathway:
 testing. Onward!
 
-<!-- [^1]: [state]({{< relref "/docs/intro/concepts/state" >}}) -->
\ No newline at end of file
+<!-- [^1]: [state]({{< relref "/docs/intro/concepts/state" >}}) -->
```
nimbinatus, line 291: Was this branch from an older version of the pathways? I could have sworn I took these out so we had time to investigate footnotes in terms of UX, so I was surprised to see them pop up here.

nimbinatus (approved):
Non-blocking question in the comments, but otherwise LGTM.

I was actually thinking... I wonder if the fact that there's a high-level Learn section and then a learn tag might be causing a collision somewhere in Hugo, which might have been part of the problem 🤔

themes/default/content/learn/building-with-pulumi/secrets/index.md
cnunciato: This is branched from latest master, and I see this line there (and in a few other places, looks like), so maybe just snuck past ya. Want me to remove them? Or maybe in a separate PR?

nimbinatus: Nah, we can leave them for now, I think. I must have missed some :)

PR #956, opened 2022-01-07
feat(post): add intro post for the hackathon series

nimbinatus:
Adding the first post for the hackathon review.


susanev (approved):
love it! thank you!!
i wonder if we should make a new `pulumi-culture` tag or similar?

nimbinatus:
Ooo, I think that would be a lovely tag to have. @cnunciato, any objection to adding another tag to the tag cloud? I don't know if we have a specific process for that as I know the blogging instructions said to avoid making a new one...

cnunciato:
@nimbinatus no process really, tag away!

PR #961, opened 2022-01-10
feat(post): circular dependencies 2021 hackathon

danielrbradley:


themes/default/content/blog/exploring-circular-dependencies/index.md
```diff
@@ -0,0 +1,226 @@
...
+draft: true
+meta_desc: Hack day project exploring approaches to solving circular resource dependencies.
+meta_image: meta.png
+authors:
+  - daniel-bradley
+tags:
+  - features
+---
+
+As part of our winter hackathon near the end of last year, we decided to explore solutions to a common thought when people are using Pulumi for their systems. A question that’s been asked in a few different forms is how to resolve circular dependencies between resources in a Pulumi program.
```
Frassle, line 13: "common thought" read a bit odd, maybe "common problem" or "issue"?

themes/default/content/blog/exploring-circular-dependencies/index.md
```diff
@@ -0,0 +1,226 @@
...
+
+A simple example of this idea is a modern web application with a static front-end and an API, where the front-end needs to know the URL of the API to be able to call it and the API needs to know the source domain of the front-end to allow it access via CORS. As these two resources rely on one another to be created, they are circular dependencies.
```
Frassle, line 15: I don't think the comma after "created" is needed.

themes/default/content/blog/exploring-circular-dependencies/index.md
```diff
@@ -0,0 +1,226 @@
...
+      OPPONENT_FN_NAME: ping.name,
+    },
+  },
+});
+
+// Engine re-runs "up" if this changes the value
+pongName.set(pong.name);
+```
+
+Deployments should always match the preview, but when triggering the re-deployment, we’re making a change that wasn’t in the preview. This action could also result in a never-ending loop where every deployment makes another change to the state and triggers another deployment. So let’s explore a different option that might not have this issue.
```
Frassle, line 115: Maybe be more explicit here that only the first deployment will match the preview?

themes/default/content/blog/exploring-circular-dependencies/index.md
```diff
@@ -0,0 +1,226 @@
...
+  },
+});
+
+// Determine affected resources and re-update
+pongName.set(pong.name);
+```
+
+The difference between refs and state is that, instead of triggering a whole re-deploy when the value changes, refs would be integrated into the dependency graph—similar to resources. When the value of the ref changes, we could internally work out which resources depend on that ref and change the target state for only those resources.
+
+A reference would only be allowed to be updated once in a program to avoid creating never-ending loops. For example, if a resource uses a reference value, then we update the reference value from that same resource; if that continues indefinitely, the program might never exit.
```
Frassle, line 150: I don't think this sort of loop is actually expressible with Refs? The whole point of refs is you can't get the updated resource outputs to be able to call Ref.Set again?

Frassle (commented):
👍 Good stuff, thanks for writing this all up nicely

nimbinatus (changes requested):
A few notes, but it's good!

themes/default/content/blog/exploring-circular-dependencies/index.md
```diff
@@ -0,0 +1,226 @@
...
+```typescript
+const pongName = new pulumi.Ref("pong-name", {
+  initialValue: "",
+});
+
+const ping = new CallbackFunction("ping", {
+  callback: pingPongHandler,
+  environment: {
+    variables: {
+      OPPONENT_FN_NAME: pongName, // Special type the Pulumi engine will track
```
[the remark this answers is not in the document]
danielrbradley: It could be either as it's just an illustration. Will add `value`

themes/default/content/blog/exploring-circular-dependencies/index.md
danielrbradley: Hmm, perhaps not - shall I just remove this paragraph then?

danielrbradley:
@nimbinatus is the cover image too abstract (a reference to spiderman pointing at spiderman meme)?

themes/default/content/blog/exploring-circular-dependencies/index.md
Frassle: TIL

nimbinatus:
> http://pulumi-hugo-origin-pr-961-83052526.s3-website.us-west-2.amazonaws.com

I *love* it! No, we need this kind of personality 😁  Keep it!

nimbinatus (approved):

themes/default/content/blog/exploring-circular-dependencies/index.md
danielrbradley: Have reworked 👍

PR #967, opened 2022-01-13
fix console refs

susanev:
😞


nimbinatus (approved):
One question, but the links look fine. +1

PR #971, opened 2022-01-14
Allowlist env vars for hugo 0.92

AaronFriel:
Hugo v0.91.0 added a new key to the config, "security", defining allowlists for env vars and other things. This change allows 0.91 and 0.92 to build and serve the Pulumi site.

TODO:
- [ ] #972

### DIFF
--- a/config/_default/config.yml
+++ b/config/_default/config.yml
@@ -40,6 +40,12 @@ module:
         - source: archetypes
           target: archetypes
 
+security:
+  funcs:
+    getenv:
+      - ASSET_BUNDLE_ID
+      - REPO_THEME_PATH
+      - PULUMI_CONVERT_URL
 
 disableKinds:
   - category


praneetloke:
@cnunciato wondering if you know about this and could review it?

cnunciato:
@praneetloke Absolutely, will do today.

cnunciato (approved):
This is fine, just a couple of things to note:

* This particular config file only applies to running `hugo` commands from the root of _this_ repository, so it'd only be used in local development workflows. When we build the website for real, in pulumi/docs, we use the config file in the root of that repo, so we'd probably want to apply these changes in that repo, too, and probably in pulumi/registry as well.
 
* We're still pinned to Hugo 0.82 across all of our website repos. We don't have to be, though -- I just haven't gotten around to updating all the repo scripts and workflows to use the latest Hugo, Go, and Node versions yet.

So for both of these reasons, this change won't have much of an effect. But I'm happy to merge it for now so that when we do move to hugo@latest, it'll work here, and then we can propagate it to all of the other repos as well. What do you think?

susanev:
hi folks, Meagan is hitting this issue as well. can we merge this soon? or at least make it clear in the dev docs the hugo version you need to be on for local dev?

cnunciato:
We do make this clear when you run `make ensure`, [which is in the README](https://github.com/pulumi/pulumi-hugo#installing-prerequisites) (you get a complaint about not having the right version of any prerequisite, including Hugo), but again I'm totally okay with merging this now, as I did see that those who install latest will run into prolems. 

I also have [this WIP PR](https://github.com/pulumi/pulumi-hugo/pull/975) going that pulls all of these tools into a Docker container such that you don't have to install _any_ of them installed locally. I'll see if I can get this merged today, too, and maybe @meagancojocar can give that a try?

susanev:
@cnunciato i dont think Meagan was seeing that complaint about the re-req when she ran make ensure, at least when she quickly demo'd the issue to me this morning i didn't see it. but she did get all the failed to execute template errors.

cnunciato:
I'm going to go ahead and merge this to unblock anyone who happens to hit it (cc @meagancojocar), but I'm actively working on https://github.com/pulumi/pulumi-hugo/pull/975 which should remove the need to install these tools locally. I also remembered this afternoon that we have a VS Code `.devcontainer` in the root of the repo that @Frassle put together a while back that also seems to work -- or at least it does for me, but since I'm not on an M1, and I know some folks are, I can't be sure either of these container-based options will work for everyone.

PR #975, opened 2022-01-15
Add a Dockerfile

cnunciato:
WIP to make local dev a little "easier" for those who don't want to wrestle with installing all the right tools and versions. To try this, make sure Docker is installed and running, then:

```
make docker
```

then browse to http://localhost:1313.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,30 @@
+FROM ubuntu:18.04
+
+ENV GO_VERSION=1.16.13
+ENV HUGO_VERSION=0.82.0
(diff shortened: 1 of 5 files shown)


cnunciato:
Meant to leave this as draft, as I'm still working on it. Will add reviewers when it's ready for 👀 .

susanev:
@cnunciato only a friendly reminder that this is still here, no worries at all for it to continue to wait until you have time. but in case its ready and you forgot thought id ping you.

cnunciato:
Yep, thanks for the reminder.

PR #978, opened 2022-01-16
fix crosswalk links to example repo

### LINKED ISSUE #977: crosswalk examples for azure and gcp are linking to aws
[themes/default/content/docs/guides/crosswalk/kubernetes/app-services.md](https://www.pulumi.com/docs/guides/crosswalk/kubernetes/app-services/)

the examples linked to on this page all link to the aws example, even if folks have azure or gcp selected in the chooser

this was originally reported by a user through a support ticket

susanev:
fixes #977 

changes are being made to `docs/guides/crosswalk/kubernetes/app-services/`

### DIFF
--- a/themes/default/content/docs/guides/crosswalk/kubernetes/app-services.md
+++ b/themes/default/content/docs/guides/crosswalk/kubernetes/app-services.md
@@ -16,41 +16,27 @@ These services tend to include datastores, and managers for ingress, DNS, and TL
 They can be shared amongst several apps or be specific to workloads, and are
 usually a mix of cloud provider and custom services.
 
+## Example Code
+
 {{% choosable cloud aws %}}
 
-The full code for the AWS app services stack is on [GitHub][gh-repo-stack].
-
-<!-- markdownlint-disable url -->
-[gh-repo-stack]: https://github.com/pulumi/kubernetes-guides/tree/master/aws/05-app-services
-<!-- markdownlint-enable url -->
+* [AWS app services stack on GitHub](https://github.com/pulumi/kubernetes-guides/tree/master/aws/05-app-services)
 
 {{% /choosable %}}
 
 {{% choosable cloud azure %}}
 
-The full code for the Azure app services stack is on [GitHub][gh-repo-stack].
-
-<!-- markdownlint-disable url -->
-[gh-repo-stack]: https://github.com/pulumi/kubernetes-guides/tree/master/azure/05-app-services
-<!-- markdownlint-enable url -->
+* [Azure app services stack on GitHub](https://github.com/pulumi/kubernetes-guides/tree/master/azure/05-app-services)
 
 {{% /choosable %}}
 
 {{% choosable cloud gcp %}}
 
-The full code for the GCP app services stack is on [GitHub][gh-repo-stack].
-
-<!-- markdownlint-disable url -->
-[gh-repo-stack]: https://github.com/pulumi/kubernetes-guides/tree/master/gcp/05-app-services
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/guides/crosswalk/kubernetes/app-services.md
```diff
@@ -16,41 +16,27 @@ These services tend to include datastores, and managers for ingress, DNS, and TL
...
 usually a mix of cloud provider and custom services.
 
-{{% choosable cloud aws %}}
+## Example Code
 
-The full code for the AWS app services stack is on [GitHub][gh-repo-stack].
+{{% choosable cloud aws %}}
 
-<!-- markdownlint-disable url -->
-[gh-repo-stack]: https://github.com/pulumi/kubernetes-guides/tree/master/aws/05-app-services
```
susanev, line 24: i didnt really understand the point of these since they arent being used anywhere else on the page, but if someone has more context pls let me know

cnunciato: Ah, nice catch -- this is a bug resulting from how these links end up being rendered at build-time (because the name `gh-repo-stack` is being reused). I generally don't love the notation that was used here, so I'm definitely 👍 on the changes to use normal Markdown links.

I'm not sure I'd add the Example Code heading at the top like this, though, since this page is meant as more of a guide than a reference like the API docs pages (assuming that's what you were thinking by adding it here). Let me read through it, though, and I'll chime back in after the break.

However I don't think I'd change the overall hierarchy here.

susanev: cool, yah i only added that because i couldn't come up with good enough link text that didn't require some kind of header on that section; tots open to other ideas

susanev: @cnunciato any thoughts on how to make this better? since the bad links were user reported id like to get them fixed soon.

cnunciato: Sure, if you want to get it fixed now, you can just change the format of these three links to be regular Markdown expressions, as that's what's causing the problem. So for example, for each cloud, from this:

```
{{% choosable cloud aws %}}

The full code for the AWS app services stack is on [GitHub][gh-repo-stack].

<!-- markdownlint-disable url -->
[gh-repo-stack]: https://github.com/pulumi/kubernetes-guides/tree/master/aws/05-app-services
<!-- markdownlint-enable url -->

{{% /choosable %}}
```

to this:

```
{{% choosable cloud aws %}}

The full code for the AWS app services stack is on [GitHub][https://github.com/pulumi/kubernetes-guides/tree/master/aws/05-app-services].

{{% /choosable %}}
```

That'd fix the bug. 👍

susanev: wait, didnt i do that? do you want me just to remove the heading?

susanev: oh and leave the link text as is? is that what you mean?

cnunciato: Yes, the heading-related bit is the part I wanted to think about. The link fixes looked good, though.

susanev: cool, will remove as soon as i escape all the meetings

cnunciato (approved):

PR #982, opened 2022-01-18
Add info about API docs as part of package docs contribution

praneetloke:
Added some info about API docs based on how things are currently.


themes/default/content/docs/guides/pulumi-packages/how-to-author.md
```diff
@@ -97,6 +97,10 @@ Future iterations of this guide will cover how to publish the first three items;
 
 For now, all package documentation on Pulumi Registry is published via the [`pulumi/registry` repository on GitHub](https://github.com/pulumi/registry). To publish your package on Pulumi Registry:
 
+> API docs for your package will be automatically generated by our automated pipelines at the time of building the registry site.
+> So there is no action for package authors for making sure API docs are generated other than making sure your package repository has
+> the right schema.json (or yaml) and that the metadata file for your package points to the correct schema file.
+
```
cnunciato, line 103: We should make this a note rather than a blockquote:

```suggestion
{{% notes %}}
API docs for your package will be automatically generated by our automated pipelines at the time of building the registry site. You do not need to take any action to generate API docs other than make sure your package repository has the right `schema.json` (or `.yaml`) and that the metadata file for your package points to the correct schema file.
{{% /notes %}}
```

I'd also suggest moving this block to the end of this section (so right above "Congratulations"), since as currently positioned, it breaks up the sentence, and its content doesn't really impact the steps you perform in the section.

cnunciato (changes requested):
Generally 👍, but would be good to use a `{{% notes %}}` instead of a blockquote and move that block to the end of the section.

themes/default/content/docs/guides/pulumi-packages/how-to-author.md
praneetloke: Thanks for the suggestion. I didn't end up moving it all the way to the end but I did move it to right after the steps for contributor and before the Pulumi employee steps begin.

cnunciato (approved):
That'll work too!

PR #997, opened 2022-01-22
Add providers and root namespace to documentation

iwahbe:
This adds documentation for some of my other PRs: https://github.com/pulumi/pulumi/issues/8796 and https://github.com/pulumi/pulumi/pull/8735.

### DIFF
--- a/themes/default/content/docs/intro/concepts/resources/options/_index.md
+++ b/themes/default/content/docs/intro/concepts/resources/options/_index.md
@@ -20,7 +20,7 @@ All resource constructors accept an options argument that provide the following 
 - [parent]({{< relref "parent" >}}): establish a parent/child relationship between resources.
 - [protect]({{< relref "protect" >}}): prevent accidental deletion of a resource by marking it as protected.
 - [provider]({{< relref "provider" >}}): pass an [explicitly configured provider]({{< relref "../providers/#explicit-provider-configuration" >}}), instead of using the default global provider.
-- [providers]({{< relref "providers" >}}): pass a set of [explicitly configured providers]({{< relref "../providers/#explicit-provider-configuration" >}}), These are used if provider is not given and are passed to child resources.
+- [providers]({{< relref "providers" >}}): pass a set of [explicitly configured providers]({{< relref "../providers/#explicit-provider-configuration" >}}). These are used if provider is not given, and are passed to child resources.
 - [replaceOnChanges]({{< relref "replaceonchanges" >}}): declare that changes to certain properties should be treated as forcing a replacement.
 - [transformations]({{< relref "transformations" >}}): dynamically transform a resource’s properties on the fly.
 - [version]({{< relref "version" >}}): pass a provider plugin version that should be used when operating on a resource.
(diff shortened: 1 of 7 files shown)


themes/default/content/docs/intro/concepts/resources/options/providers.md
```diff
@@ -0,0 +1,73 @@
+---
+title: "Providers"
+meta_desc: An in depth look at Pulumi resources and their usage.
+menu:
+  intro:
+    parent: options
+    weight: 10
+---
+
+The `providers` resource option sets a map of providers for the resource and its children. This map is combined with resource parents `providers` map. If no value is provided, the providers map is identical to the parents providers map. When checking determining which provider to use for a resource, the `providers` map is used if [provider]({{< relref "provider" >}}) is not supplied.
```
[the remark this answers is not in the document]
iwahbe: It should read "When determining". Thanks for catching that.

Frassle (approved):

themes/default/content/docs/intro/concepts/resources/options/providers.md
```diff
@@ -0,0 +1,73 @@
+---
+title: "Providers"
+meta_desc: An in depth look at Pulumi resources and their usage.
+menu:
+  intro:
+    parent: options
+    weight: 10
```
cnunciato, line 7: This should probably be changed to `11` (`10` is already in use by the provider page), and then `replaceonchanges.md`, `transformations.md`, and `version.md` changed to `12`, `13`, and `14`, respectively. Otherwise, this page won't appear in the left-hand nav.

themes/default/content/docs/intro/concepts/resources/options/_index.md
```diff
@@ -20,6 +20,7 @@ All resource constructors accept an options argument that provide the following
 - [parent]({{< relref "parent" >}}): establish a parent/child relationship between resources.
 - [protect]({{< relref "protect" >}}): prevent accidental deletion of a resource by marking it as protected.
 - [provider]({{< relref "provider" >}}): pass an [explicitly configured provider]({{< relref "../providers/#explicit-provider-configuration" >}}), instead of using the default global provider.
+- [providers]({{< relref "providers" >}}): pass a set of [explicitly configured providers]({{< relref "../providers/#explicit-provider-configuration" >}}), These are used if provider is not given and are passed to child resources.
```
cnunciato, line 23:
```suggestion
- [providers]({{< relref "providers" >}}): pass a set of [explicitly configured providers]({{< relref "../providers/#explicit-provider-configuration" >}}). These are used if provider is not given, and are passed to child resources.
```

themes/default/content/docs/intro/concepts/resources/options/providers.md
```diff
@@ -0,0 +1,73 @@
...
+---
+
+The `providers` resource option sets a map of providers for the resource and its children. This map is combined with resource parents `providers` map. If no value is provided, the providers map is identical to the parents providers map. When determining which provider to use for a resource, the `providers` map is used if [provider]({{< relref "provider" >}}) is not supplied.
```
cnunciato, line 10:
```suggestion
The `providers` resource option sets a map of providers for the resource and its children. This map is combined with resource parent's `providers` map. If no value is provided, the providers map is identical to the parent's providers map. When determining which provider to use for a resource, the `providers` map is used if [provider]({{< relref "provider" >}}) is not supplied.
```

cnunciato (changes requested):
Thanks for taking the time to do this! Few little things we should fix up before merging.

cnunciato (approved):
@iwahbe Thanks for making those changes. 🍻 I see the "Providers" item still isn't showing up in the left-hand nav, but after some digging it looks like that's because the name "Providers" is also being used in the parent menu, so by default, Hugo is suppressing it. The fix for this is to add an [`identifier`](https://gohugo.io/content-management/menus/#nesting) property to each menu item, but doing that for all of the items in this menu isn't something you should have to do, so I'll do it in a follow up and tag you on it. 😄  (The page is being linked from the top-level page, so it'll be discoverable and indexable, and I should be able to get a follow-up submitted today.) Feel free to merge when ready.

PR #998, opened 2022-01-23
Fix typo on iac marketing page

susanev:
Someone reported this through community chats, so fixing.
updates `what-is/what-is-infrastructure-as-code/`

@nimbinatus does this transcript live elsewhere that we need to update?


cnunciato (approved):

nimbinatus:
> @nimbinatus does this transcript live elsewhere that we need to update?

It's on YouTube's captions, so I updated that, but that should be it. Thanks for the heads up! These all are coming from the captioning service, so I'll have to comb them a bit more thoroughly next time one comes in.

PR #1009, opened 2022-01-28
Remove block_external_search setting from Learn content

cnunciato:
Early on, I included the `block_external_search_index` setting in our module and topic archetypes thinking we might want to publish and iterate on content (behind a "coming soon" home page or something) without risk of exposing that content it to crawlers. Now that we've launched, though, and we have PR previews, etc., we no longer need it.


nimbinatus (approved):
thank you!

Issue #1012, opened 2022-01-31
blog/dynamic-providers broken link


dixler:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details
Broken docs link at the bottom of the dynamic providers [blog](https://www.pulumi.com/blog/dynamic-providers/)

broken link:

> [docs](https://www.pulumi.com/docs/intro/concepts/resources/#dynamicproviders)

### Steps to reproduce

1. go to [page](https://www.pulumi.com/blog/dynamic-providers/)
2. click docs link at the bottom.

Expected:  link works
Actual:  link doesn't

PR #1013, opened 2022-01-31
fixed broken link in blog/dynamic-providers/

dixler:
I was just reading the docs. Haven't built the doc-site before.

fixes https://github.com/pulumi/pulumi-hugo/issues/1012


susanev (approved):
thank you!

@cnunciato i wonder if we have more of these around that need fixing?

PR #1015, opened 2022-01-31
Update the Team Page

cnunciato:
Adds Aseem Datar, removes Anu Sharma. 

Also fixes this, which I have not been able to unsee:

image


susanev:
is that a single purple pixel?

PR #1019, opened 2022-01-31
move self-hosted repos note up

susanev:
the note that says why the links don't work is below the actual links which make it difficult for folks to be aware that the links wont work. this moves it up, above the table, with the hope that folks might read it before trying to use the links.


pierskarsenbarg (approved):
Much more readable 👍

Issue #996, opened 2022-01-21
Add png downloads to brand page


aaronkao:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Current brand page (https://www.pulumi.com/brand/) only has SVGs to download. Would like to have PNGs for external users that are less design inclined.

susanev:
oh @cnunciato, Aaron messaged me about this, did he also message you? happy for you to take this if you'd prefer.

cnunciato:
@susanev Feel free to take it if you like -- I think this was filed before we synced up about this stuff, so I just took it assuming I'd do it. Up to you.

susanev:
cool ill do it today, i honestly totally forgot and meant to do it last week.

PR #1022, opened 2022-02-01
add png logos to brand page

susanev:
fixes https://github.com/pulumi/pulumi-hugo/issues/996

page being updated is at `brand/`


kimberleyamackenzie (approved):
🎊

PR #1024, opened 2022-02-02
Redirect /crosswalk

cnunciato:
We don't currently have any content on this page, and it's been search-engine indexed, so we should probably send those who find it to our current Crosswalk for AWS docs.

This change does two things:
* Asks Google not to index /crosswalk
* Redirects to https://www.pulumi.com/docs/guides/crosswalk/aws/

If we do plan to fill this page out with content, though, we might want to redirect temporarily rather than permanently.

### DIFF
--- a/themes/default/content/crosswalk/_index.md
+++ b/themes/default/content/crosswalk/_index.md
@@ -1,5 +1,7 @@
 ---
 title: Pulumi Crosswalk
 meta_desc: Pulumi Crosswalk is a collection of libraries that use well-architected best practices to make common infrastructure-as-code tasks easier and more secure.
-private: true
+layout: aws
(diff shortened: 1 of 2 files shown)


susanev (approved):
thank you!

cnunciato:
Ok, changed things up a bit here:

* Added an optional `redirect_temporarily` frontmatter param that resolves to a delayed meta-refresh, which Google will interpret as a temporary redirect. 
* Removed the no-index instruction, as we may want to index this page at some point later.
* Changed the redirect to /crosswalk/aws, since this page is the parent of /crosswalk/aws and /crosswalk/kubernetes.
* Applied the `aws` layout to make the redirect less jarring for the user.

PR #1042, opened 2022-02-09
Add TypeScript to the Pulumi Fundamentals pathway

cnunciato:
This change adds TypeScript to the Pulumi Fundamentals pathway, along with a new shortcode to display the appropriate language host based on the selected language.

### DIFF
--- a/themes/default/content/learn/pulumi-fundamentals/configure-and-provision/index.md
+++ b/themes/default/content/learn/pulumi-fundamentals/configure-and-provision/index.md
@@ -35,7 +35,19 @@ variables at the top of your program.
 Add the following configuration variables to your Pulumi program below the
 imports:
 
-{{< chooser language "python" / >}}
+{{< chooser language "typescript,python" / >}}
+
+{{% choosable language typescript %}}
+
+```typescript
+// get configuration
+const config = new pulumi.Config();
+const frontendPort = config.requireNumber("frontendPort");
+const backendPort = config.requireNumber("backendPort");
+const mongoPort = config.requireNumber("mongoPort");
+```
+
+{{% /choosable %}}
 
 {{% choosable language python %}}
 
@@ -50,7 +62,49 @@ mongo_port = config.require_int("mongo_port")
 
 Your Pulumi program should now match this code:
 
-{{< chooser language "python" / >}}
+{{< chooser language "typescript,python" / >}}
+
+{{% choosable language typescript %}}
+
+```typescript
+import * as pulumi from "@pulumi/pulumi";
+import * as docker from "@pulumi/docker";
+
+// get configuration
+const config = new pulumi.Config();
+const frontendPort = config.requireNumber("frontendPort");
+const backendPort = config.requireNumber("backendPort");
+const mongoPort = config.requireNumber("mongoPort");
+
+const stack = pulumi.getStack();
+
+const backendImageName = "backend";
+const backend = new docker.Image("backend", {
+    build: {
+        context: `${process.cwd()}/app/backend`,
+    },
+    imageName: `${backendImageName}:${stack}`,
+    skipPush: true,
+});
+
+// build our frontend image!
+const frontendImageName = "frontend"
+const frontend = new docker.Image("frontend", {
+    build: {
+        context: `${process.cwd()}/app/frontend`,
+    },
+    imageName: `${frontendImageName}:${stack}`,
+    skipPush: true,
+});
+
+// build our mongodb image!
+const mongoImage = new docker.RemoteImage("mongo", {
+    name: "mongo:bionic",
+});
+
+```
+
+{{% /choosable %}}
 
 {{% choosable language python %}}
 
@@ -92,8 +146,6 @@ mongo_image = docker.RemoteImage("mongo", name="mongo:bionic")
 Try and run your `pulumi up` again at this point. You should get an error like
 this:
 
-{{< chooser language "python" / >}}
-
 {{% choosable language python %}}
 
 ```bash
@@ -111,11 +163,27 @@ Remember how we can use the same program to define multiple stacks? Let's set
 the ports for this stack, which the Pulumi command line knows already from when
 you first initialized the project (it's the `dev` stack by default):
 
+{{< chooser language "typescript,python" / >}}
+
+{{% choosable language typescript %}}
+
+```bash
+pulumi config set frontendPort 3001
+pulumi config set backendPort 3000
+pulumi config set mongoPort 27017
+```
+
+{{% /choosable %}}
+
+{{% choosable language python %}}
+
 ```bash
 pulumi config set frontend_port 3001
 pulumi config set backend_port 3000
 pulumi config set mongo_port 27017
 ```
+
+{{% /choosable %}}
 
 This set of commands creates a file in your directory called `Pulumi.dev.yaml`
 to store the configuration for this stack.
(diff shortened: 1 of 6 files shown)


themes/default/content/learn/pulumi-fundamentals/create-a-pulumi-project/index.md
```diff
@@ -46,7 +46,15 @@ you need, so let's use that command. Use Python for this tutorial (TypeScript,
 Go, and C# are coming soon!), and the `-y` flag answers "yes" to the prompts to
```
nimbinatus, line 46: Line before this needs to be updated.

themes/default/content/learn/pulumi-fundamentals/configure-and-provision/index.md
```diff
@@ -111,12 +163,28 @@ Remember how we can use the same program to define multiple stacks? Let's set
 the ports for this stack, which the Pulumi command line knows already from when
 you first initialized the project (it's the `dev` stack by default):
 
+{{< chooser language "typescript,python" / >}}
```
rawkode, line 166: Does this actually need a chooser?

themes/default/content/learn/pulumi-fundamentals/create-docker-images/index.md
```diff
@@ -78,6 +78,20 @@ host machine.
 
 ## Build your Docker Image with Pulumi
 
+{{% choosable language typescript %}}
+
+Before we start writing our Pulumi program, we need to install the right
+provider. In this case, we want to use the `@pulumi/docker` provider for Python,
```
nimbinatus, line 84:
```suggestion
provider. In this case, we want to use the `@pulumi/docker` provider for Node.JS,
```

themes/default/content/learn/pulumi-fundamentals/create-docker-images/index.md
```diff
@@ -78,6 +78,20 @@ host machine.
 
 ## Build your Docker Image with Pulumi
 
+{{% choosable language typescript %}}
```
nimbinatus, line 81: Random question: When I switch back and forth on the choosable, the text column width also changes. It looks like the code box doesn't have a horizontal scroll. Think that's worth fixing?

themes/default/layouts/shortcodes/langfile.html
```diff
@@ -2,7 +2,7 @@
 <pulumi-chooser type="language" options="javascript,typescript,python,go,csharp,fsharp,visualbasic" option-style="none" class="inline">
 <pulumi-choosable type="language" value="javascript"><code>index.js</code></pulumi-choosable>
 <pulumi-choosable type="language" value="typescript"><code>index.ts</code></pulumi-choosable>
-<pulumi-choosable type="language" value="python">`__main__.py`</pulumi-choosable>
+<pulumi-choosable type="language" value="python"><code>__main__.py</code></pulumi-choosable>
```
nimbinatus, line 5: Odd; I had switched it to use the markdown because the dunders weren't appearing properly (they were causing things to get underlined, instead). I'm not sure why they're suddenly appearing properly 🤔  Has there been another change somewhere?

themes/default/content/learn/pulumi-fundamentals/configure-and-provision/index.md
```diff
@@ -92,8 +146,6 @@ mongo_image = docker.RemoteImage("mongo", name="mongo:bionic")
 Try and run your `pulumi up` again at this point. You should get an error like
 this:
```
nimbinatus, line 147: The error seems to have disappeared for the TS version. Does the TS version have a different error? If it doesn't, I think removing 149 and 159 should do it.

themes/default/content/learn/pulumi-fundamentals/configure-and-provision/index.md
cnunciato: It does, but only because we want the user to be able to choose the right project template for their language, and this is the first appearance of a language chooser in the pathway.

themes/default/content/learn/pulumi-fundamentals/configure-and-provision/index.md
```diff
@@ -213,7 +344,33 @@ We also need to create `Container` resources for the frontend and mongo
 containers. Put the `mongo_container` declaration just above the `backend_container` one, and the
```
nimbinatus, line 344: There's some differences here with snake_case and CamelCase for declarations for Python and TypeScript. I think it may confuse some readers. Messy, I know. Maybe that inline chooser might help here?

nimbinatus (changes requested):
A couple notes. Thank you so much!

themes/default/content/learn/pulumi-fundamentals/create-a-pulumi-project/index.md
cnunciato: Ah, thanks!

themes/default/layouts/shortcodes/langfile.html
cnunciato: Yes, I went ahead and fixed this to align with the others because it was showing up as ticks in other areas of the site:

image

The gist is that Hugo will render the content of shortcodes called with `{{% %}}`, but not for `{{< >}}`, which is why we see it rendered properly in Learn but not in the Getting Started guide. There are a few places where we call `langfile` with `%`s, so those'll need to be fixed up as well (one of which is my last blog post 🤦 -- I'll do that in this PR also), but as long as we use `{{< langfile >}}` going forward, we should be good.

themes/default/content/learn/pulumi-fundamentals/configure-and-provision/index.md
cnunciato: Ah thanks, no, that's my oversight -- the same error occurs with TS. I'll fix, thanks for spotting.

themes/default/content/learn/pulumi-fundamentals/configure-and-provision/index.md
cnunciato: Yep, another oversight, good spot. I'll add a `typescript` block for it.

themes/default/content/learn/pulumi-fundamentals/create-docker-images/index.md
nimbinatus: Sure!
A | B
-- | --
Screen Shot 2022-02-09 at 4 18 04 PM | Screen Shot 2022-02-09 at 4 18 10 PM
Screen Shot 2022-02-09 at 4 18 18 PM | Screen Shot 2022-02-09 at 4 18 25 PM

themes/default/layouts/shortcodes/langfile.html
nimbinatus: Ahhhhhh, so that makes sense. I guess the other pages I checked when I first checked the preview on that PR must have used the other version that I was attempting to use. So Hugo renders ones with % % as markdown inside, but the ones with < > as standard code with a markdown override?

cnunciato: I _think_ so, if I understand you correctly. The gist is that with percents, you're saying, "run this content through the Markdown renderer". Otherwise it just treats content as regular HTML.

themes/default/content/learn/pulumi-fundamentals/create-docker-images/index.md
cnunciato: Ah, okay. It looks like the code is extending the width of the column as much as it can and _then_ side scrolling. I'll make an issue for this on the theme repo, which is probably the right place for a fix for this. (We probably just need to apply an explicit full width to these code blocks.)

themes/default/layouts/shortcodes/langfile.html
nimbinatus: Ha, sorry. Words, ya know? That's exactly what I meant, so it makes sense :)

cnunciato:
@nimbinatus I think this is good to go -- I found a few more little things as well, and I think everything's been addressed. Let me know what you think!

PR #1045, opened 2022-02-11
chore: fix the type name for boolean

frezbo:
Fix the type name for boolean

Add docs on referncing a resource from the same document


cnunciato (approved):
LGTM, but would be good to get an approval from @pulumi/platform as well.

PR #1052, opened 2022-02-15
remove hyphen in pricing intro

susanev:
this came up in a review, we don't have the hyphen in the heading so its a bit weird to have it in the paragraph

@kimberleyamackenzie figured id just push this out now instead of waiting on the other release, but letting you know in case it messes anything up for you


kimberleyamackenzie (approved):
thank you for the heads up! i would have been staring at the rebase for many minutes trying to notice the difference

PR #1058, opened 2022-02-17
Fix s3.Bucket.get usage

hanchchch:
The second argument of `aws.s3.Bucket.get` method should be the `id` of the bucket, not`arn`. 
Following the document may cause the error `InvalidARNError`.


cnunciato (approved):
Thanks for the fix, @hanchchch!

PR #1061, opened 2022-02-17
Add TypeScript to Learn > Abstraction & Encapsulation

cnunciato:
This PR adds TypeScript language support to the Building with Pulumi pathway.

### DIFF
--- a/themes/default/content/learn/abstraction-encapsulation/component-resources/index.md
+++ b/themes/default/content/learn/abstraction-encapsulation/component-resources/index.md
@@ -29,7 +29,7 @@ A resource in Pulumi terms is basically a building block. That building block co
 
 Why would you make a component resource instead of just a "simple" logical grouping that you can call and use like a generic library full of classes and functions? A component resource shows up in the Pulumi ecosystem just like any other resource. That means it has a trackable state, appears in diffs, and has a name field you can reference that refers to the entire grouping.
 
-We've actually already started creating a component resource in the encapsulation part when we made a new class that built up an `s3.Bucket` and an `s3.BucketPolicy` with a `define_policy` function. Let's now turn that logical grouping into an actual component resource.
+We've actually already started creating a component resource in the encapsulation part when we made a new class that built up an `s3.Bucket` and an `s3.BucketPolicy`. Let's now turn that logical grouping into an actual component resource.
 
 ## Converting to a component resource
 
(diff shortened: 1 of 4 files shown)


themes/default/content/learn/abstraction-encapsulation/component-resources/index.md
```diff
@@ -29,18 +29,91 @@ A resource in Pulumi terms is basically a building block. That building block co
 
 Why would you make a component resource instead of just a "simple" logical grouping that you can call and use like a generic library full of classes and functions? A component resource shows up in the Pulumi ecosystem just like any other resource. That means it has a trackable state, appears in diffs, and has a name field you can reference that refers to the entire grouping.
 
-We've actually already started creating a component resource in the encapsulation part when we made a new class that built up an `s3.Bucket` and an `s3.BucketPolicy` with a `define_policy` function. Let's now turn that logical grouping into an actual component resource.
+We've actually already started creating a component resource in the encapsulation part when we made a new class that built up an `s3.Bucket` and an `s3.BucketPolicy`. Let's now turn that logical grouping into an actual component resource.
```
cnunciato, line 32: I only did this to avoid having to duplicate the paragraph, since the TypeScript example doesn't use this additional step. (It just defines the policy in the constructor.)

themes/default/content/learn/abstraction-encapsulation/encapsulation/index.md
```diff
@@ -24,6 +24,23 @@ Breaking apart Pulumi code is essential to keeping a clean, well-organized, and
...
+export const bucketName = bucket.id;
+```
+
+{{% /choosable %}}
+
+{{% choosable language python %}}
+
 ```python
 import pulumi
 import pulumi_aws_native as aws_native
```
cnunciato, line 46: Out of curiosity, why did we choose AWS Native here, and not just AWS Classic? I went with just-Classic for the TypeScript example, and only realized later that the Python example used both. Intentional? If so, why?

themes/default/content/learn/abstraction-encapsulation/component-resources/index.md
```diff
@@ -81,14 +154,28 @@ class OurBucketComponent(pulumi.ComponentResource):
         )
         return bucket_policy
 
+bucket1 = OurBucketClass('laura-bucket-1', 'default')
+bucket1.set_policy()
 
-
+pulumi.export("bucket_name", bucket1.bucket.id)
```
cnunciato, line 160: @nimbinatus and I discovered these lines are missing from what's in production now. 😬 Added here and in one other place.

themes/default/content/learn/abstraction-encapsulation/component-resources/index.md
nimbinatus: Gotcha; makes sense.

themes/default/content/learn/abstraction-encapsulation/component-resources/index.md
```diff
@@ -29,18 +29,91 @@ A resource in Pulumi terms is basically a building block. That building block co
...
+            Statement: [{
+                ...this.policies[policyType],
+                Resource: [
+                    pulumi.interpolate`${this.bucket.arn}/*`,
+                ],
+            }],
+        }
+    };
+
+    constructor(name: string, args: { policyType: PolicyType }, opts?: pulumi.ComponentResourceOptions) {
```
nimbinatus, line 81: Does TypeScript do exception handling in line, or does it come with the super class? I handled the exceptions in the Python code, so I was wondering if it made sense to handle them here, as well (unless TS doesn't do that).

themes/default/content/learn/abstraction-encapsulation/component-resources/index.md
nimbinatus: :shame: sorry! Copy-paste error on my part.

themes/default/content/learn/abstraction-encapsulation/component-resources/index.md
```diff
@@ -81,14 +154,28 @@ class OurBucketComponent(pulumi.ComponentResource):
...
+
+With the call to `super()`, we pass in a name for the resource, which [we recommend]({{< relref "/docs/intro/concepts/resources/components#authoring-a-new-component-resource" >}}) being of the form `<package>:<module>:<type>` to avoid type conflicts since it's being registered alongside other resources like the Bucket resource we're calling (`aws:s3:Bucket`).
+
+{{% choosable language python %}}
 
 That last call in the init, `self.register_outputs({})`, passes Pulumi the expected outputs so Pulumi can read the results of the creation or update of a component resource just like any other resource, so don't forget that call! You can [register default outputs using this call]({{< relref "/docs/intro/concepts/resources/components#registering-component-outputs" >}}), as well. It's not hard to imagine we will always want the bucket name for our use case, so we pass that in as an always-given output for our component resource.
 
+{{% /choosable %}}
+
+{{% choosable language typescript %}}
```
nimbinatus, line 173: Side note: I had started some work on an inline choosable to allow for small paragraph changes. Maybe we want to explore that further at some point? Happy to discuss it.

themes/default/content/learn/abstraction-encapsulation/encapsulation/index.md
nimbinatus: Trying to future-proof a bit and demonstrate that you can use them side-by-side, mainly. It's a common question about blending providers, and more examples that you can are good to have.

nimbinatus (changes requested):
Couple notes and questions

themes/default/content/learn/abstraction-encapsulation/component-resources/index.md
cnunciato: Yeah that'd definitely be handy! We should just do it. 😄

themes/default/content/learn/abstraction-encapsulation/component-resources/index.md
cnunciato: Ah, yes -- if you mean the `KeyError` bit, then yes, we should actually be protected from these kinds of exceptions by the compile-time type-checking we get against the keys in `policies`.

nimbinatus: Ah, ok. Makes sense. I'll resolve.

themes/default/content/learn/abstraction-encapsulation/component-resources/index.md
nimbinatus: Opened! https://github.com/pulumi/pulumi-hugo/issues/1091

nimbinatus (approved):
+1, LGTM. Thank you!

PR #1065, opened 2022-02-22
Update description of startTime for audit logs API

MitchellGerdisch:
Make it clear that audit logs that are OLDER than startTime are returned.


phillipedwards (commented):
reviewed.

PR #1067, opened 2022-02-22
Add host and sitemap to robots.txt

zchase:
Fixes: https://github.com/pulumi/home/issues/1838

### DIFF
--- a/themes/default/layouts/robots.txt
+++ b/themes/default/layouts/robots.txt
@@ -7,4 +7,4 @@ Disallow: *
 Disallow: /docs/search/$
 
 Sitemap: https://www.pulumi.com/sitemap.xml
-Host: https://www.pulumi.com
+Host: www.pulumi.com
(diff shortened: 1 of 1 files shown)


themes/default/layouts/robots.txt
```diff
@@ -5,3 +5,6 @@ Disallow: *
 
 # Disallow certain pages that are preceived as soft-404s.
 Disallow: /docs/search/$
+
+Sitemap: https://www.pulumi.com/sitemap.xml
+Host: https://www.pulumi.com
```
cnunciato, line 10: Looking at https://developers.google.com/search/docs/advanced/robots/robots_txt#syntax, I don't see `Host` listed as a property Google supports, and I don't see it listed in the spec at https://datatracker.ietf.org/doc/html/draft-rep-wg-topic-00 (assuming I'm not looking in an outdated place). @Brant-Soler is there guidance that explains this prop in more detail?

cnunciato: Looks like based on [this recommendation](https://yoast.com/ultimate-guide-robots-txt/#host-directive), we should probably omit the protocol bit:

```suggestion
Host: www.pulumi.com
```

cnunciato (approved):
One recommendation to tweak the `Host` directive, otherwise 👍!

Brant-Soler:
Yes, absolutely. Make it so...

PR #1070, opened 2022-02-23
shorten content on learn card

susanev:
## overview
one of the 3 learn cards has a longer description than the rest; we should try to keep these to no more than 3 lines. @nimbinatus happy to adjust the content in a different way if ya don't think my edit will work.

#### before
Screen Shot 2022-02-23 at 10 27 14 AM

#### after
Screen Shot 2022-02-23 at 10 27 08 AM


nimbinatus (approved):
I think this is perfect; thanks! For now, we'll just keep it in mind (I might leave a comment on the template 🤔  ), but eventually I'd love to have it be part of the linter.

PR #1093, opened 2022-02-25
fix  azure ad logo

susanev:
## overview
idk where this logo came from but its broken ive tried my best to fix it
each piece of text was independently crooked 🙃 

### before
befre-ad

### after
after-ad


kimberleyamackenzie (approved):
That before and after is extremely satisfying!

PR #1098, opened 2022-02-26
remove draft from blog template

susanev:
this removes the draft: true from our blog template.
since ive been here ive never seen anyone merge a blog set to draft on purpose.
but i have seen a lot of folks merge it without wanting it to be a draft.
and its slow to get out the fix to undraft it.

alternatively, i can leave draft in and set it to false if we think having it there so folks know about it would be helpful

@cnunciato do you have negative feelings about this? is keeping draft: true important to you?

### DIFF
--- a/BLOGGING.md
+++ b/BLOGGING.md
@@ -60,7 +60,7 @@ including all the required frontmatter parameters.
    ---
    ```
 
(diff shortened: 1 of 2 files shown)


cnunciato:
@susanev Not at all, totally fine with removing it. There are a few other mentions of how `draft` works in the blogging README, though, so it'd be good to make the appropriate adjustments to that stuff as well; I'm happy to do that if you'd rather not, up to you.

susanev:
@cnunciato k wasnt sure if that info was still helpful in case anyone does go and add draft on their own. i went ahead and removed the mentions, let me know if anything doesn't make sense. or feel free to push a commit.

cnunciato (approved):
Looks great -- thanks! :shipit:

PR #1101, opened 2022-02-28
add info on how to add urself to the website

susanev:
i know this is in onboarding docs in a few different places, but it seemed most relevant to have this content available in the repo where you add it. @cnunciato any thoughts on this?

### DIFF
--- a/pulumi-employees/add-yourself-to-the-website.md
+++ b/pulumi-employees/add-yourself-to-the-website.md
@@ -1,0 +1,8 @@
+# Add yourself to the website
+
+Clone the [Pulumi Hugo repo](https://github.com/pulumi/pulumi-hugo) locally, create a new branch (e.g. `yourfirstname/team`), and add the following files in a new commit:
+
+1. A TOML file to the [team data](https://github.com/pulumi/pulumi-hugo/tree/master/themes/default/data/team/team) directory
+1. A square sized picture to the [team images](https://github.com/pulumi/pulumi-hugo/tree/master/themes/default/static/images/team) directory.
+
+Then push your branch and open a pull request against master.


cnunciato:
@susanev could you point me to the places where these instructions already exist? I don't think I've seen them, so would be good to know where they are and what's in them.)

susanev:
@cnunciato its part of the instructions that Casie sends to folks that ive never actually seen i don't think, and its also part of the engineering checklist that some managers use for onboarding https://github.com/pulumi/home/issues/new?template=5-onboarding-checklist.md&title=Onboarding+Checklist%3A+%5Bdescription%5D

cnunciato:
Sorry for the delayed response on this. Chatted with Casie, and I agree it'd be good to have these docs close to the repo, but I'm thinking we should maybe start pulling this stuff into the repo wiki so it's a bit more navigable, easily consumable, updatable, etc. What would you think of that?

susanev:
@cnunciato i didnt see anything in the wiki for this repo. i know there's been back and forth in the eng org about when we should use wikis and when we should use markdown. i kinda would like to just get this in, and then maybe we should reach out to folks and make sure folks are mostly good with using the wiki for this stuff.

mattstratton:
Whenever these instructions that get into the onboarding and stuff happen, I would also recommend that there is some kind of option that is "if you don't have pulumi-hugo setup and don't want to, you can ask in x channel for someone to do it for you if you ask nicely" or something. setting up an entire environment for hugo, etc, when the only time someone will use it is for that one single PR seems like something that can be done elsewhere? or maybe casie already handles it?

susanev:
@mattstratton sure! these were grabbed directly from the instructions today but im totally down to update them.

@cnunciato do you feel strongly that this needs to be in the repo wiki?

mattstratton:
I'm actually curious what folks who don't use this repo do today - i'm gonna go out on an limb and say there's a nonzero number of employees who had it done for them, I wonder how they knew who to ask? :)

susanev:
Freddy used to open issues to add new folks, and then I believe Christian and Zack would open prs for those issues. I think that practice stopped, not totally sure what the process is now, Christian do you?

susanev:
@cnunciato would you prefer if i close this?

PR #1103, opened 2022-02-28
Add new images for upcoming registry change

kimberleyamackenzie:
Adding a couple of images that will be used for an upcoming change in the Registry: https://github.com/pulumi/registry/issues/577

See images below, but this change won't cause them to be used anywhere just yet:

Screen Shot 2022-02-28 at 9 25 08 AM

Screen Shot 2022-02-28 at 9 25 12 AM


anita-trimbur (approved):

susanev (approved):
@anita-trimbur @kimberleyamackenzie i think theres a little bit of extra padding around the check svg, was this intentional?

extra-padding

anita-trimbur:
@susanev, That's the default padding that comes with all the Material svgs, do you typically remove that for your icons? If yes, I can remove that so Kimberley can export it again.

susanev:
@anita-trimbur oh this is because we dont have access to the material icons in here, i see. its up to you. if the people icon doesn't also have it, it may be slightly more annoying to @kimberleyamackenzie to make them the same size, but will leave it up to yall to decide.

anita-trimbur:
@kimberleyamackenzie Let me know what's easier! I thought that the community icon had extra padding on top and bottom to keep it square, but no padding on the sides. The check icon would stay square even if we remove padding.

susanev:
k, looks like the people one lost that padding, here's what it looks like
community

kimberleyamackenzie:
I'm going to merge this as is - I have this all working locally + I think the icon looks like I'd expect without any major css magic, but @anita-trimbur let me know when you review if anything looks off and we can fix it up

PR #1106, opened 2022-03-01
Run full website builds for PRs

cnunciato:
This change adds a step to check out the docs repo and run a full website build for every PR commit. It should help us identify `{{< relref >}}` issues that would otherwise fail when pulled into pulumi/docs PRs later.

Note that this **adds** about 15 minutes to PR build times, and _does not_ push the full website in PR previews, as that'd take much longer. (Preview links will still consist of the same content they do today.) If we find that 15 minutes is too long (and I suspect we might), we could fairly easily convert this work to be done in response to a GitHub slash command later.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -40,20 +40,8 @@ jobs:
           role-to-assume: ${{ secrets.AWS_CI_ROLE_ARN }}
           role-session-name: PullRequestPreviewSession
 
-      - name: Check out the docs repo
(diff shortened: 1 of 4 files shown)


susanev (approved):

viveklak:
I wonder if we want to have this run in response to a command dispatch (e.g. `/full-site-preview`) instead? Something to consider if the constant 15 min. overhead is of concern... But generally this sounds much nicer, thanks for adding this!

PR #1107, opened 2022-03-01
Update PulumiUP landing page

zchase:


themes/default/layouts/page/pulumi-up-2022.html
```diff
@@ -0,0 +1,62 @@
...
+
+        <div class="fade"></div>
+        <div class="crawl">
+             <div class="title">
+                <p>PulumiUP</p>
+                <h1>The Cloud Awakens</h1>
+            </div>
+
+            <p class="mb-24">
+                Turmoil has engulfed the cloud community. The techniques used to solve the
```
pierskarsenbarg, line 19: I don't know when this is going live, but do we want to be talking about turmoil and factions right now?

zchase: Updated the text to remove any references to turmoil or factions.

Issue #1111, opened 2022-03-02
buttons broken on homepage on smaller screens.


susanev:
pulumi.com, atlassian case study section: buttons broken on smaller screens
<img width="492" alt="Screen Shot 2022-03-02 at 9 01 01 AM" src="https://user-images.githubusercontent.com/5489125/156410945-6353252f-64a2-4a75-aa78-9e956a302c11.png">

Issue #315, opened 2021-06-11
docs nav issues on smaller screens


susanev:
## whats the problem

on smaller screen, the docs navigation is shown at the very bottom of the page which is not easy to use, or friendly to users

## proposed fix

as described in https://github.com/pulumi/pulumi-hugo/pull/309#issuecomment-858828460 use a similar solution to the blog, a button to toggle the navigation in and out of view

PR #1116, opened 2022-03-03
Add docs nav toggle for small screens; fix overlapping buttons on small screens

kimberleyamackenzie:
This PR tackles two items affecting small screens:

1) Adds a toggle button for the docs nav at small screens.  Previously the nav content dropped to the bottom of the page which wasn't super clear or easy to use - this PR pulls in the same solution we use for the blog by adding a toggle button that shows the nav at the top of the page when toggled open.   https://github.com/pulumi/pulumi-hugo/issues/315

Since I have worked in docs less, would love some eyes to make sure I got everywhere the nav is (I checked both docs home and individual pages) - not sure if I might be missing something. 

2) Adjusts buttons for the Atlassian case study on the home page so they move to a column at small screen sizes.  https://github.com/pulumi/pulumi-hugo/issues/1111

Screen Shot 2022-03-02 at 4 39 20 PM

### DIFF
--- a/themes/default/layouts/docs/list.html
+++ b/themes/default/layouts/docs/list.html
@@ -3,7 +3,7 @@
         <div class="lg:flex">
 
             <div class="lg:hidden mb-6">
-                <button class="docs-sidebar-toggle items-center px-3 py-2 border rounded text-purple border-purple">
+                <button class="docs-sidebar-toggle items-center px-3 py-2 border rounded">
                     Toggle Docs Navigation
                 </button>
             </div>
--- a/themes/default/layouts/docs/list.html
+++ b/themes/default/layouts/docs/list.html
@@ -2,6 +2,23 @@
     <div class="container mx-auto px-4 py-8 mt-2">
         <div class="lg:flex">
 
+            <div class="lg:hidden mb-6">
+                <button class="docs-sidebar-toggle items-center px-3 py-2 border rounded text-purple border-purple">
+                    Toggle Docs Navigation
+                </button>
+            </div>
+
+            <div class="docs-sidebar-content hidden lg:block lg:w-3/12 pr-8 mb-2 pt-8 lg:order-first lg:border-none lg:pt-0 lg:mt-0">
+                <div class="sticky-sidebar">
+                    <div class="lg:mt-1 mb-6">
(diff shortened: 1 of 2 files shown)


susanev (approved):
works well for me!! thanks for fixing this, its a huge usability improvement!! 🌟

themes/default/layouts/docs/list.html
```diff
@@ -28,7 +45,7 @@ <h1>{{ .Title }}</h1>
                 </div>
             </div>
 
-            <div class="lg:w-3/12 pr-8 mb-2 pt-8 lg:order-first lg:border-none lg:pt-0 lg:mt-0">
+            <!-- <div class="hidden lg:block lg:w-3/12 pr-8 mb-2 pt-8 lg:order-first lg:border-none lg:pt-0 lg:mt-0">
```
cnunciato, line 48: Should we not just remove this?

susanev:
wait, why are the registry icons in this pr?

themes/default/layouts/docs/list.html
kimberleyamackenzie: oh 💯, i was looking at so much code that I missed not deleting this.  will get it + fix this up tomorrow, including dropped the two svgs @susanev noticed that are already in master.  i probably shoulda just taken a coffee break instead of tackling a new pr 😬

kimberleyamackenzie:
@susanev @cnunciato alright, the issues with the PR should be fixed up - sorry for the churn here!

susanev (approved):
:shipit:

themes/default/layouts/docs/list.html
```diff
@@ -2,6 +2,23 @@
     <div class="container mx-auto px-4 py-8 mt-2">
         <div class="lg:flex">
 
+            <div class="lg:hidden mb-6">
+                <button class="docs-sidebar-toggle items-center px-3 py-2 border rounded text-purple border-purple">
```
cnunciato, line 6:
```suggestion
                <button class="docs-sidebar-toggle items-center px-3 py-2 border rounded">
```

cnunciato (approved):
LGTM, only one little thing I noticed (some classes that look like they aren't doing anything).

PR #1139, opened 2022-03-04
Add track properties for pricing dialog

kimberleyamackenzie:
Add properties to the two links inside the pricing page "get started" dialog so that we can have insight into usage of this dialog to navigate to the different pages.


susanev (approved):
thank you!!

cnunciato (approved):

Issue #73, opened 2021-04-17
we need a third nav design


susanev:
on pages with a third nav, the nav is extremely easy to miss, we need a properly designed third nav

an example from the brand page (logos/colors is the nav)
https://www.pulumi.com/brand/
<img width="1915" alt="brand-page" src="https://user-images.githubusercontent.com/5489125/154309604-d8a8b643-e47f-4ca0-9bcd-a24f2ac9aa4d.png">

jonelleb:
### Design Goals
---
- Current third nav design is too easy to miss, we need a new design that brings more attention to it.

### Examples of Third Navs on the Site
---
- Brand Resources: https://www.pulumi.com/brand 
- Resources: https://www.pulumi.com/resources/#all
- Case Study Pages: https://www.pulumi.com/case-studies/  

### Additional Notes
---
We've decided that the Resources and Brand Resources pages will have a tab design similar to the design that we currently have for tabs, but with a few modifications. All throughout the site wherever tabs like this are used should be updated to keep consistency.  The Case Study pages will have more of a table of contents type of design and will not be using the new tab design.

susanev:
@jonelleb lets keep this issue for brand and resources only, and move that over to ready. and then open a new issue to tackle that case study pages.

jonelleb:
# Resources Page

[Figma File] (https://www.figma.com/file/VRCp6m8TyfCN4tZugJiyh7/Third-Nav-Design?node-id=31%3A548/ "Title")

[Figma Prototype] (https://www.figma.com/proto/VRCp6m8TyfCN4tZugJiyh7/Third-Nav-Design?node-id=41%3A2318&scaling=scale-down&page-id=31%3A548&starting-point-node-id=41%3A2318/ "Title")
**(Prototype shows hover states and gives examples of interactions)**

# 3rd Nav / Tab Design Changes
- The text for the active tab (w/ rainbow colored indicator line) now has a medium weight font.
- Icons have been added 8px to the left of text (using the same icons that are currently being used on this page)
- Inactive tabs no longer have indicator lines underneath text.
- I'm suggesting that we limit the number of links within the nav to 5. If there are more than 5 links, a carrot icon should appear on the right side of the nav that will allow the user to scroll to the right. 
- Left carrot icon appears once the user starts scrolling to the right. The left scroll icon disappears again when user scrolls back to the left; only appears when needed.

## Additional Page Changes
- A bigger size of the corresponding icon is next to the page title.

[...]

PR #1196, opened 2022-03-10
Third nav implementation on resources + brand resources pages

kimberleyamackenzie:
This PR is for implementing a new third nav design on the brand resources + resources pages.  The bulk of the code changes are in theme: https://github.com/pulumi/theme/pull/83, where I implemented a new reusable component for this nav functionality.  I used that component for the brand resources page, but had to implement the resources page separately because of how we use the nav as a filter there - but future implementations of a tertiary nav in this style should be easy with the new component.

### Changes in this PR:
- [ ] New third nav 🎉 
- [ ] On brand resources page, content is only shown for the active tab (this was already the resources page behavior)
- [ ] Scrollability on the nav in the brand resources page.  This is implemented with arrows on screens above 1024px, below that we remove those arrows (but maintain scrollability) because of how the arrow behavior gets kind of funky and hard to manage.
- [ ] Small adjustment to mobile display of brand resources page, with stacking buttons instead of side by side

Screen Shot 2022-03-10 at 10 23 00 AM
Screen Shot 2022-03-10 at 10 23 10 AM
Screen Shot 2022-03-10 at 10 25 30 AM
Screen Shot 2022-03-10 at 10 26 28 AM

### Links
Designs: https://www.figma.com/file/VRCp6m8TyfCN4tZugJiyh7/Third-Nav-Design?node-id=3%3A273 & https://www.figma.com/proto/VRCp6m8TyfCN4tZugJiyh7/Third-Nav-Design?node-id=3%3A1161&scaling=scale-down-width&page-id=3%3A273&starting-point-node-id=3%3A1161

Ticket: https://github.com/pulumi/pulumi-hugo/issues/73

### To-do
- [ ] This PR is connected to my theme PR.  Once these are ready to go in, I will update this to use the release branch of theme.

### DIFF
--- a/themes/default/content/blog/2022-03-10-hierarchical-config/index.md
+++ b/themes/default/content/blog/2022-03-10-hierarchical-config/index.md
@@ -182,7 +182,6 @@ encryptionKmsKey: string | *"arn:aws:kms:eu-west-2:..."
 issueEmail: string | *"bugs@ellingsonmineral.com"
 ```
 
-
 ## Pulumi Native Hierarchical Configuration
 
 So we've worked out that project level configuration is a data structure within our Pulumi programs and it works pretty darn well.
--- a/themes/default/content/resources/_index.md
+++ b/themes/default/content/resources/_index.md
@@ -7,4 +7,24 @@ aliases:
   - /webinar
   - /webinars
   - /events
+
+sections:
+    - label: All
+      anchor: all
+      icon: asterisk
+    - label: Featured
+      anchor: featured
+      icon: star
+    - label: Upcoming
+      anchor: upcoming
+      icon: users
+    - label: Videos
+      anchor: videos
+      icon: video
+    - label: PulumiTV
+      anchor: pulumitv
+      icon: tv
+    - label: Whitepaper
+      anchor: whitepaper
+      icon: book
 ---
--- a/themes/default/layouts/page/brand.html
+++ b/themes/default/layouts/page/brand.html
@@ -3,17 +3,13 @@
 {{ end }}
 
 {{ define "main" }}
-    <nav class="nav-section-sticky shadow">
+    <nav>
         <div class="container mx-auto">
-            <ul>
-                {{ range .Params.sections }}
-                    <li><a href="#{{ .anchor }}">{{ .label }}</a></li>
-                {{ end }}
-            </ul>
+            <pulumi-tertiary-nav items={{ .Params.sections | jsonify }}></pulumi-tertiary-nav>
         </div>
     </nav>
 
-    <section id="logos" class="mt-8 px-4 md:px-0">
+    <section id="logos" class="mb-8 px-4 md:px-0 hidden">
         <div class="container mx-auto max-w-4xl pt-8">
             <h2>Pulumi Logos</h2>
             <p>
@@ -28,11 +24,11 @@
                                 <img src="/logos/brand/{{ .file }}.svg" title="{{ .name }}">
                             </a>
                         </div>
-                        <div class="mt-6 flex flex-row place-content-center">
-                            <a class="btn btn-sm btn-brand-download mr-4" href="/logos/brand/{{ .file }}.svg" download>
+                        <div class="mt-6 flex flex-col sm:flex-row items-center place-content-center">
+                            <a class="btn btn-sm btn-brand-download mb-4 sm:mr-4 sm:mb-0 w-48" href="/logos/brand/{{ .file }}.svg" download>
                                 <span>Download SVG</span>
                             </a>
-                            <a class="btn btn-sm btn-brand-download" href="/logos/brand/{{ .file }}.png" download>
+                            <a class="btn btn-sm btn-brand-download w-48" href="/logos/brand/{{ .file }}.png" download>
                                 <span>Download PNG</span>
                             </a>
                         </div>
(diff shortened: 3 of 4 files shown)


cnunciato:
I'll need a bit of time to get through this PR and the accompanying theme changes (I've started, should be able to finish today), but a couple of things I've noticed:

* On the brand page, the tabs don't get pointer cursors (though they seem to on other pages)?
* Things seems to be getting clipped (hiding nav items) even when there's more than enough horizontal space to show all of them):
    image

    We should really only clip/hide these nav items when we absolutely have to, IMO, since they're the main way of understanding and using this page. Should be possible to continue doing this, at least on desktop (as we do today).

jonelleb:
@kimberleyamackenzie I think everything looks great! Thanks for the fast turn around! 
@cnunciato, Kimberley and I just spoke and really the issue with widening the nav space happens when there are scroll icons, so she's going to remove the scroll and show all six links and we'll only use the scroll icons when necessary (on pages that have over 6 links and/or long text). Thanks!

cnunciato (commented):
Ok, I think I have some high-level questions about this:

* We seem to be using `pulumi-tertiary-nav` to show and hide content, whereas today, we don't use the existing nav (the one this new nav is presumably intended to replace) to show and hide content -- we just show all content, and use the nav (which used to be sticky) to anchor to individual sections. I do think we should retain the behavior we have today, though; I don't think we should change things up to show only one section at a time. (Indeed, the first thing I did when I looked at the brand page was scroll down and wonder what'd happened to the colors.)

* The tertiary-nav component seems like it _wants_ to be a general-purpose tab-set component, which in principle I like (separate from using it as a nav replacement), but the component isn't quite written to work that way yet -- which makes me wonder what we're gaining by _having_ the component, given that it essentially relies on structured/page-specific frontmatter and is also aware of the global context (e.g., it uses jQuery). I'd kind of expect us to go either way -- either use jQuery alone (which would IMO be totally fine considering we're just toggling CSS classes at this point) or go full-on component and have _all_ of the display logic happen inside the component, exposing more of a slot-based API.

Just some initial thoughts; happy to sync up over Zoom as well if it'd help keep things moving.

jonelleb:
I did some research (one of them being on the Nielsen Norman Group site) and  found that in-page links are often confusing for users - normally links lead to another page, when that expectation isn’t met, users can get disoriented.
Also, users usually skim content so I think it would be better to separate the content into shorter, more manageable pieces instead having a large amount of lengthly content on one page. And a page with brief sections doesn’t require as much scrolling effort. Those were my reasons behind the new nav behavior on the Brand Resources page, but if keeping links as in-page/anchor type links is preferred that's fine.

susanev:
Yes, this was an intentional design choice. We present these as navigation but they are not navigation, they are anchor links. This is quite confusing and unexpected for users. There is also no known need for users to consume the colors and the logos at the exact same time, so there is not a need to present them on the page at the same time.

cnunciato:
Ok thanks y'all, I appreciate the discussion. I think I maybe got the wrong impression on intended scope here. 

If these design changes are meant to apply only to the Resources and Brand pages, then that's all good. I think because the Brand page (today) uses the same anchor-based subnavigation as our whitepaper and case-studies pages, and the issue linked from this PR mentions changing those pages as well, I was assuming the design changes in this PR would be applied to those pages also. It sounds like that may not be the case -- and if so, awesome, this all seems fine design-wise.

cnunciato:
Ok yeah -- now I see now that I missed everything from this comment down: https://github.com/pulumi/theme/issues/80#issuecomment-1063400956

Apologies for that!

kimberleyamackenzie:
@cnunciato I incorporated the majority of your feedback into the theme PR, and re-triggered a build here.  In addition to some of the smaller tweaks, I wrapped the intersection observer in a conditional to prevent being called on pages where the element doesn't exist and updated the resources page to show all 6 links at larger screen sizes when that's possible (see screenshot below).

> The tertiary-nav component seems like it wants to be a general-purpose tab-set component, which in principle I like (separate from using it as a nav replacement), but the component isn't quite written to work that way yet -- which makes me wonder what we're gaining by having the component, given that it essentially relies on structured/page-specific frontmatter and is also aware of the global context (e.g., it uses jQuery). I'd kind of expect us to go one way or the other, though -- either use jQuery alone (which IMO would be totally fine considering we're just toggling CSS classes at this point) or go full-on component and have all of the display logic happen inside the component, exposing more of a slot-based API of some sort.

I think one of the big benefits of having the component is it helps us avoid having to duplicate a lot of complicated css/js anywhere else we want to use this.  I agree that a slot-based API would be a little more generalizable, but that feels like a bit heavier of a lift than we need at the moment - I was just trying to solve the problem of being able to add this third nav to additional pages in a few lines, as opposed to recreating classes + styles each time we want to do so.  If you're open to it, I'd like to leave the component as-is for now, and revisit going more full-on component if and when we see that we're using this more broadly.

cnunciato (approved):
@kimberleyamackenzie Thanks for that jQuery tweak on the component; that definitely helps. Ok! Sounds like we're all on the same page with regard to the tradeoffs here, so I'm okay rolling forward with this. And thanks again for the discussion. 🚀

zchase (approved):
LGTM to but we should consider filing a follow-up issue to make `pulumi-tertiary-nav` more multi-purpose and a bit clearer on how to code it up. My personal preference is to have something like:

```
<pulumi-tertiary-nav>
    <section id="section-1">
    </section>

    // More children
</pulumi-tertiary-nav>
```

That way it is a bit clearer in the HTML what is actually being managed by the component. This sort of approach should also work for `/resources` as we can just create the static sections in HTML rather than do the JS sorting. That requires somewhat of a refactor for the `/resources` HTML code but something I'd be happy to tackle if there is a general-use component.

jonelleb (approved):
Everything looks great and is approved by me. Thanks for updating!

jonelleb:
@kimberleyamackenzie - Looks great!

PR #1211, opened 2022-03-15
De-list the thank-you page, remove some frontmatter

cnunciato:
Fixes https://github.com/pulumi/marketing/issues/121.


cnunciato:
@zchase I'm assuming it's okay to remove that frontmatter -- I didn't see it referenced anywhere in the confirmation layout -- but let me know if that isn't the case.

PR #1221, opened 2022-03-17
remove answerable from feedback msg

susanev:
noticed this today, _answerable_ reads as kinda rude, and im not sure how a user is supposed to know ahead of time if their question is _answerable_, so rephrased.


kimberleyamackenzie (approved):
🚢

PR #1226, opened 2022-03-18
Fix the PulumiUP date

cnunciato:
Updates the date and summary copy of the Pulumi Up event resource. Also makes a slight tweak to the hover behavior to underline only the title (instead all of the text on the card).

Fixes https://github.com/pulumi/marketing/issues/122.


zchase (approved):
Thank you!

SaraDPH (approved):

PR #1229, opened 2022-03-18
Whitepaper -> Whitepapers

cnunciato:
The Resources page's tab filters have an item labeled "Whitepaper" that should be "Whitepapers". (Yes, there's only one, but there will be more!)

Requires https://github.com/pulumi/theme/pull/88 which I'll merge simultaneously.

image


susanev (approved):
lol sry

PR #1242, opened 2022-03-21
Fix a broken link

cnunciato:


susanev (approved):

isaac-pulumi (approved):

isaac-pulumi (changes requested):
Appears to be a broken relref at line 677 in the preview but I can't tell why it's broken.

Upon further investigation, relrefs don't seem to work in preview but are fine in prod. Is that new behavior for preview?

isaac-pulumi:
> Appears to be a broken relref at line 677 in the preview but I can't tell why it's broken.

relrefs work in prod but don't in previews since we don't publish the entire site in preview per @cnunciato

isaac-pulumi (approved):

Issue #1220, opened 2022-03-17
add dev docs about swifttype


susanev:
we have had to relearn how to configure swifttype a few times as new folks contribute, lets get some docs added so other folks can pick the swifttype work up as needed

PR #1245, opened 2022-03-22
Update readme with information about Swiftype

kimberleyamackenzie:
Add some details about Swiftype and ways we have to adjust search results to the readme.  I added here since it's relevant to both docs and registry, but let me know if there's a better place for this info!

fixes https://github.com/pulumi/pulumi-hugo/issues/1220


susanev (approved):
yay, thank you!

i think at some point in the future we may want to consider breaking out some of this readme into other files. but nothing we need to tackle now!

PR #1252, opened 2022-03-22
Unlink the blog post title on the blog post page

cnunciato:
These needn't be linked, so this change just unlinks them.


susanev (approved):
approving in case you are blocked, looks good to me!

PR #1258, opened 2022-03-24
improving docs left-nav labels

susanev:
## overview

this is changing our converter labels to their full names which should better help folks find what they are looking for
this also updates the heading on the get started section to be more clear, ive aligned on get started instead of getting started but can switch to getting started if folks feel strongly

there was a previous convo around also updating the get started guide labels, but we decided against that for now

related [internal slack convo](https://pulumi.slack.com/archives/C85BS3LJZ/p1647957689762799)

### before
Screen Shot 2022-03-24 at 3 16 54 PM

### after
Screen Shot 2022-03-24 at 3 16 42 PM

### DIFF
--- a/themes/default/content/blog/introducing-crd2pulumi/index.md
+++ b/themes/default/content/blog/introducing-crd2pulumi/index.md
@@ -5,7 +5,7 @@ meta_desc: Generate Kubernetes CustomResource types in TypeScript, Python, C#, a
 menu:
   converters:
     identifier: crd2pulumi
-linktitle: crd2pulumi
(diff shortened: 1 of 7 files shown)


susanev:
@cnunciato random additional thought, maybe `Intro to Pulumi` instead of only `Intro`, thoughts?

cnunciato (approved):

cnunciato:
@susanev I actually kinda feel like "intro" isn't really the right title for that section, so I wouldn't be super inclined to add that particular change here.

PR #1261, opened 2022-03-25
Add support for livestream pages under Resources

mattstratton:
This is a WIP - this provides a single landing page for each episode of PulumiVision, including buttons/links for twitch, youtube, etc, as well as embedding the video after the stream is over, etc.


mattstratton:
a couple of things that are an issue currently:

- [ ] Adding the live stream menu item makes it all look wonky
- [ ] nothing shows up if you select livestreams, but the stream page shows up on the upcoming, etc

jonelleb:
Hi @mattstratton. Since I worked on the update of the Resources page Susan asked me to jump in with my thoughts. I was wondering if it would be possible to design those two new pages with a video at the top like the other video pages on the site are designed now (just to keep it consistent and also gives the user the option to play the live stream right away.) Maybe having the YouTube version being the video that's automatically there on the page and then giving them the choice of Twitch with the 'Watch on Twitch' button underneath the YouTube video on the page (kind of like how this page is set up: http://pulumi-hugo-origin-pr-1261-9771fb7b.s3-website.us-west-2.amazonaws.com/resources/from-zero-to-production-in-kubernetes )

And maybe the form being at the bottom? (Underneath the Get Started with Pulumi text and above Learn more and Contact Us). 

@susanev if you have any other suggestions or something that you think would work better, please jump in. Thanks!

mattstratton:
That's actually not a bad idea at all (embedding the video even on the upcoming - the pre-recorded does have the embedded video but it looks different than your example because the form is on the right)

this would help because we wouldn't actually need to make changes to the page when the stream was over; the "watch on twitch" could still be there, but would always just take them to our twitch channel (it would be a little confusing because twitch doesn't keep streams around for vieweing after a month or so)

definitely could move the form to the bottom; I will be a smaller form anyway. This was based on the workshops template which has the registration form on the right, but since the reminder form will be smaller, it could be pushed down

https://www.pulumi.com/resources/introduction-to-infrastructure-as-code/

mattstratton:
> adding the live stream menu item makes it all look wonky

the new resource page menu work fixed this for me :)

susanev:
@nimbinatus may pick this up in the future, closing for now.

Issue #1175, opened 2022-03-08
q1 ux team blog


susanev:
wed like to author a blog from the ux team this quarter
[deets on how to author a blog](https://github.com/pulumi/pulumi-hugo/blob/master/BLOGGING.md) @susanev is super happy to help with this though, pls reach out

[ux team blog ideas doc](https://docs.google.com/document/d/1MR4dOrgx_2ZG3TYmyWMUgi9-sitIIwpgKyKvoL4QvgM/edit?usp=sharing)

we should aim to publish this before the end of march, and we should share a draft doc in #blogs and #launch-readiness for feedback ahead of opening a pr

PR #1262, opened 2022-03-25
Adding UX blog

anita-trimbur:
fixes https://github.com/pulumi/pulumi-hugo/issues/1175


susanev:
@anita-trimbur the original meta.png can be removed, you can right-click delete in vscode
i think also our general practice is to save the meta image as meta.png so you may want to change the name of the meta img ur using to that.
once you've made those changes locally ya can push up another commit

kimberleyamackenzie (commented):
The content looks good to me, but I don't think the package.json and package lock file should need to be touched here - lemme know if I can help you drop those changes as part of this PR

kimberleyamackenzie (approved):
🚢 :shipit: 🎉 🎉

PR #1267, opened 2022-03-28
Docs: Github Action Fail Due to Python Version

Obleskar:
Naively upgrades every `python-version` in the Github Actions Continuous Delivery Guide (Python) from `3.6` to `3.9`. I've only tested it with the `push.yaml` action, so it might be better to do a smaller upgrade or limit the scope of the upgrades.

Issue for dropping Python 3.6 support: https://github.com/pulumi/pulumi/issues/8131 "When we do this, we should make sure to update the language docs to reflect the new reality."


stack72 (approved):
Thanks so much for the work here @Obleskar - I really appreciate it!

PR #1305, opened 2022-04-04
Fix link to self-hosted trial page

gthuang:
Updating link to self-hosted offering to trial requst page.


cnunciato (approved):

susanev:
Can we please fix the hover states on these buttons while we are here? The text almost disappears on hover.

gthuang:
> Can we please fix the hover states on these buttons while we are here? The text almost disappears on hover.

Can we do a separate PR/issue on this? Would like to get this link fixed first.

PR #1307, opened 2022-04-04
add a note about additions to the style guide

susanev:
adding a little note about adding things to the style guide
super open to feedback on my words

### DIFF
--- a/STYLE-GUIDE.md
+++ b/STYLE-GUIDE.md
@@ -1,6 +1,6 @@
 # Style Guide
 
-This document defines some general styles we adhere to in the docs.
+This document defines some general styles we adhere to in the docs. New additions to this document should already be accepted and in common practice. If they are not accepted or in common practice, start with an issue and relevant prs before adding to this style guide.
 
 ## Use Inclusive Language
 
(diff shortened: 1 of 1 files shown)


susanev:
It came up in convo with Luke. Would you rather it work the other way? meaning things get added to the style guide first and then the work to align to it happens?

jkodroff (approved):
Existing text works. I suggested an alternative wording.

cnunciato:
Thanks @jkodroff, that wording is better. Still, I'd like to see an example of what this workflow would actually look like in practice and how it's better than just submitting a PR that invites the same discussion. Say I wanted to propose using a particular kind of punctuation in some scenario or other. Is there some reason I shouldn't just submit a PR on this repo for that? What's better about filing an issue and a concurrent PR?

susanev:
@cnunciato lets chat more about this at our next 1-1. no urgency, ill add a do-not-merge label for now.

cnunciato:
@susanev Works for me, thanks. I’m sure there are good reasons behind the desire to have this, it’d just be clear about what they are. Thanks for the discussion.

susanev:
talked with @cnunciato going to hold off on this until we have the problem surface again.
general agreement than when you are adding something to the style guide that fundamentally changes how we've done things previously that should start with an issue, slack convos, zooms, or similar.

PR #1310, opened 2022-04-05
Additions to Projects and Stacks page

mattstratton:

### DIFF
--- a/themes/default/content/docs/guides/organizing-projects-stacks/_index.md
+++ b/themes/default/content/docs/guides/organizing-projects-stacks/_index.md
@@ -283,7 +283,6 @@ Within your Pulumi project, there are good practices to consider to help keep yo
 
 {{< chooser language "typescript,go" / >}}
 
-
 {{% choosable language typescript %}}
 Organize your code in a way that makes it easy to understand and maintain. One way to do this in Typescript is to break out your code into separate files, and then import them into your main file. In this example, the entrypoint for our Pulumi program is `index.ts`, but we use the `utils.ts` file for supporting functions.
 
--- a/themes/default/content/docs/guides/organizing-projects-stacks/_index.md
+++ b/themes/default/content/docs/guides/organizing-projects-stacks/_index.md
@@ -121,6 +121,7 @@ This project looks a bit like this:
 ```
 ├─ infrastructure
   ├── main.go
+  ├── Pulumi.yaml
   ├── Pulumi.dev.yaml
   ├── Pulumi.staging.yaml
   └── Pulumi.prod.yaml
@@ -137,11 +138,13 @@ Our example service is made up of an API and a database (RDS, CosmosDB, etc.). O
 ```
 ├── infrastructure
 │   ├── main.go
+│   ├── Pulumi.yaml
 │   ├── Pulumi.dev.yaml
 │   ├── Pulumi.staging.yaml
 │   └── Pulumi.prod.yaml
 ├── myApp
 │   ├── main.go
+│   ├── Pulumi.yaml
 │   ├── Pulumi.dev.yaml
 │   ├── Pulumi.staging.yaml
 │   └── Pulumi.prod.yaml
@@ -161,10 +164,33 @@ These component resources can be packaged up and stored alongside all of your ot
 ```
 ├── infrastructure
 │   ├── main.go
+│   ├── Pulumi.yaml
 │   ├── Pulumi.dev.yaml
 │   └── Pulumi.prod.yaml
 ├── myApp
 │   ├── main.go
+│   ├── Pulumi.yaml
+│   ├── Pulumi.dev.yaml
+│   ├── Pulumi.staging.yaml
+│   └── Pulumi.prod.yaml
+├── pkg
+│   └──application
+│     └── app.go
+└── .etc
+```
+
+To be clear, each of the applications/services inside our monorepo (and the `infrastructure` project is treated the same way as each independent application/service) are a separate Pulumi project, with their own stacks, and their own `Pulumi.yaml`. Given that each service is a separate Pulumi project, they can all use different programming languages. Let's take a look at how it might look if the `infrastructure` team prefers to write in Go, and the myApp team prefers TypeScript:
+
+```
+├── infrastructure
+│   ├── main.go
+│   ├── Pulumi.yaml
+│   ├── Pulumi.dev.yaml
+│   └── Pulumi.prod.yaml
+├── myApp
+│   ├── index.ts
+│   ├── package.json
+│   ├── Pulumi.yaml
 │   ├── Pulumi.dev.yaml
 │   ├── Pulumi.staging.yaml
 │   └── Pulumi.prod.yaml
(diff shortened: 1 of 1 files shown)


mattstratton:
This is still a work in progress, but it's the beginnings. I would appreciate suggestions on other patterns for code organization, and also, if anyone can add / suggestion support for python in a pythonic way (just dump ideas as comments on this PR, no need to submit the full commit, etc)

themes/default/content/docs/guides/organizing-projects-stacks/_index.md
```diff
@@ -158,13 +205,64 @@ Back to our example, our service needs a database and a subnet (or other network
...
+│   └── Pulumi.prod.yaml
+├── pkg
+│   └──application
+│     └── app.go
+└── .etc
+```
+
+{{% /choosable %}}
+
+To be clear, each of the applications/services inside our monorepo (and the `infrastructure` project is treated the same way as each independent application/service) are a separate Pulumi project, with their own stacks, and their own `Pulumi.yaml`. Given that each service is a separate Pulumi project, they can all use different programming languages. Let's take a look at how it might look if the `infrastructure` team prefers to write in Go, and the myApp team prefers TypeScript:
```
susanev, line 254: the parenthetical after monorepo, makes this a bit difficult to read. can we include it in a later sentence instead?

themes/default/content/docs/guides/organizing-projects-stacks/_index.md
```diff
@@ -177,3 +275,112 @@ These component resources can be packaged up and stored alongside all of your ot
 ### Other examples
 
 See also the use of multiple projects and stacks in [Crosswalk for Kubernetes]({{< relref "/docs/guides/crosswalk/kubernetes" >}}), which contains a tutorial, reference architecture, and collection of prod-first code examples that demonstrate industry best-practices for using Kubernetes in contexts where an organization of people must ship production applications.
+
+## Organizing your project code
+
+Within your Pulumi project, there are a good practices to consider to help keep your code organized, maintainable, and understandable.
```
susanev, line 281:
```suggestion
Within your Pulumi project, there are good practices to consider to help keep your code organized, maintainable, and understandable.
```

susanev (approved):

themes/default/content/docs/guides/organizing-projects-stacks/_index.md
```diff
@@ -177,3 +275,101 @@ These component resources can be packaged up and stored alongside all of your ot
...
+{{< chooser language "typescript,go" / >}}
+
+{{% choosable language typescript %}}
+Organize your code in a way that makes it easy to understand and maintain. One way to do this in Typescript is to break out your code into separate files, and then import them into your main file. In this example, the entrypoint for our Pulumi program is `index.ts`, but we use the `utils.ts` file for supporting functions.
+
+```typescript
+// index.ts
+import * as util from "./util";
+...
+const forwarderHandle = util.forwardPrometheusService(p8sService, p8sDeployment, {
```
cnunciato, line 292: Since it looks like the file is named `utils.ts`:

```suggestion
import * as utils from "./utils";
...
const forwarderHandle = utils.forwardPrometheusService(p8sService, p8sDeployment, {
```

themes/default/content/docs/guides/organizing-projects-stacks/_index.md
```diff
@@ -177,3 +275,101 @@ These component resources can be packaged up and stored alongside all of your ot
...
+...
+
+  return nil
+}
+
+```
+
+{{% /choosable %}}
+
+There are a couple of reasons that this pattern is helpful. One, in this particular case, is that the `forwardPrometheusService` function exists to forward the Prometheus service to localhost, so we can check it. If you are running in-cluster, we probably don't need it! So we could add a conditional to determine if we need to run that function - which makes our code a lot clearer.
```
cnunciato, line 373: To be fair, though, you could put that same conditional into `index.ts` or `main.go`. Since the point seems to be to clarify the benefit of breaking utility code (for example) out into separate modules or files within a package, it might be clearer to have the example be a utility function that does something super simple, like format a string in a way that's commonly used across a team or organization or something. What do you think? Otherwise, you sort of have to imagine (particularly if the body of the function is empty) what the `forwardPrometheusService` is doing and why it'd be a good candidate for extraction into a `utils` module.

themes/default/content/docs/guides/organizing-projects-stacks/_index.md
```diff
@@ -177,3 +275,101 @@ These component resources can be packaged up and stored alongside all of your ot
...
+    localPort,
+});
+```
+
+```typescript
+// utils.ts
+import * as k8s from "@pulumi/kubernetes";
+import * as pulumi from "@pulumi/pulumi";
+
+export function forwardPrometheusService(
```
cnunciato, line 302: Do you think this function is more complex that it needs to be to get the point across? I wonder if something super-simple wouldn't be better -- like a string formatter or something that's more "util"-y. There's a bunch of stuff here that, while awesome, probably isn't super relevant to what you're trying to convey. What do you think?

cnunciato (commented):
Bunch of suggestions for ya. Overall it's good, but I think it could probably benefit from a little more explanation as to what each of these parts is and what it's doing -- what `pkg` actually contains, maybe a little code showing how these parts come together in an example or two (as right now, that's mainly left to the imagination), etc. Something to think about maybe. But definitely helpful!

themes/default/content/docs/guides/organizing-projects-stacks/_index.md
```diff
@@ -134,20 +154,47 @@ These projects can be part of the [same monorepo as the infrastructure project](
 
 Our example service is made up of an API and a database (RDS, CosmosDB, etc.). Our Pulumi program for the project defines the resources for the API and the database, and it can also deploy the actual code, as well. When we add our example service, our monorepo starts to look like this:
 
+{{< chooser language "typescript,go" / >}}
+{{% choosable language typescript %}}
```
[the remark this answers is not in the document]
susanev: can we lint for this?

themes/default/content/docs/guides/organizing-projects-stacks/_index.md
mattstratton: Yep, that’s really a good idea. Otherwise this does start to also drift a bit into component resources as well which could confuse the point. And it’ll be much easier to illustrate cross-languages with something like this! Thanks for bumping me into the better direction.

susanev:
@mattstratton any way we could get this in soon, and then maybe come back later for more improvements? id love to get the things you've added out to folks soon.

mattstratton:
@susanev i need to write a function like @cnunciato suggested and change to use that - I want to ship this asap as well!

mattstratton:
@cnunciato I would like to incorporate your suggestions, but I think we should ship this as is for now and update more soon? there are other things in this PR that should get out. What do you think?

cnunciato (approved):

cnunciato:
@mattstratton Sounds good, roll forward! If we could come back to it, though, that would be awesome. 🚀

PR #1311, opened 2022-04-05
Add installation sections on Windows for winget and standalone

Zaid-Ajaj:
After [Pulumi #4676](https://github.com/pulumi/pulumi/issues/4676) was resolved, it is now possible to install Pulumi using `winget` on Windows but also using a standalone Windows x64 installer. This PR extends the installation section to include these options. 

Also makes the distinction for manual installation where "manual" means downloading binaries to distinguish from downloading the standalone installer

### DIFF
--- a/themes/default/content/docs/get-started/install/_index.md
+++ b/themes/default/content/docs/get-started/install/_index.md
@@ -111,7 +111,7 @@ Alternatively, you can install Pulumi manually. We provide a prebuilt binary for
 
 {{% choosable os windows %}}
 
-Windows 8 and 10 are supported.
+Windows 8 and later are supported.
 
 ### Chocolatey
 
@@ -129,6 +129,24 @@ Subsequent updates can be installed in the usual way:
 > choco upgrade pulumi
 ```
 
+### Winget 
+
+Install Pulumi using [winget-cli](https://github.com/microsoft/winget-cli/) package manager. This is built-in on Windows 11 and later. 
+
+```powershell
+> winget install pulumi
+```
+
+To update Pulumi to a more recent version:
+
+```powershell
+> winget upgrade pulumi
+```
+
+### Standalone Installer
+
+You can download the latest [Pulumi Installer for Windows x64](https://github.com/pulumi/pulumi-winget/releases/tag/v{{< latest-version >}}) and run it like any other installer. It will automatically add Pulumi to the path and make it available machine-wide.
+ 
 ### Installation Script
 
 1. Open a new command prompt window (**WIN+R**: `cmd.exe`):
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/get-started/install/_index.md
```diff
@@ -143,15 +161,15 @@ This will install the `pulumi.exe` CLI to `%USERPROFILE%\.pulumi\bin` and add it
...
 <!-- markdownlint-disable url -->
-1. Download [Pulumi {{< latest-version >}} for Windows x64](https://get.pulumi.com/releases/sdk/pulumi-v{{< latest-version >}}-windows-x64.zip). For prior versions and release notes, see the [Available Versions]({{< relref "/docs/get-started/install/versions" >}}) page.
+1. Download [Pulumi {{< latest-version >}} binaries for Windows x64](https://get.pulumi.com/releases/sdk/pulumi-v{{< latest-version >}}-windows-x64.zip). For prior versions and release notes, see the [Available Versions]({{< relref "/docs/get-started/install/versions" >}}) page.
 <!-- markdownlint-enable url -->
 
-1. Unzip the file and extract the contents to a folder such as `C:\pulumi`.
+2. Unzip the file and extract the contents to a folder such as `C:\pulumi`.
 
-1. Add `C:\pulumi\bin` to your path via **System Properties** -> **Advanced** -> **Environment Variables** -> **User Variables** -> **Path** -> **Edit**.
+3. Add `C:\pulumi\bin` to your path via **System Properties** -> **Advanced** -> **Environment Variables** -> **User Variables** -> **Path** -> **Edit**.
```
susanev, line 172: these numbers will automatically change to 1, 2, 3 form 1. 1. as part of rendering. and leaving it as 1, 1, etc can make it easier to add future items without adjusting all the following number.

susanev (approved):

themes/default/content/docs/get-started/install/_index.md
Zaid-Ajaj: I looked at the live version at https://www.pulumi.com/docs/get-started/install/ and there the numbers don't look like they are ordered 1, 2, 3. Instead they are numbered 1, 1, 2 🤔

susanev: ah yah the markdown lint comments messes with that functionality, ignore me then!

themes/default/content/docs/get-started/install/_index.md
```diff
@@ -129,6 +129,24 @@ Subsequent updates can be installed in the usual way:
 > choco upgrade pulumi
 ```
 
+### Winget
+
+Install Pulumi using [winget-cli](https://github.com/microsoft/winget-cli/) package manager. This is built-in on Windows 11 and later.
```
cnunciato, line 134: Super minor, but should this have an article before it? (Seems like either "using winget" or "using the winget package manager" would work, unless "winget package manager" is the actual proper name.)

```suggestion
Install Pulumi using the [winget-cli](https://github.com/microsoft/winget-cli/) package manager. This is built-in on Windows 11 and later.
```

cnunciato (approved):
One thing to consider. Thanks! Feel free to merge when you like.

Zaid-Ajaj:
Thanks @cnunciato for the suggestion, it makes sense. I applied the change and will merge when the CI goes green 😄

themes/default/content/docs/get-started/install/_index.md
```diff
@@ -129,6 +129,24 @@ Subsequent updates can be installed in the usual way:
 > choco upgrade pulumi
 ```
 
+### Winget
```
lukehoban, line 132: Curious - should this be listed first?  I'm not sure of the ecosystem usage of Winget vs. Chocolatey, but my sense is that Winget is more "official" these days.  Is there any reason from our perspective why we would expect folks to have a better experience with Chocolatey?

cnunciato: Relatedly, if that were the case, would it makes sense to update the Get Started tutorials to use Winget also?

Zaid-Ajaj: > Is there any reason from our perspective why we would expect folks to have a better experience with Chocolatey?

@lukehoban winget is indeed the "more official" package manager but my assumption is that Windows users are more familiar with chocolatey as it has been used and known for a much longer time than winget (which went [v1.0](https://devblogs.microsoft.com/commandline/windows-package-manager-1-0/) around 05/2021)

Another _catch_ of winget support is that pulumi releases are not immediately published to the winget registry: automated PRs such as [this one](https://github.com/microsoft/winget-pkgs/pull/56531) are sent to the official registry and have to be reviewed and merged by a maintainer so users might not always install latest pulumi because of this delay 

If ask me personally, I think the "Standalone Installer" could be the _default_ because it doesn't require any package manager (chocolatey, winget, etc.), it's just an MSI which means people can just uninstall it from control panel if they want and it is immediately created in [Releases of pulumi-winget](https://github.com/pulumi/pulumi-winget/releases) as soon as pulumi publishes a new version.

mikhailshilkov: > I think the "Standalone Installer" could be the default because it doesn't require any package manager

We don't have any auto-upgrade functionality, so in this sense it may be more beneficial to push users towards package managers than a manual installation. My 2c.

Zaid-Ajaj: > We don't have any auto-upgrade functionality

@mikhailshilkov That's a great point, these installers don't have that indeed.

PR #1328, opened 2022-04-07
Add a Dockerfile

cnunciato:
Still a WIP, working on local dev experience.


mattstratton:
ooh I would like this as it would help for those of us who don't want to have to update hugo if we don't have to :)

susanev:
this person was also requesting this in docs https://github.com/pulumi/docs/issues/7610

PR #1353, opened 2022-04-12
Language chooser spike

kimberleyamackenzie:


susanev:
- converters are different, but can we add that in? e.g., https://www.pulumi.com/arm2pulumi/
- how-to guides don't  have the update
- can we copy over API docs for at least aws?

PR #1362, opened 2022-04-13
Fix a bug (I think) in a Kubernetes blog-post example

cnunciato:
I believe (based on [this SO question](https://stackoverflow.com/questions/71802926/explanation-of-service-get-from-pulumi/71850966#71850966)) these should be `-master`, not `-redis-master`.


viveklak (approved):

viveklak:
Thank you!

PR #1387, opened 2022-04-18
add slack logo for pkg

susanev:


susanev:
@jkisk i think ur gonna need to update the package to use this method for the logo, but not totally sure; we want this way so we can remove the extra whitespace that slack adds to it by default

kimberleyamackenzie (approved):

stack72:
@susanev how can we control this for 3rd party packages? We can't always download a copy of the logo right? Is there something we need to do that's better here?

susanev:
@stack72 we can maybe find a tool that strips the whitespace from the logos, @kimberleyamackenzie do you know of anything that does that?

kimberleyamackenzie:
not that i know of, but i'll bet there's something out there if i poke around a little bit more

jkisk (approved):
Thank you!

PR #1391, opened 2022-04-19
Updated Lambda Function URL blog post

pierskarsenbarg:
Added examples in other languages, including a resource to run the AWS CLI and add the correct permission to allow the URL to be accessed (this is missing in the CC API and the TF provider at the time of writing hadn't updated the lambda permissions resource to handle this)


isaac-pulumi (commented):
@pierskarsenbarg is the Go example correct? It looks like the Command resource doesn't get created in that language example (but I know very little about Go, so, maybe this resource doesn't work in Go)

pierskarsenbarg:
I had an issue with Go and the command resource (https://github.com/pulumi/pulumi-command/issues/53) so updated it so it tells you to run the CLI command yourself

isaac-pulumi (approved):

isaac-pulumi:
@pierskarsenbarg that works for me

PR #1392, opened 2022-04-19
Put the console image inside the console note

cnunciato:
Since this image corresponds with the content that's inside the note, it's odd that it's both outside the note and shown before it. Putting it inside the note, and after the content it refers to, makes more sense. (Otherwise, it's unclear until you read the note why this image is even here.)

## Before

image

## After

image

### DIFF
--- a/themes/default/layouts/shortcodes/console-note.html
+++ b/themes/default/layouts/shortcodes/console-note.html
@@ -12,7 +12,7 @@
             and configuration.
         </p>
         <a href="/images/getting-started/console-update.png", target="_blank">
-            <img src="/images/getting-started/console-update.png">
+            <img src="/images/getting-started/console-update.png" alt="A stack update with console output, as shown in the Pulumi Service">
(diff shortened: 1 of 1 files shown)


sean1588 (approved):
LGTM!

susanev:
should this really be a note? i wonder if this should be worked into the regular content on the page.

sean1588:
I could definitely see this as being just part of the regular content on the page. I believe I was the one who originally added this and remember it being suggested to place this in a note, though I don't remember what the original motivation behind doing so was.

cnunciato:
> should this really be a note? i wonder if this should be worked into the regular content on the page.

That'd make sense to me too, yeah. I'm in these docs now working on some specific additions, and I'm finding lots of little opportunities for improvements like these, some that are quick and easy like this one, others that'd take a bit of adjustment to the prose. Compiling a list and planning on coming back to it as a follow-up.

PR #1399, opened 2022-04-20
remove links from get started landing pages

susanev:
a thing ive learned from @SaraDPH recently is that we should be thoughtful about linking people away from the content at the beginning. i was also talking with Luke about these pages today, and he mentioned that these links arent taking folks to a super useful place. so this is proposing removing them.

im open to hearing alternative thoughts and concerns!

### DIFF
--- a/themes/default/layouts/shortcodes/cloud-intro.html
+++ b/themes/default/layouts/shortcodes/cloud-intro.html
@@ -3,9 +3,8 @@
 <div>
     <p>
         Pulumi's infrastructure-as-code SDK helps you create, deploy, and manage {{ .Get 0 }} containers,
-        serverless functions, and infrastructure using <a href="{{ relref . "/docs/intro/languages" }}">familiar programming languages</a>. The Pulumi {{ .Get 0 }} provider packages
-        and <a href="{{ relref . "/docs/reference/cli" }}">CLI</a>
-        help you accomplish all these within minutes.
+        serverless functions, and infrastructure using TypeScript, Python, Go, or C#. The Pulumi {{ .Get 0 }} provider packages
+        and CLI help you accomplish all these within minutes.
     </p>
     <p>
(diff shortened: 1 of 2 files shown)


themes/default/layouts/shortcodes/cloud-intro.html
```diff
@@ -3,9 +3,8 @@
 <div>
     <p>
         Pulumi's infrastructure-as-code SDK helps you create, deploy, and manage {{ .Get 0 }} containers,
-        serverless functions, and infrastructure using <a href="{{ relref . "/docs/intro/languages" }}">familiar programming languages</a>. The Pulumi {{ .Get 0 }} provider packages
-        and <a href="{{ relref . "/docs/reference/cli" }}">CLI</a>
-        help you accomplish all these within minutes.
+        serverless functions, and infrastructure using TypeScript, Python, Go, or C#. The Pulumi {{ .Get 0 }} provider packages
```
cnunciato, line 6: Any reason to break with the existing language?

```suggestion
        serverless functions, and infrastructure using familiar programming languages. The Pulumi {{ .Get 0 }} provider packages
```

If we did want to change this, I might suggest a slight tweak, as to me, "using TypeScript, Python, Go, or C#" reads a bit like "these are the four languages we support". An easy way around that would be something like, "using programming languages like TypeScript, Python, C#, or Go." Up to you though.

cnunciato (approved):
Minor suggestion, otherwise 👍 from me. Thanks!

themes/default/layouts/shortcodes/cloud-intro.html
susanev: @cnunciato this is speaking to the actual language selectors that exist within these guides, but will add the word like if you think that's more clear.

susanev, line 6:
```suggestion
        serverless functions, and infrastructure using programming languages like TypeScript, Python, Go, or C#. The Pulumi {{ .Get 0 }} provider packages
```

PR #1417, opened 2022-04-25
Upgrade hugo to 0.97.3

zchase:


cnunciato (approved):
Let's give it a shot! Thanks @zchase.

PR #1425, opened 2022-04-27
Adjusting Python variables to be consistent

AndrewDavis1191:
Adjusting to snake case on pulumi.com [asset & archives page](https://www.pulumi.com/docs/intro/concepts/assets-archives/)


cnunciato (approved):
Great catch, thanks @AndrewDavis1191! 🎉

PR #1429, opened 2022-04-27
Org Tokens Page

djgrove:
Added an org tokens landing page for the docs site. This will be linked to from the feature itself in the application, and Can be further linked from other pages in the docs where it makes sense.

Caught that the "Delete Stack" REST endpoint is duplicated in docs and removed.

Added a bunch of links to the new access tokens page which should help users understand the distinctions between org/personal tokens.

## Issues
Fixes https://github.com/pulumi/pulumi-service/issues/8478 and


themes/default/content/docs/intro/pulumi-service/org-tokens/_index.md
```diff
@@ -0,0 +1,89 @@
+---
+title: "Organization Access Tokens"
```
susanev: id recommend aligning the title you go with, with the URL, if possible

themes/default/content/docs/intro/pulumi-service/org-tokens/_index.md
```diff
@@ -0,0 +1,89 @@
...
+meta_desc: Overview of the Organization Access Tokens feature and how to use it.
```
susanev: might be nice to try and explain what organization tokens are here, if possible

themes/default/content/docs/intro/pulumi-service/org-tokens/_index.md
```diff
@@ -0,0 +1,89 @@
...
+Organization Access Tokens are viewed by going to the organization’s Settings page, then selecting Access Tokens from the menu. As with Personal Access Tokens, this table displays all tokens belonging to your organization, and at what time they were last successfully used to carry out an operation. Only organization admins of an organization with an Enterprise Growth or Business Critical subscription will see access to this page.
+
+## Deleting an Organization Access Token
+
+An access token can be deleted by any organization admin at any time. Click the ellipsis icon at the end of the row, and select “Delete token”.
+<img src="./ellipsis-delete-token.png">
+
+You will be asked to confirm. If you choose to delete a token, its access will immediately be revoked and all further operations using it will fail as unauthorized.
+
+## Audit Logs
```
susanev: id recommend using a more specific heading, something like `Organization Token Audit Logs`
same for permissions, and API access

alternatively if you can make these h3s in a section then you could maybe drop that addition

susanev (approved):
i talked with Devon over zoom, but my recommendation is..
- a single access tokens page, that contains the information for both personal and org tokens; this will help inform folks about the types of tokens we offer and how they are different in a single location which should help reduce people creating and using these different types of tokens more successfully
- if folks feel very strongly that, this will be a poor user experience i ask that we create both org and personal tokens pages, and pull the current tokens content that's in accounts out into its own page
- i also recommend following the pattern through the rest of the service docs, with using an ordered list of steps for the actions you are describing instead of paragraph form, check out the access tokens content in account for an example
- we avoid using the word `click` throughout the docs as its pointer centric, instead use `select` (we have a style guide if you are interested in more things we try and align to: https://github.com/pulumi/pulumi-hugo/blob/master/STYLE-GUIDE.md)
- there's a bunch of extra white space in your screenshot, did you mean to do it that way?

i went through this very quickly, hope its helpful! let me know if you want me to take a closer look at anything

themes/default/content/docs/intro/pulumi-service/org-tokens/_index.md
susanev: im worried about someone searching for audit logs for example in our docs search and this being a result, and maybe not what they are looking for

themes/default/content/docs/intro/pulumi-service/access-tokens.md
```diff
@@ -0,0 +1,135 @@
...
+1. Select the ellipsis button.
+1. Choose **Delete token**. You will be prompted in a dialog to confirm your choice.
+
+## Organization Access Tokens
+
+Organization Access Tokens provide Enterprise Growth and Business Critical customers the opportunity to manage resources and stack operations for their organization independent of a single-user account.
+
+Leveraging Organization Access Tokens for your organization's automation delivers two core benefits:
+
+* Prevents organizations from needing to pass around a Personal Access Token belonging to a privileged organization member or admin in their automated tooling. This eliminates risk with modifying or removing organization memberships, as members can be decoupled from such automation pipelines.
```
EvanBoyle, line 42: Suggestion

Leveraging Organization Access Tokens for your organization's automation delivers many benefits over Personal Access Tokens:

* Organization Tokens belong to the organization, rather than an individual member. Any admin in the organization can view, create, and delete all Organization Tokens. If a member of your organization leaves, you don't have to worry about losing access to a core CI/CD token attached to their personal account. 
* Promotes less privileged access, as an Organization Access Token, unlike a Personal Access Token, is granted privileges only to the organization in which it was created, rather than to all organizations a single user belongs to.
* Audit logs and update history are attributed to the organization and the name of the token, rather than an individual user.

themes/default/content/docs/intro/pulumi-service/access-tokens.md
```diff
@@ -0,0 +1,135 @@
...
+
+| Action | Org Token Access |
+|--------|------|
+| Stacks |
+|--------|------|
+| List Stacks | ✅ |
+| Get Stack | ✅ |
+| Get Stack State | ✅ |
+| Transfer Stack |  |
+| Delete Stack |  |
```
EvanBoyle, line 97: Is it true that org tokens cannot delete a stack? 

If so that seems like a bug. What is the rationale?

EvanBoyle: Same with deleting a team.

EvanBoyle (approved):
Overall looks great. 

I was hoping that we could specifically highlight `Organization Access Tokens` in that navigation menu rather than `Access Tokens`. Can we think about tweaking that portion?

themes/default/content/docs/intro/pulumi-service/access-tokens.md
djgrove: I like the wording of these better. Included in [e3a2d04](https://github.com/pulumi/pulumi-hugo/pull/1429/commits/e3a2d04b6baacdb5515c7e330143b684314c0284)

themes/default/content/docs/intro/pulumi-service/access-tokens.md
djgrove: Updating to include that we support this, and will correct this in the main payload. Docs fixed in [22f33ac](https://github.com/pulumi/pulumi-hugo/pull/1429/commits/22f33ac8dc2afa81aedcfeeb294d1ffeae486f23)

djgrove: Not true, but has been corrected in docs [22f33ac](https://github.com/pulumi/pulumi-hugo/pull/1429/commits/22f33ac8dc2afa81aedcfeeb294d1ffeae486f23)

themes/default/content/docs/intro/pulumi-service/organization-access-tokens.md
```diff
@@ -0,0 +1,106 @@
...
+
+If you choose to delete a token, its access will immediately be revoked and all further operations using it will fail as unauthorized.
+
+## Auditing Organization Token Actions
+
+Since an organization can have multiple Organization Access Tokens, it’s necessary to be able to identify them uniquely in Audit Log Events. All Audit Log Events which were triggered by an Organization Access Token will surface the token’s unique name, and in the event of Audit Log Export, the token’s UUID as well.
+
+## Permissions/Authorization
+
+Put simply: Organization Access Tokens behave like an organization member with [stack write permissions](https://www.pulumi.com/docs/intro/pulumi-service/projects-and-stacks/#stack-permissions) to all of the organization’s stacks. They do not grant any privileges to view the console, or to create additional Organization or Personal Access Tokens. See below for a full list of accessible APIs:
```
susanev, line 54:
```suggestion
Organization Access Tokens behave like an organization member with [stack write permissions](https://www.pulumi.com/docs/intro/pulumi-service/projects-and-stacks/#stack-permissions) to all of the organization’s stacks. They do not grant any privileges to view the Pulumi Service ui, or to create additional Organization or Personal Access Tokens. See below for a full list of accessible APIs:
```

sry just noticed this in the last change, id suggest against using this language. it can feel intimidating for folks that may not find the concepts simple. in general, i recommend against describing things as easy or simple as we serve many folks with many different experience levels and things that may seem simple and easy to one person may be quite challenging to another.

we also moved away from using the term console, id go with either web console or Pulumi Service ui.

djgrove: 👍  in this case I'm going more for tl;dr. This clause doesn't provide much benefit anyway, and understand your point

EvanBoyle:
I love striving for excellence here but it is late and we need to get these docs published tonight (among other things) so let's hold further feedback as post-launch follow ups.

susanev:
I had no idea y'all were launching these docs last night, def would love to know that in the future so I can customize my feedback appropriately.

PR #1433, opened 2022-04-28
add oracle logo

susanev:


stack72 (changes requested):
I *think* this needs to be called oci to match the provider right?

susanev:
@stack, oh sry, will fix

PR #1440, opened 2022-04-29
Migrate ECS AWSX guide to multi-lang

danielrbradley:
Oustanding tasks:

- [ ] Add Go
- [ ] Remove props listings and just link to the registry (assuming this will exist)
- [x] Make the service-less tasks section clearer and probably shorter
- [ ] Docker ECR image integration rework
- [ ] _Maybe_ add a section on EC2?
- [ ] Remove or rework the fire-and-forget example ... this relies heavily on callback magic that's not multi-lang. Perhaps just add a note about it being classic-only?


stack72:
This was merged into another branch

PR #1475, opened 2022-05-04
Update modify-program.md

hmizael:
fixed the type of StorageAccountStaticWebSite object and, removed "./" from "./index.html" to avoid XML error in ARM API


mikhailshilkov:
Thank you @hmizael! I'm trying your changes and they LGTM but I also had to add this to the Blob resource to make it work:

```
  index-html:
    type: azure-native:storage:Blob
    properties:
      ...
      type: block
```

Was it needed for you too? If so, do you mind adding it to this PR? Thanks again.

hmizael:
Add.

susanev (approved):

PR #1480, opened 2022-05-05
Fix kubernetes template

iwahbe:
The Kubernetes template doesn't display. This fixes it.


themes/default/content/docs/intro/languages/yaml.md
```diff
@@ -42,7 +42,7 @@ The fastest way to get up and running is to choose from one of the following Get
             <img class="h-8 mx-auto" src="/logos/tech/gcp.svg" alt="Google Cloud">
         </a>
     </div>
-    <div class="flex-1 pb-4">
+    <div class="flex-1 pb-4 md:mr-4">
```
cnunciato, line 45: Since this is the last item in the horizontal list, we leave this extra spacing out, so the rightmost edges line up properly -- otherwise you end up with a little gap:
 
image


```suggestion
    <div class="flex-1 pb-4>
```

cnunciato (approved):
One little tweak, otherwise 👍 -- thanks!

PR #1484, opened 2022-05-06
fix learn nav item

susanev:


sean1588 (approved):
Oh wow, sorry about that. Thanks so much for fixing!! 🙂

Issue #1495, opened 2022-05-11
Documentation for the "List Teams" Action is Missing from the Pulumi Service REST API Documentation


phillipedwards:
### What happened?

In the [Pulumi Service REST API](https://www.pulumi.com/docs/reference/service-rest-api/) documentation, the "List Teams" action is missing from the documentation.

### Steps to reproduce

Navigate to https://www.pulumi.com/docs/reference/service-rest-api/#organizations look for the "List Teams" action.

### Expected Behavior

You should be able to see a "List Teams" action, much like the "List Users" action.

### Actual Behavior

The "List Teams" action is missing from the documentation. Specifically, for the `api.pulumi.com/api/orgs/{org}/teams`
![image](https://user-images.githubusercontent.com/25461821/167744824-7213d9c4-2a02-45cc-a6ba-d8bb409abf3e.png)

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #1497, opened 2022-05-11
Add 'List Teams' Action to Pulumi Service REST API docs

phillipedwards:
Fixes [#1495](https://github.com/pulumi/pulumi-hugo/issues/1495)


susanev (approved):
thank you!

any chance you could add it to the table in the org tokens page as well?
https://www.pulumi.com/docs/intro/pulumi-service/organization-access-tokens/

phillipedwards:
@susanev I'm not sure where the `default/components` directory came from. Good catch. Was that a left-over from something that was removed a longtime ago, possibly? My `pulumi-hugo` branch was way out of date, until I updated yesterday.

PR #1505, opened 2022-05-13
self-hosted: Document FunctionMode env vars

stevesloka:
ServiceFunctionMode determines the mode in which the service runs based
upon the env variable "PULUMI_API_CONTAINER_MODE". If this value is not defined,
then it defaults to "FULL". If the value does not match a known type, it defaults
to "FULL".

Usage:
  ServiceModeFull: Runs both the Pulumi API & Background Jobs
  ServiceModeAPI: Runs only the Pulumi API
  ServiceModeBackgroundJobs: Runs only background jobs

Signed-off-by: Steve Sloka <steve@pulumi.com>

### DIFF
--- a/themes/default/content/docs/guides/self-hosted/components/api.md
+++ b/themes/default/content/docs/guides/self-hosted/components/api.md
@@ -44,7 +44,7 @@ The API service is a Go-based application. This is a single binary application t
 
 This container runs an HTTP server which provides the APIs needed by the Console and the CLI. By default, this container will serve using port `8080` over standard HTTP. If [TLS](#tls-environment-variables) is configured the server will instead serve over port `8443` using HTTPS.
 
-#### Function Mode 
+#### Function Mode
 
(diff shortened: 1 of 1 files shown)


susanev:
@stevesloka in the future if ya dont use a fork, we can run the tests on github

themes/default/content/docs/guides/self-hosted/components/api.md
```diff
@@ -44,6 +44,16 @@ The API service is a Go-based application. This is a single binary application t
 
 This container runs an HTTP server which provides the APIs needed by the Console and the CLI. By default, this container will serve using port `8080` over standard HTTP. If [TLS](#tls-environment-variables) is configured the server will instead serve over port `8443` using HTTPS.
 
+#### Function Mode
```
susanev, line 47:
```suggestion
#### Function Mode
```

this was failing on `make lint`

susanev (approved):

themes/default/content/docs/guides/self-hosted/components/api.md
stevesloka: 👍

PR #1517, opened 2022-05-17
Prettier Spike

kimberleyamackenzie:
As a follow-up to a recent frontend code reviews retro, I spiked on adding Prettier to pulumi-hugo.  Let's use this PR to discuss whether we want to add this, and if so, some of the details about config.  I will plan to talk about this more in depth at the frontend working group meeting on 5/20, but am opening this PR now so folks can give it a look.

### Reviewing
- There are some open questions about what choices we want to make here.  Please consider sharing your thoughts on the open questions below.
- I had to manually make a handful of changes, mostly for mismatched closing tags or invalid html syntax.  I put these manual changes in a separate commit, to make them easier to review - please flag if you think any of the manual changes are incorrect or will cause another issue.
- You can also look at the config commit to see those changes in isolation for easier review.
- Eventually, before we merge this, we should smoke test the preview to make sure nothing went sideways - I don't think we need to worry about that yet, until some discussion is had.

### Open questions:
- I used the same config we used with theme, with a couple Hugo-specific additions.  Are we happy with this base config?
- Right now, I have Prettier ignoring markdown files.  Do we want Prettier to ignore or include md files? Other files we should add or ignore?
- Our current `make lint` runs some markdown specific linting.  I didn't think it made sense to bundle the Prettier formatting in with that step, since Prettier is writing changes, so I created `make format`.  Are we happy with that?
- Should we add this as a step on the build?


susanev:
thanks for doing the manual changes in a separate commit!!

>I used the same config we used with theme, with a couple Hugo-specific additions. Are we happy with this base config?
I am

>Right now, I have Prettier ignoring markdown files. Do we want Prettier to ignore or include md files? Other files we should add or ignore?
I am good with starting here

>Our current make lint runs some markdown specific linting. I didn't think it made sense to bundle the Prettier formatting in with that step, since Prettier is writing changes, so I created make format. Are we happy with that?
make format sounds good to me

>Should we add this as a step on the build?
i think we should

cnunciato:
Thanks for the separate commit! ⭐ 🍦 

> Do we want Prettier to ignore or include md files?

I'd probably suggest having Prettier ignore them yeah, since we're already linting them with `markdownlint`. We could come back, though, and apply Prettier as a replacement to `markdownlint` (here and elsewhere) to reduce the number of tools we're using for this stuff.

> I didn't think it made sense to bundle the Prettier formatting in with that step, since Prettier is writing changes, so I created make format. Are we happy with that?

It looks like `make format` only runs `--check`, though, which I assume means "tell me what's wrong and I'll correct it myself". If this is the behavior we want, then I think adding this check to `make lint` would make more sense, since `make lint` generally just tells you what's wrong and leaves it up to you to fix it. (Whereas `make format` to me sounds more verb-like, as in "Do the formatting."). Then maybe we could use `make format` to run `--write` to do the actual formatting (and presumably leave the working tree dirty). What do you think? 

> Should we add this as a step on the build?

If we're thinking of adding this check to `make lint` (i.e., `prettier --check`), that's definitely make sense to me! 

> with a couple Hugo-specific additions

Which ones were those?

README.md
```diff
@@ -41,14 +41,18 @@ make serve
...
-## Linting and testing
+## Linting and formatting
 
 To check your code and your Markdown files for issues before submitting, run:
 
 ```
 make lint
 ```
 
+This repo uses Prettier for code formatting, to keep styling + formatting issues aligned without the need to be reviewed in PRs.  We currently do not automatically run the formatter as a build check, but may opt to do so in the future as needed.  To see what files have formatting diffs, you can run `make format` from the project root, which will list all the files with diffs.  To fix these diffs in bulk, you can run `yarn prettier --write .` also from the root.
```
jkodroff, line 52: FWIW, I would run lint in the build. And _maybe_ have targets or yarn scripts that both _check_ the formatting (returning non-zero if changes are proposed) and _does_ the formatting, but that might be overkill.

.github/workflows/pull-request-closed.yml
```diff
@@ -1,29 +1,29 @@
 name: Close pull request
 on:
-  pull_request:
-    types:
-      - closed
+    pull_request:
```
jkodroff, line 3: Not sure if the YAML spacing changes are intentional, but I think 2 spaces is more common. Suggest adding a `.editorconfig` or similar to help keep this consistent across team members.

jkodroff (changes requested):
Overall, this is a clear improvement (Just look at those closing tags and semicolons! So lovely!) and would be fine to merge as-is, with the exception of the YAML reformatting.

Optionally, I think JS, TS, HTML, and CSS all read the same or better at 2 spaces - your call.

themes/default/layouts/learn/glossary/list.html
```diff
@@ -27,7 +26,7 @@ <h1>{{ .Title }}</h1>
             {{ range $terms }}
                 {{ $anchor := .term | anchorize }}
                 <div class="flex-1 mt-8">
-                    <h2 id="{{ $anchor }}" class="mt-2 mb-0">{{ .term }} </h2>
```
zchase, line 30: Having prettier lint remove this extra space might introduce a subtle bug where that space is intentional. Depending on when we run prettier this could be confusing to someone without context on how prettier works. Though would like to hear what other folks think.

themes/default/layouts/migrate/cloudformation.html
```diff
@@ -39,56 +42,49 @@ <h2>Increase productivity with programming languages</h2>
...
-                    other examples here.</a>
+                    <a class="link" href="https://github.com/pulumi/examples" target="_blank">Find many other examples here.</a>
                 </p>
             </div>
             <div class="md:w-1/2 md:ml:4">
                 {{ $code := `const aws = require("@pulumi/aws");
 
-let size = "t2.micro";
-let ami  = "ami-7172b611"
+                    let size = "t2.micro";
```
zchase, line 58: This causes weird formatting issues. Not sure if we can ignore it or will need to move any inline code blocks to the front matter.

zchase (commented):
Nice! Thanks for doing this!

I left a few comments about a few things I saw during my quick look at the output. I think we are going to want to think about a few of the rules and whether we want to adopt them.

Is there a way we can be really strict about what rules we are adopting? I think if we start with a small set of rules and incrementally adopt new ones it will be smoother than trying to do it all at once.

themes/default/layouts/learn/glossary/list.html
susanev: can you give an example of where we are using an extra space here? or a situation you imagine we would want that in the future?

djgrove:
Call out from FE working group Zoom meeting: we should consider making the format commit excluded from git blame

djgrove:
I think if we do this: https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame

and then do a merge without squashing this should work. A squash merge would change the commit hash so it would still show up in git blame

themes/default/layouts/migrate/cloudformation.html
kimberleyamackenzie: We can def do one-off ignores: https://prettier.io/docs/en/ignore.html.  If you want to flag here areas where you think we should add them, I can do it in this PR, or we can re-address in a follow-up.

kimberleyamackenzie:
Thanks y'all for the input!  Summarizing some actions/responses to feedback here:

#### Spacing
Generally, in other repos where we've used Prettier, we have standardized on 4 spaces (versus 2).  I updated our config in this repo to make it 2 (which is Prettier's default, actually) for yaml files to follow best practices there.

#### Rules we're using
So, Prettier's philosophy is to be very opinionated, and it has its own defaults for everything - however, in past discussions of adding Prettier to repos at Pulumi, we've had conversations where folks have wanted an explicit set of rules that are different than those defaults - and that's where the set I have in this config comes from.  I am happy to go either way, but whichever route we choose (dropping some of these rules to let Prettier do its own thing, or keeping this set of config), we should try and be consistent across repos where it makes sense. 

### Editor config
I added an .editorconfig to make it easier for folks to follow our formatting standards.  Please give this a look, and make any suggestions here you'd like.

#### Commands
I updated `make format` to run `--write`, and added a `--check` to the existing lint step we have, so we're covered both places there.  Since we lint on PR builds, builds will now fail if they're not properly formatted.

#### Next steps
- Please re-review with the changes I've incorporated
- If you can give the preview link a browse, specifically in places where I have made manual changes, that would be great!
- Before merging I will use the link @djgrove shared to ensure that the major formatting commit is ignored from git blame

.editorconfig
```diff
@@ -0,0 +1,12 @@
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 4
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.yml, *.yaml]
```
jkodroff, line 10:
```suggestion
[*.yml, *.yaml, *.json]
```

.prettierrc.json
```diff
@@ -0,0 +1,26 @@
...
+    "useTabs": false,
+    "overrides": [
+        {
+            "files": ["*.html"],
+            "options": {
+                "parser": "go-template"
+            }
+        },
+        {
+            "files": ["*.yml", "*.yaml"],
```
jkodroff, line 20: It looks like you can reference `.editorconfig` so that you do not have repeat these overrides: https://prettier.io/docs/en/configuration.html#editorconfig

If you're already explored this option, please ignore my comment.

Makefile
```diff
@@ -13,6 +13,10 @@ ensure:
 lint:
 	./scripts/lint.sh
 
+.PHONY: format
+format:
```
jkodroff, line 17: If you want an even shorter name, `fmt` has become a familiar abbreviation to many developers because of the `go fmt` command. For example, TF uses `terraform fmt`.

susanev: id prefer to not abbreviate format, since we have lots of folks who will be contributing here, that are not developers

djgrove (approved):

susanev (approved):

PR #1529, opened 2022-05-19
[DO NOT MERGE] mobile nav updates

sean1588:
Issue: https://github.com/pulumi/pulumi-hugo/issues/427

Updates to mobile nav design dscribed in above issue.

For Reviewers:

I recommend testing the preview link out in both the mobile view using the chrome dev tools as well as on a touch device (e.g. iphone) as the mechanics seem to be a bit different between the two. I'll elaborate on some of these differences as well as some compromises I made. I would definitely like some feedback here concerning this. Also, check to verify this makes no changes to the desktop version of the site as everything should remain as is.

When I was developing this I mostly used the mobile view in the chrome dev tools. I then tried this out on my iphone and noticed a couple issues that may affect the mechanics of this. @jonelleb would love your input on the items below and would be happy to chat with you and/or other folks about it over zoom if needed for clarification since it is a bit complicated to describe but will do my best :)

- **Pinning the footer that contains the Getting Started button**: since we are pinning the footer with the “Get Started” button to the bottom, it seems the iphone is covering up much of this area at the bottom of the screen because of an overlay that get layed over the bottom portion of the screen until you start scrolling. This seems to remain until you start scrolling up which un-pins the top banner and then causes the overlay to disappear. If we were to pin the banner at the top then the screen would not be scrollable until one of the links is opened and fills up the screen so the overlay would remain covering the footer. So my question is, is this behavior that I have ok, or should we make further adjustments? 

Image from iOS (1)resized

- **Pining the currently selected menu item to the top of the menu**: the best I could get this to work was with the product menu. I could not get this to work reasonably with the learn menu, since the menu is unable to scroll the learn menu to the top of the nav when being selected becaause there is not enough content on the pane to further scroll. Let me know if this is ok.
Peek 2022-05-19-09-28

- **iphone screen shape and bar at bottom of screen**: this is hard to tell from the screenshot since it is rectangular, but on the iphone the screen itself is actually rounded at the corners at the bottom which tends to cutoff the getting started button a bit, also, the bar that ios overlays tends to interfere with the click behavior. Should we consider either moving the footer up a bit or just completely unpinning it? 
Image from iOS (2)resized

Btw, I chatted with @jonelleb about the behavior when it comes to folding up menus that are currently open when another one is clicked. We decided the currently open menus should remain open and not fold up when another menu is selected. @jonelleb what do you think about the font sizes, spacing, etc. Let me know if you think further adjustments need to be made.

Also, having some issue getting the carets to rotate... for whatever reason they won't transform.... anyone have any ideas?


jonelleb (changes requested):
Everything looks really good! I do have a few changes though. My comments are below.

- **As far as the footer with the Getting Started button, I see what you mean about the iPhone covering them up and they're not seen until the user scrolls. I really would like to keep the footer & button pinned so if there's nothing we can do, I say keep it as it is.** (This is a pretty common behavior for things that are pinned to the bottom of the screen on an iPhone.)
- **I think that the way you have the scroll working and the pinning of the menus to the top is fine.** I know you mentioned not being able to do it for the learn menu but when testing it out for myself, I feel that how you have it now will work.
- **I like that the footer is pinned, and I think that makes sense so the user has those footer links and the Get Started button available to them without having to scroll all the way down to the bottom of a long mobile page. I would prefer that we just move the footer up a bit to try to help the iPhone screen shape situation.** (I have no specifics, but I trust your judgement on how much to move it up.)
- **Oh, and if would be nice if the carets rotated but if we can't do that, that's a very small thing and no problem if you can't figure it out. I would like the carots to be clickable along with the link text though.** I clicked on the caret expecting for a menu to open and nothing happened.

Thanks!

sean1588:
thanks @jonelleb! Ok sounds good! I'll mess around with the footer a bit more to see what we can do to make it play better a little with the iPhone. I'll try moving it up a bit like you suggested and see if that helps. Thanks again for your insight!

kimberleyamackenzie:
@sean1588 The reason the carets aren't flipping is because you can't transform on inline elements, so if you make the span display block + then fix up the alignment, it should work.

Otherwise the good LGTM - i can test on my iphone for another set of eyes once you push your changes, just let me know.  Thanks for diving so deep on the mobile experience!!

kimberleyamackenzie (approved):
Code lgtm - haven't manually tested pending the changes @jonelleb requested!

sean1588:
Thanks @kimberleyamackenzie!!  I'll give that a try to see if it gets the carets to rotate. :slightly_smiling_face:

sean1588:
Sorry for the wait... I went and moved up the footer a bit to make it work with the iphone. I had to move it up a bit more than I expected in order for it to become clickable so let me know what you all think. I was able to get the carets to rotate now thanks to @kimberleyamackenzie for the help! The changes for that are in the theme [PR](https://github.com/pulumi/theme/pull/141).  The footer starts out pinned directly at the bottom until you scroll up and remove the banner. I did this since on the iphone you wouldnt be able to see it any way until you started scrolling and would look normal to start for other devices. See gif below. I had to move it up quite a bit, since the iphone seems to make a significant part of the bottom portion of the screen not clickable (at least this is the behavior on my iphone), I'm guessing to not interfere with being able to swipe up to close the app.... Anyways... this is what I came up with. Let me know what you all think...
Peek 2022-05-26 10-29

jonelleb (changes requested):
Looks great! And thanks for fixing the rotating caret. I have 2 small changes. Would it be possible to have the caret turn that light shade of purple when it rotates? This is currently happening on the desktop version of the nav so I'd like it to be consistent. Also, can the caret when clicked open the menus as well as the text? Right now when I click on the caret nothing happens and just in case users click on that rather than the text I'd like the menu to appear. Thanks!

susanev:
@sean1588 you up for picking this back up next iteration?

sean1588:
yeah, I can pick it back up.

sean1588:
hey everyone, I think this is ready to be reviewed now. I updated the caret behavior to reflect what @jonelleb suggested in the above comment. Please test out the preview link on the iPhone or some other mobile device, since the mobile view using chrome dev tools doesn't accurately reflect the behavior on a real touch device.

Also, some things have changed since I first picked this up a few months ago. The top navigation menu no longer has "Cloud engineering" as an option, it was changed to "Why Pulumi?" and also moved to the right of the "Product" menu. Some new items were added to some of the menus as well. I updated the mobile nav to reflect the updated menu structure that is on the desktop version of the site and also updated the links. Let me know what you all think.

One last thing, there was a "Templates" link added under the "Learn" menu. The icon used there is not one in our font-awesome icon set. It looks like it is just an svg that was added. The color of it though is different than the color in figma design for the mobile menus (#484848 (gray.700) which is a color we have as a brand color). The color on the svg is the same as the color we are using for icons on the desktop menu (#4A5568). Since it is an svg file, I can't update the color just using css, so right now the color is slightly different than the other icons on the mobile menu.

So should I ...
- update the color of the desktop menu to be gray.700 like we have on the mobile menu design and then get the svg color updated?
- update the mobile nav color to be the same as the desktop menu (#4A5568)?
- or leave the colors as they are and have the mobile menu icons be a slightly different color and create another svg for the templates icon with the mobile menu icon color?

sean1588:
I think this is ready now. If you all could give it once last glance. Also, test out the preview on the iphone or other touch device since it behaves slightly differently on real touch devices than in the chrome emulator. 

@jonelleb I went and updated the svg icon as well as updated the icons on the desktop version of the nav menu to be gray.700. Let me know if it looks good to you.

Also, @kimberleyamackenzie can you give the code one last glance. I think you may have approved a long time ago, but I made a couple changes since then. Also if you could look at the theme [PR](https://github.com/pulumi/theme/pull/141/files) as well that would be great!

cc @susanev

jonelleb (approved):
Checked on my iphone and it looks great! Thanks!

shughes26 (approved):

susanev:
@sean1588 i havent had time to look yet, but no reason to block on me if other folks have approved! :shipit:

PR #1535, opened 2022-05-20
Add newsletter cta to bottom of blog posts

zchase:
This PR adds a CTA to the bottom of the blog posts that asks people to subscribe to the newsletter.

Fixes: https://github.com/pulumi/pulumi-hugo/pull/1535

subscribe-blog

### DIFF
--- a/themes/default/layouts/blog/single.html
+++ b/themes/default/layouts/blog/single.html
@@ -25,6 +25,18 @@
                         <section data-swiftype-index="true">
                             {{ .Content }}
                         </section>
+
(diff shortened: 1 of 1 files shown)


lindsaymarolich (approved):

cnunciato (approved):
Totally beat me to it! 🎉

susanev (commented):
tots fine if you come back later; but the heading is not centered above the inputs
and that heading is getting a link icon, which i don't think we want
id honestly recommend not using a heading above this

newsletter

Issue #1536, opened 2022-05-20
"Get started" on https://www.pulumi.com/kubernetes/ links to AWS getting started


squaremo:
### What happened?

I was looking for Kubernetes-specific reference docs, and ended up at https://www.pulumi.com/kubernetes/. Out of curiosity I clicked the prominent "Get started" button, and it took me to an AWS-oriented tutorial. I would expect to be taken to a Kubernetes-oriented tutorial.

### Steps to reproduce

- go to https://www.pulumi.com/kubernetes/
- click on "Get started"

### Expected Behavior

Shown a Kubernetes tutorial.

### Actual Behavior

Shown an AWS tutorial.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

susanev:
oh no! thats a mistake, ill fix that up!

PR #1538, opened 2022-05-20
fix get started buttons

susanev:
fixes https://github.com/pulumi/pulumi-hugo/issues/1536


squaremo (approved):
The mis-linked button from #1536 is fixed in the preview :+1: I can't comment authoritatively on the other button changes (they look reasonable!)

PR #1540, opened 2022-05-20
Add YAML workshop for june 2022

zchase:
Part of: https://github.com/pulumi/marketing/issues/334


lindsaymarolich:
@zchase this looks great, thank you for adding the CTA to the blog as well! Only a few quick requests. Could we put the CTA on the blog post higher (right above the video for YAML) with a note, similar to what we did on this post:
https://www.pulumi.com/blog/announcing-aws-native/

Could you please add this as the meta image for the YAML workshop?
IntroIaC_6-16

lindsaymarolich (approved):

Issue #1534, opened 2022-05-20
something happened to the learn heading


susanev:
- [ ] h1 its much smaller than the rest of the h1s on the website
- [ ] also, why someones here if we can add a line-length or similar to those paragraphs they are much too long and folks will lose their place trying to read them. accessibility best practices recommend 45-75 characters per line
- [ ] and there are 2 h1s on this page, we need exactly 1 h1 per page, that second heading should be an h2

![learn-heading](https://user-images.githubusercontent.com/5489125/169432540-4560a177-e395-4745-9074-cf331be71008.png)

susanev:
@sean1588 any chance you might be able to take this sometime soon? @zchase do you think this happened when the marketing css was pulled out?

sean1588:
yeah I should be able to take it on tomorrow. :+1:

cnunciato:
It looks like this change is the reason for the smaller header text: https://github.com/pulumi/theme/pull/97/files#diff-dfcd873b65c820df18f3f95d14b705ac22f67246c26170d31738f68291550392L115-L116

susanev:
thanks @sean1588!!

mind shortening those paragraphs, so they look a bit better? maybe...

```
Learn all the concepts of Pulumi, along with best practices and architectural patterns.

Each tutorial will step you through exploring Pulumi and completing a project. Start with Pulumi Fundamentals to get comfortable with vocabulary, then try another pathway to explore more!
```

PR #1542, opened 2022-05-20
Learn page fixes

sean1588:
Fixes: https://github.com/pulumi/pulumi-hugo/issues/1534
Theme PR: https://github.com/pulumi/theme/pull/144

This PR addresses the following fixes that are also outlined in the issue linked above. 

- [x] h1 its much smaller than the rest of the h1s on the website - _this is addressed in the above theme PR. more info there._
- [x] add a line-length or similar to those paragraphs they are much too long and folks will lose their place trying to read them. accessibility best practices recommend 45-75 characters per line 
- [x] there are 2 h1s on this page, we need exactly 1 h1 per page, that second heading should be an h2

Here is a screenshot. Also, check out the preview when it becomes available. Let me know what you all think and if we need to make any other adjustments to other things on the page. The page is located at the `/learn` route.

Screenshot from 2022-05-20 16-02-54

### DIFF
--- a/themes/default/content/learn/_index.md
+++ b/themes/default/content/learn/_index.md
@@ -22,17 +22,19 @@ tutorials:
         Tutorials are series of hands-on activities that help you along your
         learning journey with Pulumi.
 
+# There are 2 trailing spaces appended to each line in the paragraphs below
+# in order to force a new line to limit the line length to 75 characters.
 overview: |
-    <h1>Welcome to Learn Pulumi!</h1>
+    <h2>Welcome to Learn Pulumi!</h2>
 
-    These learning pathways will help you learn all of the concepts of Pulumi,
-    along with best practices and architectural patterns, through real-life
-    examples.
+    These learning pathways will help you learn all of the concepts of Pulumi,  
+    along with best practices and architectural patterns, through real-life  
+    examples.  
 
-    In each tutorial, you'll complete a series of steps while exploring Pulumi
-    to complete a project or build a system. Start with Pulumi Fundamentals to
-    get comfortable with the vocabulary, and then head to another pathway to
-    explore further.
+    In each tutorial, you'll complete a series of steps while exploring Pulumi  
+    to complete a project or build a system. Start with Pulumi Fundamentals to  
+    get comfortable with the vocabulary, and then head to another pathway to  
+    explore further.  
 ---
 
 Ipsum lorem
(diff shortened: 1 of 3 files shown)


themes/default/content/learn/_index.md
```diff
@@ -22,17 +22,19 @@ tutorials:
         Tutorials are series of hands-on activities that help you along your
         learning journey with Pulumi.
 
+# There are 2 trailing spaces appended to each line in the paragraphs below
+# in order to force a new line to limit the line length to 75 characters.
```
sean1588, line 26: I modified this to be 75 characters/line. I would have liked to do this with CSS, but the content for those paragraphs is being piped into markdownify when we render and  whatever "magic" it is doing seems to have styling that is overriding what I was trying to do and kept messing with the text alignment. So I ended up just adding new lines in the actual markdown.

sean1588:
Thanks @kimberleyamackenzie. That worked well for the most part! I ended up needing to make some adjustments to flex with some of the elements in order to get the paragraphs items to align correctly which worked fine but when I collapsed the screen to test for mobile view, looks like some of the text was getting cut off and wouldn't wrap correctly... I messed around with it for a while but couldn't get exactly the results I wanted so I ended up just reverting to making the new lines in the mark down. Thanks anyways for the help! :slightly_smiling_face: 

I went ahead and updated the learn site to use the wording @susanev suggested. Screenshot below... 

I notice there is a bit of empty space between the overview text and the Learn Pulumi title banner. This space is also there on the currently deployed version of the site. When I inspect, it looks like there is an empty div in between the banner and the overview text that is causing it. It looks like this may have been added intentionally so I didn't remove it. I can remove it fairly easily if we feel it would be more aesthetically pleasing. Let me know.

Screenshot from 2022-05-23 11-11-45

susanev (approved):
@sean1588 i think that space is that way on every page, so lets hold off changing it for now. lolz at it being an empty div though.

sean1588:
I updated the h2 tag to be consistent with the sizing on the other pages. I made this change in the theme [PR](https://github.com/pulumi/theme/pull/144). More info on why I did it that way there. Either way I think we should do some investigation as to why certain pages like the learn page are not inheritting the same styling as other pages on the site and figure out how we can gain better consistency across all the pages when it comes to the styles being applied. I'll open an issue to look into this.

PR #1546, opened 2022-05-23
fix kube+python get started

susanev:
this came up in two usability studies
as being misaligned from the template: https://github.com/pulumi/templates/blob/master/kubernetes-python/__main__.py

[internal slack convo](https://pulumi.slack.com/archives/C9SGX9QA1/p1653333169391999)


lukehoban:
This catches the docs up with the template changes made in https://github.com/pulumi/templates/pull/133.

kimberleyamackenzie (approved):

Issue #969, opened 2022-01-13
use fully qualified refs


susanev:
we use some refs that are really difficult to track down when changing links, we should fix this

some convo on this in a past pr: https://github.com/pulumi/pulumi-hugo/pull/968#discussion_r784253243

susanev:
refs no longer exist

PR #1549, opened 2022-05-24
Fix relative link to auto-naming in intro

squaremo:
I looked for similar mislinks to this anchor in the vicinity, on the basis that it was caused by a file move or other rearrangement. I think this is the only one though.

### DIFF
--- a/themes/default/content/docs/intro/concepts/resources/_index.md
+++ b/themes/default/content/docs/intro/concepts/resources/_index.md
@@ -114,7 +114,7 @@ resources:
 
 {{< /chooser >}}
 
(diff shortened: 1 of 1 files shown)


squaremo:
See link with text "auto-names" in http://pulumi-hugo-origin-pr-1549-c20c8dde.s3-website.us-west-2.amazonaws.com/docs/intro/concepts/resources/ vs the same link in https://www.pulumi.com/docs/intro/concepts/resources/

susanev (approved):
@squaremo thx for the fix! would you m find updating the ones you are changing to be fully qualified though? this will enable us to get useful error messages in the future.

ref: https://github.com/pulumi/pulumi-hugo/issues/969

squaremo:
Owph, sorry for all the `git push --force-with-lease`, Pulumi Bot.
OK this is ready to review again.

PR #1557, opened 2022-05-26
Ensure camelCase warning is above nested property excamples for IgnoreChanges

guineveresaenger:
I was trying to use https://www.pulumi.com/docs/intro/concepts/resources/options/ignorechanges/ during a repro using Python for the first time and did not see the very important Note because it was placed on the bottom of the doc.

This PR makes the note appear only for the relevant languages (Go, C#, and Python) where casing for nested resources differs from the Pulumi schema.

### DIFF
--- a/themes/default/content/docs/intro/concepts/resources/options/ignorechanges.md
+++ b/themes/default/content/docs/intro/concepts/resources/options/ignorechanges.md
@@ -91,6 +91,37 @@ One reason you would use the `ignoreChanges` option is to ignore changes in prop
 
 In addition to passing simple property names, nested properties can also be supplied to ignore changes to a more targeted nested part of the resource's inputs. Here are examples of legal paths that can be passed to specify nested properties of objects and arrays, as well as to escape object keys that contain special characters:
 
+{{< chooser language "javascript,typescript,python,go,csharp,java,yaml" >}}
+
+{{% choosable language python %}}
+
+{{% notes "info" %}}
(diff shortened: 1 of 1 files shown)


susanev (approved):
thank you for fixing this! i think you could do a shortcode for this instead so ur not repeating the same content over and over again, but no need to block on that if you'd like to get this in quickly.

themes/default/content/docs/intro/concepts/resources/options/ignorechanges.md
```diff
@@ -91,6 +91,37 @@ One reason you would use the `ignoreChanges` option is to ignore changes in prop
 
 In addition to passing simple property names, nested properties can also be supplied to ignore changes to a more targeted nested part of the resource's inputs. Here are examples of legal paths that can be passed to specify nested properties of objects and arrays, as well as to escape object keys that contain special characters:
 
+{{< chooser language "javascript,typescript,python,go,csharp,java,yaml" >}}
+
+{{% choosable language python %}}
+
+{{% notes "info" %}}
+The property names passed to `ignoreChanges` should always be the "camelCase" version of the property name, as used in the core Pulumi resource model.
```
susanev, line 99: why is camel case in quotes here?

guineveresaenger: It's how I found it! I think it makes sense since camelCase isn't a real English word, or at least a very niche one. Happy to remove

PR #1567, opened 2022-06-02
Copy helper unit test method for C# in tutorial

### LINKED ISSUE #1566: Apply is mentioned but not explicit in the Tutorial code samples for C#
[themes/default/content/docs/guides/testing/unit.md](https://www.pulumi.com/docs/guides/testing/unit/)

Hi Team I found a section in this tutorial that needs to be clarified,

The tutorial mentions the need to use the `apply` method but in the code samples on the page, its use is not evident. The `apply` method is encapsulated inside the `GetValueAsync` method and this is not explained either. Upon navigating to the GitHub repo, then it becomes evident where the `apply` method is used. Can you please fix this? Here is the text in question:

_Because all Pulumi resource properties are [outputs](https://www.pulumi.com/docs/intro/concepts/inputs-outputs/)—since many of them are computed asynchronously—we need to use the **apply** method to get access to the values._

Thanks for your help,

davidcconrad:
Copied the C# helper method from the c# unit test directory: https://github.com/pulumi/examples/blob/74db62a03d013c2854d2cf933c074ea0a3bbf69d/testing-unit-cs/Testing.cs

This should fix #1566

### DIFF
--- a/themes/default/content/docs/guides/testing/unit.md
+++ b/themes/default/content/docs/guides/testing/unit.md
@@ -501,6 +501,17 @@ public async Task InstanceHasNameTag()
     tags.Should().NotBeNull("Tags are not defined");
     tags.Should().ContainKey("Name");
 }
+
+public static Task<T> GetValueAsync<T>(this Output<T> output)
+{
+    var tcs = new TaskCompletionSource<T>();
+    output.Apply(v =>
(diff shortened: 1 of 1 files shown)


jkodroff (approved):
LGTM!

PR #1583, opened 2022-06-03
Add Kubecrash demo blog post

guineveresaenger:
Resolves https://github.com/pulumi/home/issues/2054.


SaraDPH (approved):
My only recommendation is to add more to the meta_Dec, like: **_Learn how to customize the multi-cloud developer experience with Kubernetes and Pulumi, using TypeScript._**

Active voice performs on organic searches and sets the expectations.

PR #1588, opened 2022-06-06
fix(link): patch link real quick

nimbinatus:
noticed this when digging something else out


susanev (approved):
i legit cannot tell what ur fixing thx to this all being on one line and suggested changes not helping, but i trust ya!

nimbinatus:
😅  Not sure why it didn't highlight. There's an extra bracket in the Registry link:
Screen Shot 2022-06-06 at 5 59 59 PM
I removed it so the link renders. Last line of the first paragraph: http://pulumi-hugo-origin-pr-1588-64f8977f.s3-website.us-west-2.amazonaws.com/blog/pulumiup-native-providers/

Leaving this here for an explanation; thanks for the LGTM!

PR #1596, opened 2022-06-07
Add Documentation for RespectSchemaVersion

harribo11:
Added documentation for RespectSchemaVersion which was added [here](https://github.com/pulumi/pulumi/pull/8881) but not documented.


guineveresaenger:
cc @iwahbe for a second set of 👁️ 👁️

guineveresaenger (approved):
thank you so much!

PR #1610, opened 2022-06-12
Add Service Provider registry link to docs page

mchaynes:


themes/default/content/docs/intro/pulumi-service/_index.md
```diff
@@ -60,6 +60,7 @@ Collaborate with other developers and coordinate on updates. Provide fine-graine
 Integrate Pulumi with your current continuous delivery pipeline. Build your own extensions, and create reusable templates.
 
 * [CI/CD Integration Assistant]({{< relref "/docs/intro/pulumi-service/ci-cd-integration-assistant" >}})
+* [Pulumi Service Provider]({{ relref "/registry/packages/pulumiservice/" }})
```
mchaynes, line 63:
```suggestion
* [Pulumi Service Provider]({{< relref "/registry/packages/pulumiservice/" >}})
```

susanev (approved):
this section was originally a list of links to pages within this intro > service navigation but it seems to have become something else, so adding this here is fine but at some point, we should possibly rethink how we group service-related docs together. another thing to note is that this is taking them to the registry while the rest of the links in this section are navigating them to docs content.

i do think that this particular link could use a description to help folks understand where this is going to take them.

possibly consider linking here https://www.pulumi.com/docs/reference/pkg/ if ur intent is only to include a link somewhere in the docs.

PR #1622, opened 2022-06-15
Actually fail when the Markdown linter fails

cnunciato:
When the Prettier check was added to `lint.sh`, we forgot to add a line to tell the script to fail and exit on any error, which has allowed a few Markdown-related changes to sneak by that _should've_ failed, but didn't, because the Prettier check that runs _after_ `markdownlint` passed. (In other words, as of now, only a Prettier failure will cause `lint.sh` to exit nonzero, which means we're essentially no longer linting Markdown.)

This change adds a line to handle this as we do elsewhere and fixes up the Markdown things that've gotten through.

### DIFF
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -1,4 +1,6 @@
 #!/bin/bash
 
+set -o errexit -o pipefail
+
 node ./scripts/lint/lint-markdown.js
 yarn prettier --check .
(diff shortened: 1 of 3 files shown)


kimberleyamackenzie (approved):
🎉 thanks @cnunciato !

PR #1631, opened 2022-06-16
[Learn] Import/Migrate pathway

nimbinatus:
fixes https://github.com/pulumi/marketing/issues/230


joeduffy:
This looks incredible! One thing that I wondered as I read is if we can “generate” the import script from the Terraform state file. It’s just JSON, all the types and IDs are in there, and the Pulumi plug-in is capable of giving all the mapping info. I think it could be a 5 line script. If so, might be handy?

joeduffy:
@nimbinatus Not sure what you thought of my latest comment, however, @lukehoban just published this tool which automates basically exactly this idea. It's worth checking out. I do think a lot of folks will be frustrated with the manual process of figuring out IDs and type tokens (indeed I know this as I've gotten that feedback from multiple customers recently). That said, I do know it's a "beta tool" so might not be appropriate to rely on from a Learn pathway. Either way, would love to hear your thoughts either way https://github.com/pulumi/tfstate2pulumiimport

PR #1635, opened 2022-06-17
Change .NET Core to .NET is our docs prose

mikhailshilkov:
The technology is now called ".NET" and the latest version is ".NET 6", which is the only current LTS (long-term support) version. It's the default version that users will get if they follow our links to download pages. We are also about to change our templates and examples to .NET 6, which. also brings nice simplifications into programs.

I searched for all mentions of ".NET Core" and updated everything except blog posts, case studies, and old workshop guide.

### DIFF
--- a/config/_default/menus.yml
+++ b/config/_default/menus.yml
@@ -107,7 +107,7 @@ reference:
     url: https://pkg.go.dev/github.com/pulumi/pulumi/sdk/v3/go/pulumi
     weight: 3
 
-  - name: .NET Core
+  - name: .NET
     parent: Pulumi SDK
     url: /docs/reference/pkg/dotnet/Pulumi/Pulumi.html
     weight: 4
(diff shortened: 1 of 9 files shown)


Frassle (approved):
Looks fine to me

susanev (approved):
@mikhailshilkov should we also update the .net logo, it says .net core.

mikhailshilkov:
@susanev Yes, we should. Who can I ask to prepare the new image file? (I don't try myself with that)

susanev:
@mikhailshilkov I'll do it in a separate pr no worries!

config/_default/menus.yml
```diff
@@ -107,7 +107,7 @@ reference:
     url: https://pkg.go.dev/github.com/pulumi/pulumi/sdk/v3/go/pulumi
     weight: 3
 
-  - name: .NET Core
+  - name: .NET
```
cnunciato, line 110: This is good, but these config files only apply to their local repositories; Hugo doesn’t bring them across when it imports a module, so we’ll need to make this same change in the docs repo as well. (Sorry about that — one of those things.)

cnunciato (approved):
Looks good! Just one little note.

config/_default/menus.yml
mikhailshilkov: To be sure - do I need to open a PR to change this file https://github.com/pulumi/docs/blob/master/config/_default/menus.yml? It looks like that file is actually newer - has Java and YAML - do we need to sync other things?

PR #1642, opened 2022-06-20
Add PULUMI_TEST_MODE to guides/testing/unit

aviflat:
I was pulling my hair out trying to figure out why `pulumi.isDryRun`
was returning `false` when I was running my unit tests, given that
[the docs on the method][1] say:

> Note that we always consider executions in test mode to be
> “dry-runs”, since we will never actually carry out an update, and
> therefore certain output properties will never be resolved.

I’m still not sure if/where there is any canonical documentation on
“test mode” but I feel that it should probably always be enabled when
running unit tests. I’d prefer that there be a programmatic/imperative
way to enable it, e.g. `runtime.setDryRun(true)` or something similar,
but in lieu of that I feel the examples should include setting the env
var so as to plant seeds of awareness about this feature and how it
works.

[1]: https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/pulumi/runtime/#isDryRun

### DIFF
--- a/themes/default/content/docs/guides/testing/unit.md
+++ b/themes/default/content/docs/guides/testing/unit.md
@@ -683,7 +683,7 @@ That's it&mdash;now let's run the tests.
 The command line to run your Mocha tests would therefore be:
 
 ```bash
-$ mocha -r ts-node/register ec2tests.ts
+$ PULUMI_TEST_MODE=true mocha -r ts-node/register ec2tests.ts
 ```
 
 {{% /choosable %}}
@@ -691,7 +691,7 @@ $ mocha -r ts-node/register ec2tests.ts
 Run the following command to execute your Python tests:
 
 ```bash
-$ python -m unittest
+$ PULUMI_TEST_MODE=true python -m unittest
 ```
 
 {{% /choosable %}}
@@ -699,7 +699,7 @@ $ python -m unittest
 Run the following command to execute your Go tests:
 
 ```bash
-$ go test
+$ PULUMI_TEST_MODE=true go test
 ```
 
 {{% /choosable %}}
@@ -707,7 +707,7 @@ $ go test
 Run the following command to execute your Python tests:
 
 ```bash
-$ dotnet test
+$ PULUMI_TEST_MODE=true dotnet test
 ```
 
 {{% /choosable %}}


aviflat:
I may have opened this too soon… based on the docs I’d found, it looked like setting PULUMI_TEST_MODE to `true` would enable test mode, and that would cause `runtime.isDryRun` to return `true` based on the docs on the method as I quoted above.

However, that doesn’t appear to be the case. I added that env var to my test script and then ran it, and `runtime.isDryRun` is still returning `false`.

I did some more digging and eventually found that I can pass additional args to `runtime.setMocks` — the last arg is named `preview` and defaults to `false` but if I set it to `true` then after that `runtime.isDryRun` will return `true`.

This is all quite baroque and confusing.

At the very least, it seems like the docs on `runtime.isDryRun` are out of date — “test mode” does not, in fact, result in the method returning `true`. Instead one needs to either use `runtime.setMocks` and pass `true` for `preview`.

Also, reading through the code, it _looked_ as though setting the env var PULUMI_NODEJS_DRY_RUN should result in `runtime.isDryRun` returning `true`. However, in my case, when running my tests, that didn’t work. My hypothesis is that when I invoked `runtime.setMocks` and didn’t pass a value for the arg `preview`, the default of `false` was applied, overriding the env var. Which is OK I guess, perhaps the env vars aren’t supposed to have an effect during testing.

Anyway… it seems like this PR should be updated; rather than setting PULUMI_TEST_MODE, the examples should pass a value for the arg `preview` to `setMocks` — perhaps with a note that this will result in `isDryRun` returning `true`.

That said, I find it surprising that the default for `preview` is `false` — it seems clear that `setMocks` is meant to support testing, and isn’t testing semantically, intrinsically, always a dry run? So I’d think that if someone doesn’t supply the arg `preview` then it should default to `true`.

susanev:
@aviflat thank you for opening this pr and sry for all the trouble! we will take a look as soon as we can.

susanev:
@mikhailshilkov who on ur team would be best to review this?

aviflat:
> Anyway… it seems like this PR should be updated…

To clarify, I’m willing to update the changes in the PR, but I’d like some guidance from the team first.

iwahbe:
Doing a little digging, `PULUMI_TEST_MODE` exists to allow running unit tests without depending on the engine (for which it is no longer used). It is unrelated to running in preview (which is what `isDryRun()` indicates). Since `PULUMI_TEST_MODE` exists only for python and typescript, I am enclined to discourage it's usage. I agree with @cnunciato that it should be depreciated. I have reopened https://github.com/pulumi/pulumi/issues/3045.

aviflat:
Thanks all.

I’ve pushed:

* c048bcc06bb53 — reverts the original change that added PULUMI_TEST_MODE
* 6a0489d634461 — revises the examples to show how to set the flag `dryRun` via `setMocks` et al

As I noted in the message for 6a0489d634461:

> Now that I’m writing this commit message, the situation with `dryRun` is confusing enough (especially given the docs on `runtime.isDryRun`) that it could probably merit adding a small dedicated section to this page. I’ll ask in the PR whether that’d be desirable.

So what do y’all think? Any interest in the addition of a short section on this topic, just a couple of sentences to inform readers about the relationship between the `dryRun` flag and previewing, etc?

(I still think that the docs for `runtime.isDryRun` should be revised — but that’d be a change for the main `pulumi` repo, so I’ll open a separate issue and/or PR there.)

aviflat:
> So what do y’all think? Any interest in the addition of a short section on this topic, just a couple of sentences to inform readers about the relationship between the `dryRun` flag and previewing, etc?

Such a section would provide an opportunity to mention that `runtime` also has the method `_setIsDryRun` which can be used in unit tests to test code behavior during preview vs update, which’d be reasonable and maybe even prudent if/when someone has used `isDryRun` in their project source.

iwahbe:
Unfortunately it doesn't look like go has a way to set dry run. Given that, @aviflat's new docs have all relevant languages represented. I would definitely support more documentation on `isDryRun` in pulumi/pulumi. @cnunciato any comments?

themes/default/content/docs/guides/testing/unit.md
```diff
@@ -247,7 +247,11 @@ pulumi.runtime.setMocks({
     call: function(args: pulumi.runtime.MockCallArgs) {
         return args.inputs;
     },
-});
+},
+  "project",
+  "stack",
+  false, // Sets the flag `dryRun`, which is true at runtime during a preview.
```
iwahbe, line 253: I'm not sure how clear this language is. I always find it hard to write comments explaining booleans.
```suggestion
  false, // Sets the flag `dryRun`, which indicates if pulumi is running in preview mode.
```
I'm not sure that my suggestion is better. @cnunciato What do you think about the linguistic clarity of the message.

iwahbe (approved):
Thanks for helping us clean up the comments!

themes/default/content/docs/guides/testing/unit.md
cnunciato: @iwahbe I like your suggestion, yeah -- that seems clearer to me. Although looking at the Node.js SDK docs, should `dryRun` be `preview`?

```typescript
/**
 * setMocks configures the Pulumi runtime to use the given mocks for testing.
 *
 * @param mocks: The mocks to use for calls to provider functions and resource consrtuction.
 * @param project: If provided, the name of the Pulumi project. Defaults to "project".
 * @param stack: If provided, the name of the Pulumi stack. Defaults to "stack".
 * @param preview: If provided, indicates whether or not the program is running a preview. Defaults to false.
 */
export declare function setMocks(mocks: Mocks, project?: string, stack?: string, preview?: boolean): void;
```

cnunciato: > So what do y’all think? Any interest in the addition of a short section on this topic, just a couple of sentences to inform readers about the relationship between the `dryRun` flag and previewing, etc?

Yes, this sounds great! The one bit of feedback I have has to do with how this property and the others two (project and stack) are meant to be used in the context of testing. This still isn't clear to me, and I'm sure other users will wonder about it as well, especially when they see them presented alongside `mocks` in this introductory example. (My first question when I saw them was whether they were placeholder values or somehow significant,  and as presented now, it's not clear they're optional.) 

What would you think about pulling all three of these optional properties into this follow-up section as well? That might make the distinction clearer and give you some room to explain how to use them.

aviflat: Yup, on it.

aviflat: > Although looking at the Node.js SDK docs, should `dryRun` be `preview`?

FWIW I do think it should be renamed. But my impression is that since be an API change, it’d be best to propose it in a new issue in pulumi/pulumi — is that right?

> The one bit of feedback I have has to do with how this property and the other two (project and stack) are meant to be used in the context of testing. This still isn't clear to me, and I'm sure other users will wonder about it also, especially when they see them alongside `mocks` in this introductory example. (My first question was whether “project” and “stack” were placeholder values or somehow significant, and as they’re presented now, it's also not clear they're optional.)

Makes 100% sense, this had occurred to me for a fleeting moment but I decided to take a light touch with the changes and see what feedback came back. I’ll address this.

On a related note, fwiw, it seems to me that the name of the method `setMocks` is not fully aligned with how it works / what it does. Yes, you can set the mocks. But it has these additional args that have nothing to do with mocks. They’re more like settings. I understand why they were added; `setMocks` is now used more or less as if it were named `setTestSettings` or something like that. And in fact the docs would be easier to write if it were named something like that. I don’t know if I have the time to make such a proposal to the SDK maintainers, but I figured it couldn’t hurt to share the thought.

cnunciato: > FWIW I do think it should be renamed. But my impression is that since be an API change, it’d be best to propose it in a new issue in pulumi/pulumi — is that right?

💯 couldn't agree more! That'd be awesome if you'd be up for it. I will heavily 👍 this proposal. 😄 

> Makes 100% sense, this had occurred to me for a fleeting moment but I decided to take a light touch with the changes and see what feedback came back. I’ll address this.

Awesome, thanks!

> On a related note, fwiw, it seems to me that the name of the method setMocks is not fully aligned with how it works / what it does.

Ah, that's definitely good feedback, yeah -- makes sense given these additional properties.

iwahbe:
@aviflat I think I may have merged this PR prematurely and if so I'm sorry. Were you done with this change?

aviflat:
I was going to add another commit, but nbd, what was merged is fine, I was just going to add an additional section. I’ll open a follow-up PR.

PR #1646, opened 2022-06-21
Gerdisch/about updates 060122

MitchellGerdisch:
Adding a few team members


susanev (commented):
@MitchellGerdisch you have multiples of some of these pics in here, some with jpeg and some with jpg

susanev (approved):

Issue #1620, opened 2022-06-15
No Documentation on Options ignored by Componets


RobbieMcKinstry:
### What happened?

Some [resource options](https://www.pulumi.com/docs/intro/concepts/resources/options/) don't apply to Component resources and are ignored by the engine when applied to components.

In pulumi/pulumi, we're adding warnings to the engine to inform users when they've applied an option to a resource which will have no effect. (https://github.com/pulumi/pulumi/issues/9704)

This issue is to update the documentation to explain this behavior to users.

Right now, we know the `ignoreChanges` does not apply to Component resources. However, it's possible other options similarly don't have an effect on components (and also don't have this behavior documented). Some other [possible offenders](https://github.com/pulumi/pulumi/pull/9863#issuecomment-1155736892) include:

1. `additionalSecretOutputs`
2. delete/replace/retain options.

I'm investigating these options, and I'll update this issue when I nail down which options are ignored by components.

This fix for this issue would be to include a warning on the documentation page for each of these options.

### Steps to reproduce

1. Visit the [`ignore changes`](https://www.pulumi.com/docs/intro/concepts/resources/options/ignorechanges/) documentation.
2. Read the page.
3. Note that nowhere does it say this option is ignored by components.

### Expected Behavior

I expected an alert at the top of the page that this option is ignored by components.

### Actual Behavior

There's no documentation of this behavior, which defies one's expectations.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

RobbieMcKinstry:
@susanev I guess I mostly wanted someone else to see it. I didn't want to go sticking my nose into somewhere it doesn't belong. So if I could get confirmation that adding an alert to the [ignoreChagnes](https://www.pulumi.com/docs/intro/concepts/resources/options/ignorechanges/) page, the [retainOnDelete page](https://www.pulumi.com/docs/intro/concepts/resources/options/retainondelete/), etc. is acceptable from a UX perspective, I can poke around and try to add it.

PR #1653, opened 2022-06-23
Add warning to resource options docs

RobbieMcKinstry:
This PR documents the warnings generated as part of https://github.com/pulumi/pulumi/pull/9921
It closes https://github.com/pulumi/pulumi-hugo/issues/1620

(Some) resource options are ignored when applied
to component resources. This commit adds a warning
to the docs explaining this behavior.

### DIFF
--- a/themes/default/content/docs/intro/concepts/resources/options/additionalsecretoutputs.md
+++ b/themes/default/content/docs/intro/concepts/resources/options/additionalsecretoutputs.md
@@ -9,7 +9,7 @@ menu:
 ---
 
 {{% notes "warning" %}}
-The `additionalSecretOutputs` resource options has no effect on component resources. If applied to a component resource, it will be ignored. 
+The `additionalSecretOutputs` resource options has no effect on component resources. If applied to a component resource, it will be ignored.
 {{% /notes %}}
 
 The `additionalSecretOutputs` resource option specifies a list of named output properties that should be treated as [secrets]({{< relref "/docs/intro/concepts/secrets" >}}), which means they will be encrypted. It augments the list of values that Pulumi detects, based on secret inputs to the resource.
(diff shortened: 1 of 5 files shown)


susanev:
@RobbieMcKinstry is warning the right level here? If folks don't read this will it be disastrous for them? 

@cnunciato have we written down what the different levels mean in the repo yet?

iwahbe (approved):
Nit: When applied to a resource option, it won't be ignored. It will instead warn. Would it be clearer to say:
```
The `foo` resource option does not apply to component resources.
```

RobbieMcKinstry:
Hm, I'd like the warning to be more than one sentence to provide a little additional clarity. Can we workshop the language a bit?
How about:
> The `foo` resource option does not apply to component resources, and will not have the intended effect.

susanev:
@RobbieMcKinstry examples of how to use the notes are in the style guide
https://github.com/pulumi/pulumi-hugo/blob/master/STYLE-GUIDE.md#examples

RobbieMcKinstry:
Ok, thank you susan! I didn't know there was a style guide; awesome!
By the guide provided, these indeed should sit at the warning level, since ignoring the note "could lead to negative or unexpected consequences." :D

themes/default/content/docs/intro/concepts/resources/options/additionalsecretoutputs.md
```diff
@@ -81,3 +81,7 @@ resources:
 {{< /chooser >}}
 
 Only top-level resource properties can be designated secret. If sensitive data is nested inside of a property, you must mark the entire top-level output property as secret.
+
+{{% notes "warning" %}}
```
cnunciato, line 85: If you want to use warning level, these'll need to be tweaked, as right now they're appearing as regular notes:

 ```suggestion
{{% notes type="warning" %}}
```

PR #1662, opened 2022-06-25
Blog post: Shared config with AWS SSM

cnunciato:
Fixes https://github.com/pulumi/devrel-team/issues/38.


katcosgrove (approved):

themes/default/content/blog/shared-config-with-aws-systems-manager-parameter-store/index.md
```diff
@@ -0,0 +1,405 @@
...
+outputs:
+  motd_param_name: ${motd_param.name}
+  motd_secret_param_name: ${motd_secret_param.name}
+```
+
+At this point, you might be wondering why we're only exporting our parameters' _names_, and not their values.
+
+We certainly _could_ export their values, since we already have them at hand. But if we did, there wouldn't be much point to using Parameter Store at all, since Pulumi itself is perfectly capable of managing plain-text and encrypted values, and stack references give other stacks an easy way of accessing them programmatically. Why not export the values, then?
+
+For one, it'd be less flexible. If we exposed these values as stack outputs, downstream stacks referencing them would only be able to read them at deploy-time --- i.e., _during_ their own Pulumi updates. For `shared-config` values that change rarely, that might be okay, but for others, it'd be less than ideal in that all downstream stacks would have to be updated immediately whenever one of those values were changed.
```
rawkode, line 107: You kind of mention it, but maybe not explicitly enough - but only exporting the name of the parameters actually allows to better RBAC/ABAC restrictions on which downstream stacks can actually read the values too.

rawkode (approved):
Great article. Nicely done

themes/default/content/blog/shared-config-with-aws-systems-manager-parameter-store/index.md
cnunciato: Ah, nice point! I'll see if I can work that in.

PR #1664, opened 2022-06-27
Add Alex Qiu to the team

aq17:


susanev (commented):
@aq17 can you pls use a square image? otherwise this is what it will look like.
rectangle

aq17:
@susanev should be better now 😄

Issue #1686, opened 2022-06-29
s3:CreateBucket in the get started guides is potentially confusing


susanev:
during a recent usability study a participant got a little bit stuck looking for a `s3:CreateBucket` iam permission. can we clarify what we are asking users to do in this step?

[internal slack convo w/ clip](https://pulumi.slack.com/archives/C85BS3LJZ/p1654813123450929)

PR #1691, opened 2022-06-30
removing confusing line about policy in aws get started guide

susanev:
fixes https://github.com/pulumi/pulumi-hugo/issues/1686

### DIFF
--- a/themes/default/content/docs/get-started/aws/begin.md
+++ b/themes/default/content/docs/get-started/aws/begin.md
@@ -64,7 +64,7 @@ Next, install the required language runtime, if you have not already.
 
 ### Configure Pulumi to access your AWS account
 
-Pulumi requires cloud credentials to manage and provision resources. You must use an IAM user account that has **Programmatic access** with rights to deploy and manage resources handled through Pulumi. In this guide, your IAM user must have the `s3:CreateBucket` IAM policy action granted.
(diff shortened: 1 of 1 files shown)


cnunciato:
Do you think that if we delete this one, we should do the same for the other clouds as well? Both Azure and Google Cloud have similar guidance.

susanev:
@cnunciato i looked at azure and google cloud, there language is different and not as confusing. do you feel strongly that they need to be removed?

lukehoban (approved):

cnunciato:
> do you feel strongly that they need to be removed?

Not at all, just wanted to make sure we'd looked at 'em. 

Might be nice to have the AWS example align a bit better with the other two, which do seem clear:

Azure:

> In this guide, you will need a user account with permissions to create and populate Blob storage containers

GCP:

> In this guide, you will need an IAM user account with permissions that can create and populate a Cloud Storage bucket

Whereas with AWS:

>  Pulumi requires cloud credentials to manage and provision resources. You must use an IAM user account that has **Programmatic access** with rights to deploy and manage resources handled through Pulumi.

.. but that's not necessarily something we need to do for this PR.

cnunciato:
> i don't think it makes sense to go further in depth to this in the guide

Nor do I -- I'll just submit a PR that shows what I mean.

Issue #1644, opened 2022-06-21
consider exposing aws profile info in get started guide


susanev:
during a recent usability study, a participant wanted to know if they could use an aws profile instead. they ended up finding the information they needed but it was a bit of a struggle. can we make this information more accessible?

their flow:
- exported aws variables
- wondered if they could use an aws profile instead
- tried a docs search, and could not find anything
- went to google search
- ended up on the config page for aws classic in the registry
- answered their question

[internal clip](https://dovetailapp.com/tags/1T2o1GieqT7LtA9gB9GS01)

stack72:
@susanev should we link to the installation guide there? https://www.pulumi.com/registry/packages/aws/installation-configuration/ this includes the use of `AWS_PROFILE` ?

susanev:
yes, i think we should consider that. that is where the person ended up just in a round-about way.

stack72:
@susanev that installation-configuration page has just been overhauled and I think we should just link to that going forward now?

susanev:
@stack72 its already linked to on that page, i believe the intent of this issue is to mention they can get info about profiles by navigating there. i opened a pr but could not reach alignment, if someone else wants to try pls do!

PR #1692, opened 2022-06-30
fix up content about profiles+config in aws guide

susanev:
fixes https://github.com/pulumi/pulumi-hugo/issues/1644

### DIFF
--- a/themes/default/content/docs/get-started/aws/begin.md
+++ b/themes/default/content/docs/get-started/aws/begin.md
@@ -99,7 +99,7 @@ $ export AWS_SECRET_ACCESS_KEY=<YOUR_SECRET_ACCESS_KEY>
 {{% /choosable %}}
 {{< /chooser >}}
 
-For additional information on setting and using AWS credentials, see [AWS Setup]({{< relref "/registry/packages/aws/installation-configuration" >}}).
+For information about AWS profiles, and configuration, see [AWS Setup]({{< relref "/registry/packages/aws/installation-configuration" >}}).
 
 Next, you'll create a new Pulumi project.
 


themes/default/content/docs/get-started/aws/begin.md
```diff
@@ -99,7 +99,7 @@ $ export AWS_SECRET_ACCESS_KEY=<YOUR_SECRET_ACCESS_KEY>
 {{% /choosable %}}
 {{< /chooser >}}
 
-For additional information on setting and using AWS credentials, see [AWS Setup]({{< relref "/registry/packages/aws/installation-configuration" >}}).
+For information about AWS profiles, and configuration, see [AWS Setup]({{< relref "/registry/packages/aws/installation-configuration" >}}).
```
cnunciato, line 102: Couldn’t we keep the same language and just also mention profiles?

```suggestion
For additional information on setting and using AWS credentials or profiles, see [AWS Setup]({{< relref "/registry/packages/aws/installation-configuration" >}}).
```

susanev: @cnunciato what additional information is on that page that doesn't include profiles or credentials?

susanev: @cnunciato do you specifically not want me to mention config? what credential information is there that's not in the guide and not in the profile content?

dixler:
Hey, @susanev we should mention a code fence with
```
$ aws configure
```
in it in order to imply that if they've set up their machine to work with the AWS CLI or AWS SDK, pulumi will work as well(like other tools they're used to)

I made this suggestion in #1729 and didn't see this PR.

dixler:
@susanev sure! I have #1721 which condenses the `get-started`'s into single pages so we should probably review that first before changing the content.

Issue #1658, opened 2022-06-24
webiny restructured their repo, blog links need to be updated


susanev:
/blog/how-webiny-built-a-serverless-application-framework/

:link: [/blog/how-webiny-built-a-serverless-application-framework/](https://www.pulumi.com/blog/how-webiny-built-a-serverless-application-framework/) → https://github.com/webiny/webiny-js/tree/next/packages/cwp-template-aws/template/common/apps/admin/pulumi (HTTP_404)
:link: [/blog/how-webiny-built-a-serverless-application-framework/](https://www.pulumi.com/blog/how-webiny-built-a-serverless-application-framework/) → https://github.com/webiny/webiny-js/blob/next/packages/cwp-template-aws/template/common/apps/admin/pulumi/app.ts (HTTP_404)
:link: [/blog/how-webiny-built-a-serverless-application-framework/](https://www.pulumi.com/blog/how-webiny-built-a-serverless-application-framework/) → https://github.com/webiny/webiny-js/blob/next/packages/cwp-template-aws/template/common/apps/admin/pulumi/cloudfront.ts (HTTP_404)
:link: [/blog/how-webiny-built-a-serverless-application-framework/](https://www.pulumi.com/blog/how-webiny-built-a-serverless-application-framework/) → https://github.com/webiny/webiny-js/blob/next/packages/cwp-template-aws/template/common/apps/admin/pulumi/index.ts (HTTP_404)
:link: [/blog/how-webiny-built-a-serverless-application-framework/](https://www.pulumi.com/blog/how-webiny-built-a-serverless-application-framework/) → https://github.com/webiny/webiny-js/tree/next/packages/cwp-template-aws/template/common/api (HTTP_404)
:link: [/blog/how-webiny-built-a-serverless-application-framework/](https://www.pulumi.com/blog/how-webiny-built-a-serverless-application-framework/) → https://github.com/webiny/webiny-js/blob/next/packages/cwp-template-aws/template/ddb-es/api/pulumi/index.ts#L19-L25 (HTTP_404)
:link: [/blog/how-webiny-built-a-serverless-application-framework/](https://www.pulumi.com/blog/how-webiny-built-a-serverless-application-framework/) → https://github.com/webiny/webiny-js/blob/next/packages/cwp-template-aws/template/common/apps/admin/pulumi/index.ts#L6-L11 (HTTP_404)
[...]

PR #1695, opened 2022-07-01
Webiny Blog - Update Outdated GH Links

adrians5j:
All of the Webiny GH links were pointing to our `next` branch. Because it changes all the time, this is not a good idea.

So, all of the links are now pointing to our `v5.28.0` branch. That way we're sure none of the links end up being broken over time.

Closes [W1658](https://github.com/pulumi/pulumi-hugo/issues/1658) https://github.com/webiny/webiny-js/issues/2503


adrians5j:
@susanev

stack72 (approved):
Thanks so much for this @adrians5j

susanev:
cc/ @sean1588 heads up on this, since i think you had changed these to next

PR #1699, opened 2022-07-02
Disable Prettier for HTML

### LINKED ISSUE #1639: Don't require two spaces after Go template expressions
The current Prettier config, as inherited from the [`prettier-plugin-go-template`](https://www.npmjs.com/package/prettier-plugin-go-template) parser, enforces two empty lines after a Go template (`{{ ... }}`) expression:

```
</div>

{{ if something }}
    {{ something_else }}
{{ end }}


<div ...>
```
☝️ 😢 

This isn't great, though, as we don't write our template code like this. We should figure out a way around it such that we're linting for one empty line instead. (Sadly, `prettier-plugin-go-template` doesn't appear to expose this as an option.) (cc @kimberleyamackenzie as we've chatted about it out of band.)

cnunciato:
I love the idea of linting our HTML, but the plugin we're currently using for that (`prettier-plugin-go-template`) is more trouble than it's worth, so I'd like to remove it. 

Hopefully we can find a better replacement and switch HTML linting and auto-formatting back on. In the meantime, though, this plugin's not giving us enough value to be worth the hassle. I lose a lot of time to it myself, and I know others hit errors in PRs they can't decipher, too, which usually leads them to ping me for help, I have to step in to unblock them, they have to wait for subsequent checks to pass, etc., all because of random nits that none of us would actually care about in practice. All things considered, and since it's also leaving the code in a not-great state (see #1639), it seems reasonable to pull it out.

cc @kimberleyamackenzie 

Closes #1639.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -9,3 +9,4 @@ public
 yarn-error.log
 .hugo_build.lock
 *.md
+*.html
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -11,12 +11,6 @@
     "useTabs": false,
     "overrides": [
         {
-            "files": ["*.html"],
-            "options": {
-                "parser": "go-template"
-            }
-        },
-        {
             "files": ["*.yml", "*.yaml"],
             "options": {
                 "tabWidth": 2
--- a/package.json
+++ b/package.json
@@ -12,7 +12,6 @@
         "sass": "^1.39.2"
     },
     "devDependencies": {
-        "prettier": "^2.6.2",
-        "prettier-plugin-go-template": "^0.0.13"
+        "prettier": "^2.6.2"
     }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -376,13 +376,6 @@ picomatch@^2.0.4, picomatch@^2.2.1:
   resolved "https://registry.npmjs.org/picomatch/-/picomatch-2.3.0.tgz"
   integrity sha512-lY1Q/PiJGC2zOv/z391WOTD+Z02bCgsFfvxoXXf6h7kv9o+WmsmzYqrAwY63sNgOxE4xEdq0WyUnXfKeBrSvYw==
 
-prettier-plugin-go-template@^0.0.13:
-  version "0.0.13"
-  resolved "https://registry.yarnpkg.com/prettier-plugin-go-template/-/prettier-plugin-go-template-0.0.13.tgz#b4047bce76430bc89a8ee8f27fad1b1c14d942be"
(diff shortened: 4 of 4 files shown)


kimberleyamackenzie:
@cnunciato we talked about this at the last frontend working group meeting, and thought we'd try out a pre-commit hook that automatically formats on commit, to prevent any of the issues with undecipherable errors from Prettier.  That PR is here: https://github.com/pulumi/pulumi-hugo/pull/1750.  If this goes in, and after a bit, we don't feel like it's solving our problem, we can look at dropping Prettier entirely.

susanev:
closing this out as wont fix, pls come to the next front-end working group if you'd like to continue discussing this.

cnunciato:
I was on vacation at the last working group, so unfortunately couldn't make it. I'd like to leave this open and will plan on making the next one. (I haven't had a chance to try out the pre-commit hook, but will do before then.)

susanev:
I believe the whole point of prettier is to eliminate the conversations about what formatting is "correct". that's been my understanding of the direction the frontend working group is trying to move in. glad you can make the next meeting though.

susanev:
one follow-up about your description in this pr...

>which usually leads them to ping me for help

i think we really need to encourage people to ask for help outside of your dms, so this burden (in general outside of this issue) is not solely on you. i recommend encouraging people to use #docs or #frontend depending on the situation.

susanev:
@cnunciato how are you feeling about this? how has it been working for you? do you still want to keep this open?

cnunciato:
@susanev Honestly, I grumble every time I make a commit and have to wait 10-15 seconds while this hook runs, checks hundreds of files I haven't touched, reformats the few I have in a way I don't agree with (and have no control over), and then fails because of something related to Markdown that I'm not interesting in fixing right now -- at which time I remember, _Oh yeah, I need to pass `--no-verify`_, so I pass that and then wait another 10-15 seconds before getting on with my life. 😂. So TBH, I'm not super-stoked with the current setup in general -- but that said, I also don't have a replacement for it at the ready, so I'm willing to cope for the time being until I/we do.

It's worth mentioning, though, that this plugin doesn't work in the Registry, where we're doing some dynamic tag building that the plugin doesn't support. (There's an open issue for this, but it doesn't appear to have been acknowledged by the maintainers, and several others, including this one, that don't seem to be getting fixed.) So since we aren't able to use this plugin in the Registry, and we'll end up needing to do something similar in this repo as well, we'll probably have to to find a replacement sooner or later anyway. But for now, since there's nothing to be done, I'll close.

PR #1719, opened 2022-07-08
Retarget YAML SDK link

squaremo:
- Retarget the link from the SDK page to the YAML spec, which is here as of https://github.com/pulumi/pulumi-hugo/pull/1665; and,
- Remove the implicit menu item, so https://github.com/pulumi/docs/pull/7764 doesn't result in an item in two different places in the navigation.


cnunciato (approved):
@squaremo Perfect! I was going to do this same thing next, for exactly the reasons you [mention here](https://github.com/pulumi/pulumi-hugo/pull/1665#issuecomment-1178743384). Thanks!

PR #1728, opened 2022-07-11
General functions documentation

### LINKED ISSUE #1628: Add "Functions" page in Architecture & Concepts
We don't have a page in our conceptual docs that talks about functions/invokes generally.  We do have a "Getter Functions" page, but that is only a small part of the broader concept on functions.

We should likely rename the "Getter Functions" page to "Functions" and then cover both functions/invokes and Getter Functions.

We also need to decide where to document all the YAML built-in functions.  I'm inclined to believe those should go into https://www.pulumi.com/docs/intro/languages/yaml/ since it is so YAML specific, but they could also in principle go into this "Functions" page.

squaremo:
This turns the explanation of getter functions into a general explanation of provider functions, with a special case of getter functions.

I found it difficult to uncover the purpose of some of the InvokeOptions fields, and I suspect `version` and `pluginDownloadURL` in particular are more for plumbing than for end users, so I have been a little circumspect in places. Advice on improving those and any other bits is welcome!

Fixes #1628.

### DIFF
--- a/themes/default/content/docs/intro/concepts/resources/functions.md
+++ b/themes/default/content/docs/intro/concepts/resources/functions.md
@@ -7,7 +7,7 @@ menu:
     weight: 7
 ---
 
-A provider may make **functions** available in its SDK as well as resource types. For example, the AWS provider includes the function [`aws.apigateway.getDomainName`](https://www.pulumi.com/registry/packages/aws/api-docs/apigateway/getdomainname/), among many others.
+A provider may make **functions** available in its SDK as well as resource types. These "provider functions" are often for calling a platform API to get a value that is not part of a resource. For example, the AWS provider includes the function [`aws.apigateway.getDomainName`](https://www.pulumi.com/registry/packages/aws/api-docs/apigateway/getdomainname/):
 
 <div><pulumi-examples>
 <div><pulumi-chooser type="language" options="typescript,python,go,csharp,java,yaml"></pulumi-chooser></div>
@@ -133,10 +133,9 @@ Provider functions are exposed in each language as regular functions, in two var
 
 The documentation for a provider function will tell you the name and signature for each of the variations.
 
-### Invoke options
+#### Invoke options
 
-Each function also accepts "invoke options", either as an object or as varargs depending on the host
-language. The options are as follows:
+Each function and method also accepts "invoke options", either as an object or as varargs depending on the host language. The options are as follows:
 
 | Option | Explanation                                                  |
 |--------|--------------------------------------------------------------|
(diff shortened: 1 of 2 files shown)


themes/default/content/docs/intro/concepts/resources/functions.md
```diff
@@ -0,0 +1,291 @@
...
+
+| Option | Explanation                                                  |
+|--------|--------------------------------------------------------------|
+| parent | Supply a parent resource, which will be used to determine default providers, and to calculate the protect property. |
+| provider | Supply the provider to use explicitly. |
+| version | Use exactly this version of the provider plugin. |
+| pluginDownloadURL | Download the provider plugin from this URL. The download URL is otherwise inferred from the provider package. |
+| _async_ | _This option is deprecated and will be removed in a future release_ |
+
+The `parent` option has a similar purpose to the [parent option]({{< relref "/docs/intro/concepts/resources/options/parent" >}}) used when creating a resource. The parent is consulted when determining the provider to use, and the [`protect`]({{< relref "/docs/intro/concepts/resources/options/protect" >}}) status of any resources created as a result of the function invocation.
```
lukehoban, line 151: Protect doesn't apply to invokes - so I believe that shouldn't be included here.

themes/default/content/docs/intro/concepts/resources/functions.md
```diff
@@ -0,0 +1,291 @@
...
+
+The `provider` option gives an explicit provider to use when running the invoked function. This is useful, for example, if you want to invoke a function in each of a set of AWS regions.
+
+The `version` option specifies an exact version for the provider plugin. This can be used when you need to pin to a specific version to avoid a backward-incompatible change.
+
+The `pluginDownloadURL` option gives a URL for fetching the provider plugin. It may be necessary to supply this for third-party packages (those not hosted at [https://get.pulumi.com](https://get.pulumi.com)).
+
+## Getter functions
+
+A special case of functions is the static `get` function, which is available on all resource types, to look up an existing resource’s ID. The `get` function is different from the `import` function. The difference is that, although the resulting resource object’s state will match the live state from an existing environment, the resource will not be managed by Pulumi. A resource read with the `get` function will never be updated or deleted by Pulumi during an update.
```
lukehoban, line 161: I'm not sure we should call this a special case of package functions. This is really a wholly different thing. 

Most deeply, while "functions" are just executed and return values - the .get methods actually lead to a resource operation which is recorded in the state file and in the progress display. 

For example, this mean that you can put try/catch around functions, but not around .get (because the latter failing is a resource operation failing in the engine, which will fail the deployment).

The .get calls are a lot more like a `new Res()` call in how they operate - with the key difference that they never create a resource, they always look it up based on id.

lukehoban (approved):

themes/default/content/docs/intro/concepts/resources/functions.md
squaremo: Okay yes, I took the spirit of the original issue as being "both of these are functions and should be discussed together". But as you've explained it just above, it sounds more like they are different enough to each deserve their own treatment, as separate pages.

themes/default/content/docs/intro/concepts/resources/functions.md
squaremo: Do you mean it's irrelevant, or not respected?

lukehoban: There is nothing to protect (there's no resource being managed, so nothing to prevent deletion of), so I suppose "irrelevant".

squaremo: I was unclear on whether functions could create resources. If they respect `parent`, the suggestion is that they might; and if so, why is `protect` irrelevant? (And if not, should I remove `parent` there as well?)

themes/default/content/docs/intro/concepts/resources/functions.md
squaremo: Putting them on the same page will imply that they're mostly the same, even if I remove the offending claim :-) If there's consequences for an end user to mistakenly assuming they are the same, I'd prefer to keep them on separate pages.

themes/default/content/docs/intro/concepts/resources/functions.md
squaremo: I've taken the mention of `protect` out. (The answer to my question is that `Invoke` in the engine doesn't give the callee the means to re-enter the engine, in order to e.g., create resources.)

themes/default/content/docs/intro/concepts/resources/functions.md
squaremo: I rearranged things and gave functions and methods their own page -- they are more similar to each other than to getter functions. I left the latter page as it is.

squaremo:
@lukehoban To address your comments:

 - I gave "provider functions" (invoke) and "methods" (call) their own page, and left the explanation of getter functions as-is.
 - I removed the mention of `protect` from the explanation of invoke options

The methods come with examples in the SDK docs (or perhaps it's just those I looked at), so I had to include the declarations instead, and those are lacking in Java and YAML. It might have to do for now.

susanev:
@squaremo was wondering the status on this pr? do you need any help finishing it out?

squaremo:
> do you need any help finishing it out?

I asked Luke to re-review, since I changed things around to address some of his concerns. But we could quit while ahead and take the original approval as sufficient permission.. wdyt, shall I just merge it?

susanev (approved):

PR #1733, opened 2022-07-12
Fix Java example at create-a-pulumi-project/index.md

AlexITC:
The shown example won't compile, let's fix this.


susanev:
thank you!!

dixler (approved):
Thanks. This is an improvement, however I'm still having an issue with this example(java and docker) on my linux machine, but have other things at the moment.

AlexITC:
@dixler you likely hit https://github.com/pulumi/tutorial-pulumi-fundamentals/issues/5 which I got into, the simpler solution would be to publish an image that isn't tied to arm64.

PR #1736, opened 2022-07-13
Fix #1652: python comments are '#'!

### LINKED ISSUE #1652: Python ECR examples mistranslated, uses "//" for comments
We are using `//` instead of `#` for comments.

See https://www.pulumi.com/docs/guides/crosswalk/aws/ecr/?language=python.  This example, for instance:

```python
import pulumi
import pulumi_awsx as awsx

// Create a repository.
repo = awsx.ecr.Repository("my-repo");

// And publish its URL, so we can push to it if we'd like.
pulumi.export("url", repo.url)
```

Should be:

```python
import pulumi
import pulumi_awsx as awsx

# Create a repository.
repo = awsx.ecr.Repository("my-repo");

# And publish its URL, so we can push to it if we'd like.
pulumi.export("url", repo.url)
```

There are other examples affected too. It's worth auditing beyond just the ECR pages too.

jf:


dixler (approved):
thanks! lgtm

PR #1746, opened 2022-07-18
Remove Fn::StackReference doc

aq17:
Stack references will now be a resource that can be declared, rather than accessed via a function

### DIFF
--- a/themes/default/content/blog/shared-config-with-aws-systems-manager-parameter-store/index.md
+++ b/themes/default/content/blog/shared-config-with-aws-systems-manager-parameter-store/index.md
@@ -150,7 +150,7 @@ $ cd ../my-website
 $ pulumi new aws-yaml
 ```
 
-Once again, follow the prompts to create a new `dev` stack, then replace the contents of `Pulumi.yaml` with the program below. Here, we're fetching the names of the parameters being managed by `shared-config`, pulling their values from Systems Manager, and creating a one-page website to render those values in the browser. (Be sure to adjust the [`Fn::StackReference`](https://github.com/pulumi/pulumi-yaml#fnstackreference)s to point to your stack instead of mine.)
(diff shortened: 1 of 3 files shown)


aq17:
@AaronFriel @iwahbe We should probably document somewhere else that the user can declare stack references, as well as what other resources `pulumi:pulumi:{resourceName}` can be created other than the default `Stack` and now `StackReference` (if any)

AaronFriel:
If you have GitHub Code Search enabled, this query should find any relevant replacements, there are a few in this repo:

https://cs.github.com/?scope=org%3Apulumi&scopeName=pulumi&q=Fn%3A%3AStackReference

Looks like:

https://github.com/pulumi/pulumi-hugo/blob/429ed7bfcd95afc3095b341f5fc682914e402388/themes/default/content/docs/intro/concepts/stack.md#L671

https://github.com/pulumi/pulumi-hugo/blob/429ed7bfcd95afc3095b341f5fc682914e402388/themes/default/content/blog/shared-config-with-aws-systems-manager-parameter-store/index.md#L153

AaronFriel:
It looks like there are still two `Fn::StackReference` usages in concepts/stack.md.

aq17:
> It looks like there are still two `Fn::StackReference` usages in concepts/stack.md.

My bad, thanks for catching that! Think that should be all of them now

aq17:
> I think we should pair this with an equivalent section on how you now _should_ create stack references.

The update for the [Stacks intro page](http://pulumi-hugo-origin-pr-1746-6e119e9e.s3-website.us-west-2.amazonaws.com/docs/intro/concepts/stack/) might suffice? Before YAML was a bit inconsistent in using a function when the docs said to `create an instance of the StackReference type using the fully qualified name of the stack as an input`

PR #1747, opened 2022-07-19
Add k8s guide for server-side apply

lblackstone:


themes/default/content/registry/packages/kubernetes/how-to-guides/managing-resources-with-server-side-apply.md
```diff
@@ -0,0 +1,759 @@
...
+
+{{< chooser language "typescript,python,go,csharp,java,yaml" >}}
+
+{{% choosable language typescript %}}
+
+```typescript
+import * as pulumi from "@pulumi/pulumi";
+import * as kubernetes from "@pulumi/kubernetes";
+
+const provider = new kubernetes.Provider("k8s", {enableServerSideApply: true});
```
joeduffy, line 30: I assume I can also use config instead?

`pulumi config set kubernetes:enableServerSideApply true`

if I am not using explicit providers? Should we call this out?

lblackstone: Yes, good point. I'll update to show the alternative config methods.

viveklak (approved):

PR #1750, opened 2022-07-19
Add pre-commit hook that formats code

kimberleyamackenzie:
In the last frontend working group meeting, we discussed the challenge of mis-formatted code causing confusing errors for folks when their branches were pushed.  We decided that we'd try a pre-commit hook that automatically formats code with Prettier, and see if that resolved some of the difficulties.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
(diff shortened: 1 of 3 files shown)


susanev (approved):

cnunciato:
@kimberleyamackenzie You'll need to push a new commit onto this branch to "kick it" to get past the error in your build logs. (The error has to do with a previous run not finishing properly.)

PR #1755, opened 2022-07-20
Add provider configuration documentation

danielrbradley:


squaremo (approved):
Useful additional info to have in there :+1:

themes/default/content/docs/intro/concepts/config.md
```diff
@@ -284,7 +284,20 @@ public static void stack(Context ctx) {
 
 {{< /chooser >}}
 
-## Special Configuration Options
+## Provider Configuration Options
+
+There are three ways to configure providers:
+
+1. Set configuration keys in the stack configuration file: `pulumi config set [PROVIDER]:[KEY] [VALUE]`
```
lukehoban, line 291: Throughout the rest of this document - this is referred to as "Pulumi configuration" but here we use the term "stack configuration file". Suggest we continue to use "Pulumi configuration" in this section for consistency.

themes/default/content/docs/intro/concepts/config.md
```diff
@@ -284,7 +284,20 @@ public static void stack(Context ctx) {
...
+2. Set a provider-specific environment variable
+3. Pass arguments to the provider's SDK constructor, in your program
+
+Please note:
+
+* Configuration file settings are only used by the default provider. If you instantiate a provider object, it will not read values from the stack configuration.
```
lukehoban, line 297: Cross link "provider" to the documentation on Provider's (and possibly capitalize and format as code `Provider`?)

Cross link default provider to documentation on default providers?

themes/default/content/docs/intro/concepts/config.md
```diff
@@ -284,7 +284,20 @@ public static void stack(Context ctx) {
...
+* The precedence of configuration sources (configuration file, environment and args) can vary between providers. Please refer to the provider's documentation for specific configuration instructions.
```
lukehoban, line 298: Perhaps cross link to the registry?

themes/default/content/docs/intro/concepts/config.md
```diff
@@ -284,7 +284,20 @@ public static void stack(Context ctx) {
...
+
```
lukehoban, line 299: Is there an example that would be useful here?  Possibly not - especially if cross linking works to send folks to content that does have examples.

Issue #1759, opened 2022-07-20
Inconsistency in calling stack output.


zagruby:
[themes/default/content/docs/get-started/azure/deploy-stack.md](https://www.pulumi.com/docs/get-started/azure/deploy-stack/)

the line for calling stack works fine with python tutorial:
`pulumi stack output primary_storage_key`

however in Go this line produces an error message as it should be changed to `pulumi stack output primaryStorageKey `

PR #1761, opened 2022-07-21
add missing chooser to aws, azure, gcp deploy stack step

susanev:
this came up in a recent study, and someone opened up an issue https://github.com/pulumi/pulumi-hugo/issues/1759
its not clear that this command can vary between languages because this was missing
this adds the chooser to reduce confusion

fixes  https://github.com/pulumi/pulumi-hugo/issues/1759


mikhailshilkov (approved):
This makes sense given the difference between languages. Alternatively, we can change our Python template and tutorial to export `bucketName` instead of `bucket_name` to align it to all other languages.

susanev:
id still argue that leaving out the language chooser leads to confusion even if the command is the same between languages; feel free to open a new pr with the changes you've proposed, happy to support them.

PR #1768, opened 2022-07-22
Add docs about plugins section in Pulumi.yaml

Frassle:
See https://github.com/pulumi/pulumi/pull/10146 for implementation

### DIFF
--- a/themes/default/content/docs/reference/pulumi-yaml.md
+++ b/themes/default/content/docs/reference/pulumi-yaml.md
@@ -48,3 +48,9 @@ It contains the following required and optional attributes:
         - `description`: (optional) a description for the config setting.
         - `default`: (optional) the default value of the config setting - which will be presented to the user as a default.
         - `secret`: (optional) if `true` indicates that this configration value should be marked as secret.
+
+- `plugins`: (optional) allows the overriding of plugin selection to make use of plugins not installed into the global plugin cache
+  - `providers`/`analyzers`/`languages`: Each entry has a list of plugins underneath it.
+    - `name`: (required) the name of the plugin.
+    - `path`: (required) the path to the folder containing the plugin.
+    - `version`: (optional) the version of the plugin, if not set this plugin will match for any version the engine requests.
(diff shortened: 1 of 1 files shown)


harribo11:
Should we mark providers/languages/analyzers as optional?

harribo11 (approved):

themes/default/content/docs/reference/pulumi-yaml.md
```diff
@@ -48,3 +48,9 @@ It contains the following required and optional attributes:
         - `description`: (optional) a description for the config setting.
         - `default`: (optional) the default value of the config setting - which will be presented to the user as a default.
         - `secret`: (optional) if `true` indicates that this configration value should be marked as secret.
+
+- `plugins`: (optional) allows the overriding of plugin selection to make use of plugins not installed into the global plugin cache
```
susanev, line 52:
```suggestion
- `plugins`: (optional) allows the overriding of plugin selection to make use of plugins not installed into the global plugin cache.
```

themes/default/content/docs/reference/pulumi-yaml.md
```diff
@@ -48,3 +48,9 @@ It contains the following required and optional attributes:
...
+  - `providers`/`analyzers`/`languages`: (optional) Each entry has a list of plugins underneath it.
```
susanev, line 53:
```suggestion
  - `providers`/`analyzers`/`languages`: (optional) each entry has a list of plugins underneath it.
```

would only saying `list of plugins.` make sense?

susanev (approved):

themes/default/content/docs/reference/pulumi-yaml.md
Frassle: Yes, but given the slight odd format of this I was erring on a longer description so people trigger that this is multiple entries like:
```
plugins:
  providers:
  - name: aws
    path: ../aws
  analyzers:
  - name: company_policy
    path: ./policy
```

PR #1792, opened 2022-07-28
Create UX-STANDARDS.md

susanev:
this is only a start, and a space for us to add stuff over time.
this doc is collectively owned and everyone is welcome to propose changes to this doc going forward.

### DIFF
--- a/UX-STANDARDS.md
+++ b/UX-STANDARDS.md
@@ -2,10 +2,11 @@
 
 These standards include some basic user experience guidelines we align to on the marketing website, docs, and the registry.
 
-- Every page should have exactly one primary action.
-- Use descriptive text for button labels and links, and avoid labels such as "Learn more".
-- Use unique labels for distrinct actions.
-- Use existing components over creating new components unless you identify a need that is not addressed by the current components.
-- Use line lengths between 45 – 75 characters.
+- Every page should have no more than one primary action.
+- Use descriptive text for button labels and links, and avoid labels such as "Learn more."
+- Use unique labels for distinct actions i.e., having two buttons with the same label that take different actions is confusing to users.
+- Use existing components, or improve existing components over creating new components unless you identify a need that is not solveable with the current components.
+- Use line lengths between 45 – 75 characters (this is in reference to the rendered page, not the markdown).
 - Links should take you to another page, and buttons should take an action on the page.
-- When possible links should look like links, and buttons should look like buttons.
+- When possible, links should look like links, and buttons should look like buttons.
+- When not a special case (e.g., navigation), links should exist within the context of text.
--- a/UX-STANDARDS.md
+++ b/UX-STANDARDS.md
@@ -1,0 +1,11 @@
+# UX Standards
+
+These standards include some basic user experience guidelines we align to on the marketing website, docs, and the registry.
+
+- Every page should have exactly one primary action.
+- Use descriptive text for button labels and links, and avoid labels such as "Learn more".
+- Use unique labels for distrinct actions.
+- Use existing components over creating new components unless you identify a need that is not addressed by the current components.
+- Use line lengths between 45 – 75 characters.
+- Links should take you to another page, and buttons should take an action on the page.
+- When possible links should look like links, and buttons should look like buttons.


UX-STANDARDS.md
```diff
@@ -0,0 +1,11 @@
+# UX Standards
+
+These standards include some basic user experience guidelines we align to on the marketing website, docs, and the registry.
+
+- Every page should have exactly one primary action.
+- Use descriptive text for button labels and links, and avoid labels such as "Learn more".
```
cnunciato, line 6: I actually think labels like these -- "Get started", "Learn more" -- are fine when the context is clear, especially on buttons (and indeed we do this ~everywhere), but I agree that in prose, we should definitely be specific. Maybe some examples would help make this distinction clearer?

UX-STANDARDS.md
```diff
@@ -0,0 +1,11 @@
...
+- Use unique labels for distrinct actions.
```
cnunciato, line 7: I'm sure I'd agree, but I'm not sure I necessary have the same thing in mind as you do here. An example or two here would help make this clearer as well.

UX-STANDARDS.md
```diff
@@ -0,0 +1,11 @@
...
+- Use existing components over creating new components unless you identify a need that is not addressed by the current components.
```
cnunciato, line 8: Personally I'd go a bit farther here and suggest that even when a need _is_ identified that isn't addressed by an existing component (assuming we're talking about web components here), we should still consider extending the existing component first. Sometimes it won't make sense to do that (e.g., with components that are purpose-built, which is common and IMO fine), but in general we should lean toward extending those components that were built with reusability in mind.

cnunciato (commented):
Thanks for writing this up! Some initial feedback for ya.

UX-STANDARDS.md
susanev: this goes against user experience best practices. can you link me to some sources that document that these best practices have changed?

UX-STANDARDS.md
```diff
@@ -0,0 +1,11 @@
+# UX Standards
+
+These standards include some basic user experience guidelines we align to on the marketing website, docs, and the registry.
+
+- Every page should have exactly one primary action.
```
jkodroff, line 5: Could we link to an example? I'm familiar with what this means in the context of e.g. a dialog box, but not in the context of a content-centric page.

UX-STANDARDS.md
jkodroff, line 7:
```suggestion
- Use unique labels for distinct actions.
```

UX-STANDARDS.md
```diff
@@ -0,0 +1,11 @@
...
+- Use descriptive text for button labels and links, and avoid labels such as "Learn more".
+- Use unique labels for distrinct actions.
+- Use existing components over creating new components unless you identify a need that is not addressed by the current components.
+- Use line lengths between 45 – 75 characters.
```
jkodroff, line 9: For Markdown, I do not think this is desirable. Because whitespace is meaningful (especially when suggesting edits), I would recommend Markdown paragraphs go on a single line and instead use soft wrap in the editor.

susanev: I will update this, this is specific to the user experience not the markdown that creates that experience.

jkodroff: So we're basically referring to the CSS that would set the width of the enclosing element?

susanev: Unfortunately, our css is kind of a mess so there's not a single rule that will fix this everywhere.

kimberleyamackenzie: I actually just poked into this, and we could probably do a global-ish fix pretty easily with the `ch` unit for width, if you wanted to make a ticket.  It wouldn't fix everything if we have overriding specificity, but could be a good start.

kimberleyamackenzie (approved):
This LGTM!  I agree that adding examples over time will be helpful, but I think this is a great starting point for folks to add as they'd like.  

Kind of an aside, but I wonder if there's somewhere else this can live too/be duplicated because right now, it's mostly guidelines that are also true for the service and would prob be helpful as kind of a global set of guidelines.  But that's totally just something to keep in the back of our minds, not a problem needing solving now.

UX-STANDARDS.md
susanev: cool, will do!

susanev:
>Kind of an aside, but I wonder if there's somewhere else this can live too/be duplicated because right now, it's mostly guidelines that are also true for the service and would prob be helpful as kind of a global set of guidelines. But that's totally just something to keep in the back of our minds, not a problem needing solving now.

@kimberleyamackenzie i agree! the only reason I'm starting here is because the style guide has been somewhat successful in this repo as a resource. if the ux standards end up being a useful tool then we can figure out how to define this for multiple repos without duplication.

UX-STANDARDS.md
jonelleb: I agree with using descriptive text for button labels and links. Here's a little something I came across regarding this: https://www.accessguide.io/guide/descriptive-link-text
It gives examples of how we can compromise, for example, instead of saying "Learn more" we could say "Learn more about Pulumi" or whatever we want the user to learn more about. Possibly "Get started" could be the exception? Since I'm thinking that in situations where we use that, the context would be very clear and obvious.

susanev: ya, get started should be an exception, especially because we have indexed our SEO so much on it

jonelleb: I'm not sure I agree about every page having one primary action... I think this is a good rule to have in general, but couldn't there be instances/exceptions where we have more informational type pages where really the only action a user might take is to go back to the previous page?

UX-STANDARDS.md
```diff
@@ -0,0 +1,11 @@
...
+- Links should take you to another page, and buttons should take an action on the page.
+- When possible links should look like links, and buttons should look like buttons.
```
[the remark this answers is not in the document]
jonelleb: I like that we're making a distinction between links and buttons and both their individual purposes, as well as noting that the two should look different. Since buttons should always take an action on the page, I'm assuming links would primarily be within the context of the text on a page?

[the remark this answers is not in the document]
susanev: yes! i think thats a good thing to be clear about here

anita-trimbur (approved):
I'm in agreement with this as a standard and it seems like the conversation has brought out any mentionable nuances. LGTM!

UX-STANDARDS.md
susanev: @jonelleb absolutely! let me fix this up to say "no more than" for the case you describe in which we would not have a primary button. though i would like us to strive to get to a place where we are building pages intentionally around a single primary action.

PR #1794, opened 2022-07-28
Fix a couple of papercuts

cnunciato:
Fixes a couple of things that came up in a recent working group meeting:

* Mentions self-managed options in the getting-started CLI note
* Fixes the (totally broken!) EC2 instance example 

cc @joeduffy on the former just 'cause it was something you mentioned.

### DIFF
--- a/themes/default/content/docs/aws/ec2.md
+++ b/themes/default/content/docs/aws/ec2.md
@@ -20,29 +20,87 @@ The `@pulumi/aws` library enables fine-grained control over the AWS EC2 resource
 ```javascript
 const aws = require("@pulumi/aws");
 
-const eip = new aws.ec2.Eip("myeip");
-
-const securityGroup = new aws.ec2.SecurityGroup("mysecuritygroup", {
-    ingress: [
-        { protocol: "tcp", fromPort: 80, toPort: 80, cidrBlocks: ["0.0.0.0/0"] },
-    ],
+// Create a VPC.
+const vpc = new aws.ec2.Vpc("vpc", {
+    cidrBlock: "10.0.0.0/16",
 });
 
-const vpc = new aws.ec2.Vpc("myvpc", {
-    cidrBlock: "10.0.0.0/16"
-})
-
-const internetGateway = new aws.ec2.InternetGateway("myinternetgateway", {
+// Create an an internet gateway to give the instance access to the internet.
+const internetGateway = new aws.ec2.InternetGateway("internet-gateway", {
     vpcId: vpc.id,
 });
 
-const publicRouteTable = new aws.ec2.RouteTable("myroutetable", {
+// Create a public subnet.
+const subnet = new aws.ec2.Subnet("subnet", {
+    vpcId: vpc.id,
+    cidrBlock: "10.0.1.0/24",
+    mapPublicIpOnLaunch: true,
+});
+
+// Create a route table.
+const routes = new aws.ec2.RouteTable("routes", {
+    vpcId: vpc.id,
     routes: [
         {
             cidrBlock: "0.0.0.0/0",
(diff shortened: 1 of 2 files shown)


themes/default/layouts/shortcodes/cli-note.html
```diff
@@ -5,11 +5,10 @@
...
-            <a href="https://app.pulumi.com" target="_blank">Pulumi service</a>. The <a href="{{ relref . "/docs/reference/cli" }}">Pulumi CLI</a> works in tandem with the Pulumi
-            service in order to deliver a reliable experience. It is free for individual use, with
-            <a href="{{ relref . "/pricing" }}" target="_blank"> features available for teams</a>. Hitting <code>ENTER</code> at the prompt opens up a web browser allowing you to
-            either sign in or sign up.
+            If this is your first time running <code>pulumi new</code> or other <code>pulumi</code> commands, you may be prompted to sign in to the
+            <a href="https://app.pulumi.com" target="_blank">Pulumi Service</a>. The Pulumi <a href="{{ relref . "/docs/reference/cli" }}">CLI</a> and Service
+            work in tandem to deliver a reliable experience. It's free for individual use, with <a href="{{ relref . "/pricing" }}">features available for teams</a>, and
+             <a href="{{ relref . "/docs/intro/concepts/state" }}">self-managed options</a> are also available. Hitting Enter at the prompt opens a browser for you to sign in or sign up.
```
susanev: - pedantically speaking the cli uses `log in` not `sign in`.
- could we keep `Pulumi` in the CLI link? and also include it in the Service mention?

cnunciato: > pedantically speaking the cli uses log in not sign in

Yeah, despite that you "sign in" to the Service pretty much everywhere, sticking with "login" or "log in" at the CLI level since that maps to the command seems like a fine thing to me. Will tweak.

> could we keep Pulumi in the CLI link? and also include it in the Service mention

Ah yes, that was unintentional -- I'll fix.

themes/default/content/docs/aws/ec2.md
```diff
@@ -20,29 +20,87 @@ The `@pulumi/aws` library enables fine-grained control over the AWS EC2 resource
...
+            fromPort: 0,
+            toPort: 0,
+            protocol: "-1",
+        },
+    ],
+});
+
+// Create and launch an Amazon Linux EC2 instance into the public subnet.
+const instance = new aws.ec2.Instance("instance", {
+    ami: "ami-098e42ae54c764c35",
```
stack72, line 86: I think the only enhancement here would be to look up the AMI ID but that's just an enhancement

themes/default/content/docs/aws/ec2.md
```diff
@@ -20,29 +20,87 @@ The `@pulumi/aws` library enables fine-grained control over the AWS EC2 resource
...
+});
+
+// Associate the route table with the public subnet.
+const routeTableAssociation = new aws.ec2.RouteTableAssociation("route-table-association", {
+    subnetId: subnet.id,
+    routeTableId: routes.id,
+});
+
+// Create an elastic IP for the public subnet.
+const elasticIP = new aws.ec2.Eip("elastic-ip", {
```
stack72, line 58: This doesn't do anything...

stack72 (approved):

themes/default/content/docs/aws/ec2.md
cnunciato: Oh, ha, look at that. 😄 Thanks, copy-pasta.

themes/default/content/docs/aws/ec2.md
cnunciato: I like that, I'll add.

PR #1797, opened 2022-08-02
Troubleshoot Pulumi & system wide proxy server

ringods:
System wide proxy should not proxy local traffic required by Pulumi

Signed-off-by: Ringo De Smet <ringo@de-smet.name>

### DIFF
--- a/themes/default/content/docs/support/troubleshooting.md
+++ b/themes/default/content/docs/support/troubleshooting.md
@@ -196,6 +196,24 @@ You may encounter an error when you downgrade provider versions _after_ your sta
 If you must downgrade the version of a provider your `pulumi` program depends on, you will need to [manually edit your deployment](#editing-your-deployment)
 and change the version of the provider your stack depends on and then import that as the latest state of your stack.
 
+### Nothing happens when running Pulumi
+
+You run Pulumi and nothing happens, with output resembling like this:
+
+```
+$ pulumi up
+Previewing update (<stack name>):
+
+Resources:
+
+$
+```
+
+You probably have a system-wide proxy server running on your machine. The [Pulumi architecture ](https://www.pulumi.com/docs/intro/concepts/how-pulumi-works/)
+has 3 different components, running as separate processes which talk to each other using a bidirectional gRPC protocol
+on IP address `127.0.0.1`. Your proxy server should be configured **NOT** to proxy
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/support/troubleshooting.md
```diff
@@ -196,6 +196,24 @@ You may encounter an error when you downgrade provider versions _after_ your sta
...
+
+Resources:
+
+$
+```
+
+You probably have a system-wide proxy server running on your machine. The [Pulumi architecture ](https://www.pulumi.com/docs/intro/concepts/how-pulumi-works/)
+has 3 different components, running as separate processes which talk to each other using a bidirectional gRPC protocol
+on IP address `127.0.0.1`. Your proxy server should be configured **NOT** to proxy
+these local network connections. Check if you can add both `127.0.0.1` and `localhost` to the exclusion list of your proxy server.
```
susanev, line 215:
```suggestion
If you have a system-wide proxy server running on your machine, it may be misconfigured. The [Pulumi architecture ](https://www.pulumi.com/docs/intro/concepts/how-pulumi-works/)
has three different components, running as separate processes which talk to each other using a bidirectional gRPC protocol
on IP address `127.0.0.1`. Your proxy server should be configured **NOT** to proxy
these local network connections. Add both `127.0.0.1` and `localhost` to the exclusion list of your proxy server.
```

themes/default/content/docs/support/troubleshooting.md
```diff
@@ -196,6 +196,24 @@ You may encounter an error when you downgrade provider versions _after_ your sta
 If you must downgrade the version of a provider your `pulumi` program depends on, you will need to [manually edit your deployment](#editing-your-deployment)
 and change the version of the provider your stack depends on and then import that as the latest state of your stack.
 
+### Nothing happens when running Pulumi
```
susanev, line 199: can we add proxy to this heading somehow?

susanev (approved):
the headings are an absolute mess on this page, i will clean them up in a separate pr.

thx for adding this!! only take whats useful to you <3

themes/default/content/docs/support/troubleshooting.md
ringods: I will review it later today. 😉

ringods: Updated the title.

PR #1798, opened 2022-08-02
A handful of updates

joeduffy:
Here are some updates based on recent conversations and feedback:

* Update the homepage subheading to be a bit clearer about what
  Pulumi is and what it does.

* Change the Cloud Engineering menu item to Why Pulumi, and sort
  it to the right of Product.  This more clearly speaks to what
  this navigation item represents as well as its importance
  relative to learning about Pulumi's Product offerings.

* Change the Pulumi Product landing page to say Pulumi Overview,
  not Cloud Engineering Platform, since that's what you click to
  get there and it currently feels very disconnected from all of
  the surrounding marketing which speaks in terms of IaC.

* Add a paragraph to that overview page speaking simply to what
  Pulumi is, and making sure to mention open source.

* Add a sentence to the Get Started page about open source.

* Add more customer logos -- we should be updating this much more
  often!!  Also change the logos to scroll horizontally so that
  we aren't afraid to continue adding new ones from time to time.

* Add recent news stories -- we should be updating this much
  more often too!!  Apparently it's been over a year, despite having
  tons of news coverage to share.

* Update About Us page to be closer to present reality.


themes/default/content/_index.md
```diff
@@ -1,13 +1,12 @@
...
 title: Pulumi - Universal Infrastructure as Code
 meta_desc: Pulumi's open source infrastructure as code SDK enables you to create, deploy, and manage infrastructure on any cloud, using your favorite languages.
 
 hero:
   title: [ "Universal", "Infrastructure as Code" ]
   description: |
-    Program the cloud and bridge the Dev and Ops divide with cloud engineering.
-    <br />From open source to enterprise.
+    Define infrastructure on any cloud using familiar programming languages.
+    For developers, infrastructure teams, and everybody in between.
```
susanev, line 9: without the <br/> i believe this will display all on one line, which will be difficult to read

themes/default/content/product/_index.md
```diff
@@ -4,13 +4,18 @@ meta_desc: Learn how Pulumi's products enable your team to get code to any cloud
 type: page
 layout: product
 
-heading: Pulumi Cloud Engineering Platform
+heading: Pulumi Overview
```
susanev, line 7: very supportive of this change, but i believe that marketing had previously advocated against this as they were trying to build up the cloud engineering connection.

susanev (approved):
thank you for these updates!!

a couple things...
- the tfir logo has white space built into it which will make it artificially smaller than the rest, happy to push a commit if you'd like me to fix it up
- we removed the scrolling previously as it was very abrupt and not working well, supportive of adding it back if we no long believe its a usability issue

joeduffy:
> the tfir logo has white space built into it which will make it artificially smaller than the rest, happy to push a commit if you'd like me to fix it up

Ah, good catch, if you're able to do this easily, that'd be great, else I can fix it up in a bit.

> we removed the scrolling previously as it was very abrupt and not working well, supportive of adding it back if we no long believe its a usability issue

I fiddled a bit with the settings -- as well as ensuring the sizes look good on mobile -- and it seems to work pretty well now.  It's very similar to what we're using for the social quotes a few lines down.  Take a look and let me know if you're concerned.  The ability to keep cramming infinite amounts of logos in there is very appealing :-)

themes/default/content/_index.md
joeduffy: It actually looks great to my eyes:
image

In fact, the `<br>` has the effect of forcing awkward line breaks on different mobile screens.  For example:

image

The enclosing div has a max width set, so I think we ought to just let the CSS do its thing?

themes/default/content/product/_index.md
joeduffy: I know :-)  I still explained what Cloud Engineering is on the page, but the current experience just doesn't make much sense IMHO -- especially now that we're leaning heavily on the IaC messaging.

themes/default/content/_index.md
susanev: ah right i guess the container has a max width that's wrapping stuff. the dangling `For` isn't great, but no need to block.

PR #1803, opened 2022-08-09
fix: Adds Java and YAML language-specific shortcodes

kpitzen:
While working through some onboarding tasks, I ran into this bug:
pulumidocsbug

While selecting Java or YAML, the language names and language hosts would revert to `TypeScript` and would not recover without a refresh.  This PR adds those languages and corresponding runtimes.

### DIFF
--- a/themes/default/content/docs/support/troubleshooting.md
+++ b/themes/default/content/docs/support/troubleshooting.md
@@ -209,8 +209,7 @@ Resources:
 $
 ```
 
-If you have a system-wide proxy server running on your machine, it may be misconfigured. The [Pulumi architecture ](https://www.pulumi.com/docs/intro/concepts/how-pulumi-works/)
-has three different components, running as separate processes which talk to each other using a bidirectional gRPC protocol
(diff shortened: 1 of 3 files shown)


themes/default/content/docs/support/troubleshooting.md
```diff
@@ -209,8 +209,7 @@ Resources:
 $
 ```
 
-If you have a system-wide proxy server running on your machine, it may be misconfigured. The [Pulumi architecture ](https://www.pulumi.com/docs/intro/concepts/how-pulumi-works/)
-has three different components, running as separate processes which talk to each other using a bidirectional gRPC protocol
+If you have a system-wide proxy server running on your machine, it may be misconfigured. The [Pulumi architecture](https://www.pulumi.com/docs/intro/concepts/how-pulumi-works/) has three different components, running as separate processes which talk to each other using a bidirectional gRPC protocol
```
kpitzen, line 212: I also encountered a linting error here.

mikhailshilkov (approved):
LGTM ❤️

PR #1805, opened 2022-08-09
Add lint script to pre-commit hook

zchase:
Recently we added a pre-commit hook that runs `prettier` and in a recent blog post PR, I hit a lint error which failed my first build. This PR adds the lint script to the pre-commit hook so that folks will see the lint error before they push up to CI, hopefully saving folks time.


kimberleyamackenzie (approved):
I'm 💯 % on board with this, nice to not have to remember another manual command to run

PR #1807, opened 2022-08-09
remove some simplys

susanev:
we recently added avoiding the use of the word simply because it assumes everything is simple for everyone which is rarely true and can set folks up to feel quite awful if its not simple for them. this pr is removing some simplys we had in our docs. i left blogs, marketing material, and partner papers alone.


guineveresaenger (approved):
yesssss thank you this is epic!

PR #1811, opened 2022-08-10
Improve download discoverability

joeduffy:
We know users have trouble simply downloading Pulumi. This used to be
very easy, but over time, as we optimized the Getting Started flows, it
got pushed further and further away from the core user experience.

I don't know about you, but the first thing I care about when I'm
trying out a new open source tool is downloading it!

This change aims to do two things:

1. Make downloading a more prominent CTA.

2. Improve the download page so it's less noisy and more focused.

This entails:

* Adding a DOWNLOAD secondary CTA on the homepage.

* Summarizing the recommended download options at the top of the
  download page, very clearly, and without any preamble. This
  hopefully tells you instantly what you wanted for the 80% case.

* I exerted some artistic freedom which I'd love feedback on. The
  recommended options were our Official Brew Tap for macOS, curl
  command for Linux, and MSI Installer for Windows. Peers to those
  are simple download links for the binaries, as that's the simplest
  possible thing, which today is actually the hardest thing to find.
  Notably for Windows, I thought of using Chocolatey or Winget, but
  I don't perceive that either is "the default" for Windows users.
  Winget is the future but it isn't supported pre-Win11, which I have
  to assume most users aren't on yet. MSI has been around since the
  dinosaurs, so it seems like the safest choice to promote.

* Moving the list of download options for each operating system
  underneath a collapsable accordion list, which is collapsed by
  default, and clearly labeling it as "Other"; as in, if the heading
  didn't work for you, here are some other options.

* A few other wordsmithing tweaks to make the page a little more
  streamlined and to flow better.

This is absolutely NOT the final destination for any of this,
however, I am hopeful it will be a simple incremental improvement
that moves the needle on key metrics. We'll watch it in the weeks
to come and course correct as needed -- as well as continuing to
think about ways we can improve all of this overall!

One note: This depends on a new secondary hero button style that
isn't yet merged in the upstream Hugo component library. Assuming
I did that correctly (a big if!) I'll need to rev the go.mod file
after it merges. See: https://github.com/pulumi/theme/pull/159.


themes/default/content/docs/get-started/install/_index.md
```diff
@@ -1,6 +1,6 @@
 ---
-title: "Download and Install"
-meta_desc: This page contains detailed instructions for installing Pulumi.
+title: "Download Pulumi"
```
susanev, line 2: slight hesitation about loosing the install language here

themes/default/content/docs/get-started/install/_index.md
```diff
@@ -15,20 +15,47 @@ NOTE: To update this page with a new binary release, do the following:
...
 macOS Sierra (10.12) or later is required.
 
-### Official Pulumi Homebrew Tap
+The latest version of Pulumi is {{< latest-version >}}. For older versions, see [Available Versions]({{< relref "/docs/get-started/install/versions" >}}).
+
+{{< get-started-note >}}
+
+## Other Installation Methods
+
+In addition to the above, there are many ways to install Pulumi described below:
```
susanev, line 47:
```suggestion
In addition to the above, there are many ways to install Pulumi described:
```

directional words are rarely accessible, its better to link to the section you are describing or drop them entirely

themes/default/content/docs/get-started/install/_index.md
```diff
@@ -92,12 +162,47 @@ If you do not wish to use the previous options, you can install Pulumi manually.
...
+</div>
+</div>
 
+The latest version of Pulumi is {{< latest-version >}}. For older versions, see [Available Versions]({{< relref "/docs/get-started/install/versions" >}}).
+
+{{< get-started-note >}}
+
+## Other Installation Methods
+
+In addition to the above, there are many ways to install Pulumi described below:
```
susanev, line 194:
```suggestion
In addition, there are many ways to install Pulumi:
```

themes/default/content/docs/get-started/install/_index.md
```diff
@@ -117,14 +232,48 @@ Alternatively, you can install Pulumi manually. We provide a prebuilt binary for
...
 Windows 8 and later are supported.
 
-### Chocolatey
+The latest version of Pulumi is {{< latest-version >}}. For older versions, see [Available Versions]({{< relref "/docs/get-started/install/versions" >}}).
+
+{{< get-started-note >}}
 
+## Other Installation Methods
+
+In addition to the above, there are many ways to install Pulumi described below:
```
susanev, line 265:
```suggestion
In addition, there are many ways to install Pulumi:
```

themes/default/content/docs/get-started/install/_index.md
```diff
@@ -153,12 +313,33 @@ To update Pulumi to a more recent version:
...
+<input type="checkbox" class="absolute hidden" id="windows-standalone-installer" />
+<label for="windows-standalone-installer" class="accordion-label">
+<h5 class="mt-2 w-2/3">Standalone Installer (MSI)</h5>
+<div class="flex flex-grow justify-end items-center">
+<span class="closed-accordion">+</span>
+<span class="open-accordion hidden">-</span>
+</div>
+</label>
+<div class="accordion-item-body-no-animation text-base">
+You can download the latest [Pulumi Installer for Windows x64](https://github.com/pulumi/pulumi-winget/releases/download/{{< latest-version >}}/pulumi-{{< latest-version >}}-windows-x64.msi) and run it like any other installer. It will automatically add Pulumi to the path and make it available machine-wide.
```
susanev, line 329:
```suggestion
Download the latest [Pulumi Installer for Windows x64](https://github.com/pulumi/pulumi-winget/releases/download/{{< latest-version >}}/pulumi-{{< latest-version >}}-windows-x64.msi) and run it like any other installer. It will automatically add Pulumi to the path and make it available machine-wide.
```

susanev (commented):
im not interested in blocking this, but that download page is sooo heavy and has so much content on it when most folks are just going to want the command we give them. that command is about the least prominent thing on the page thanks to our code block styles and the new headings you added above are also in a battle with folks being able to see the download command. happy for you to go forward, but it would be rad to find time to fix the user experience of these things.

other future ideas ill mention...
- displaying the download command directly on the home page, with a link to other download options (the download page if needed)
- revisiting the get started page i designed and implemented a while ago
- displaying the download command in the footer, so its accessible from every single page that has the footer (Anita has some thoughts on a footer design that we can leverage with this work as well)

themes/default/content/docs/get-started/install/_index.md
joeduffy: I can revert, always unfortunate when we make one thing sound like two, IMHO they're basically the same action from the user's perspective. But I'm only like 10% on this and happy to keep as is if you feel the old phrasing is better.

joeduffy:
Agree with you 100% on the suboptimal styling as well as next steps. My hope is we can land something fast here and then keep iterating. I often feel we wait for the perfect solution before making even small steps towards progress, and my hope is this is better than what we've got, and we can come up with something afterwards that's much better still.

By the way, if there's better styling we can apply right now to this change, I'd be all for it.  I tested it out with other fellow developers and overall feedback was positive, but the "business" and styling issues that you called out definitely came up.

themes/default/layouts/index.html
```diff
@@ -13,7 +13,7 @@ <h1 class="relative">
             <div class="overlay-middle mt-8 flex text-center justify-center items-center">
                 <div class="pb-12">
                     <a class="home-hero-btn-primary mr-4" href="{{ relref . "/docs/get-started" }}">{{ .Params.hero.cta_text }}</a>
-                    <a class="home-hero-btn-secondary ml-4" href="{{ relref . "/docs/get-started/install" }}">Download</a>
+                    <a class="btn-secondary ml-4 !px-12 !py-4" href="{{ relref . "/docs/get-started/install" }}">Download</a>
```
susanev, line 16:
```suggestion
                    <a class="btn-secondary !px-12 !py-4" href="{{ relref . "/docs/get-started/install" }}">Download</a>
```

susanev (approved):

PR #1816, opened 2022-08-11
fix buttons on home on smaller screens

susanev:
## before
before

## after
after


susanev:
@zchase @kimberleyamackenzie this works locally for me but that preview link, it goes to column but then back to row for super small screens so imma prolly need some of yalls help tomorrow if possible

lukehoban:
I don't see the same thing shown in the first screenshot myself - but I do see a slightly more minor issue that the page now has a horizontal scroll on mobile, seemingly because the download button overflows the viewport. Will the change here also address that?

image

image

susanev:
@lukehoban can you try the latest preview?

lukehoban:
> @lukehoban can you try the latest preview?

Yep. That looks great for me on mobile. No horizontal scroll.

zchase (approved):
THANK YOU!!!

PR #1825, opened 2022-08-12
Clarify Copy on Supported Py Version

RobbieMcKinstry:
This PR clarifies some unclear language around the minimum supported Python version.
As it was written, I can't imagine how we would support Python 3.x < 3.6.


susanev (approved):
thank you!!

PR #1838, opened 2022-08-17
Update troubleshooting.md

jkodroff:


themes/default/content/docs/support/troubleshooting.md
```diff
@@ -50,16 +50,13 @@ intentionally expose some known credentials to aid with debugging, so these log
...
-```
+```bash
 $ pulumi up --logtostderr --logflow -v=9 2> out.txt
 ```
 
-Individual resource providers may also have flags and environment entries to customize their
-diagnostic logging.  For example, for any Pulumi resource providers that expose a Terraform resource
-provider into Pulumi, you can use [`TF_LOG`](https://www.terraform.io/docs/internals/debugging.html)
-set to `TRACE`, `DEBUG`, `INFO`, `WARN` or `ERROR`.
+Resource providers may also have their own flags or environment variables that control diagnostic logging.  For example, for any Pulumi providers that use a bridged Terraform provider can make use of the [`TF_LOG`](https://www.terraform.io/docs/internals/debugging.html) environment variable (set to `TRACE`, `DEBUG`, `INFO`, `WARN` or `ERROR`) in order to provider additional diagnostic information.
```
susanev, line 57:
```suggestion
Diagnostic logging can also be controlled with flags and environment variables of the resource providers. For example, Pulumi providers that use a bridged Terraform provider can make use of the [`TF_LOG`](https://www.terraform.io/docs/internals/debugging.html) environment variable (set to `TRACE`, `DEBUG`, `INFO`, `WARN` or `ERROR`) in order to provide additional diagnostic information.
```

are there other situations than with the bridged providers? if not, suggest being even more specific and clear than im suggesting here.

susanev (approved):

themes/default/content/docs/support/troubleshooting.md
jkodroff: I think your description here is more accurate. Accepting.

PR #1839, opened 2022-08-17
add static website angular blog post

sean1588:
Fixes: https://github.com/pulumi/growth/issues/11


themes/default/content/blog/deploy-aws-static-website-with-angular/index.md
```diff
@@ -0,0 +1,116 @@
...
+We also use some components that come with the Angular framework.
+
+Each of the components has a TypeScript file that contains the programming logic and an HTML file that describes how the component will render.
+
+## Building the Pulumi Café Website
+
+Here are the instructions to build the website.
+
+1. Clone the GitHub repository located [here](https://github.com/sean1588/angular-static-site-pulumi).
+1. Run `npm install` to install the website’s node dependencies.
```
cnunciato, line 55: The `npm install` command fails for me with Node 16 or above, apparently because of something having to do with a transitive Python dependency somewhere:

```
npm ERR! gyp verb check python checking for Python executable "python" in the PATH
npm ERR! gyp verb `which` failed Error: not found: python
```

I do have `python`, but it's aliased to `python3` (apparently macOS removed the built-in `python` executable), so I'm not sure what the problem is exactly. Downgrading to Node 14 (with `nvm`) makes things work, but since 16 is active LTS and 18 is current, we should probably assume some number of folks will be a version greater than 14. Not sure what the best solution is, but figured I'd mention it as something to be aware of.

themes/default/content/blog/deploy-aws-static-website-with-angular/index.md
```diff
@@ -0,0 +1,116 @@
...
+1. Run `npm build` to generate the static website files and place them in a directory called **dist**. We will reference this directory when we configure Pulumi to deploy the site.
```
cnunciato, line 56: When I run this (with Node 14 now), I get:

```
npm WARN build `npm build` called with no arguments. Did you mean to `npm run-script build`?
```

`npm run build` does work, though.

themes/default/content/blog/deploy-aws-static-website-with-angular/index.md
```diff
@@ -0,0 +1,116 @@
...
+    mkdir pulumi && cd pulumi
+
+    ```
+
+1. You’re now in the **pulumi** directory. Generate a Pulumi project file and create a Pulumi.yaml file with this command:
+
+    ```bash
+    pulumi new aws-yaml
+    ```
+
```
cnunciato, line 74: We should probably mention that the project should be named `angular-static-website`, just to match what's in the example code. I'll usually say something along the lines of "follow the prompts" as well just so they know that's coming.

themes/default/content/blog/deploy-aws-static-website-with-angular/index.md
cnunciato: Could we also use `backticks` instead of **bold** for these literals? That'd be more consistent with how we normally quote them.

themes/default/content/blog/deploy-aws-static-website-with-angular/index.md
```diff
@@ -0,0 +1,116 @@
...
+## Modify the .yaml File
+
+We need to add a few lines of code to the Pulumi.yaml file. In the `resources` section, the code references the AWS Static Website component and sets the **sitePath** property to the location of the **dist** build directory. We also want to know the URL of the site once it’s deployed. That’s what the `outputs` section  does. The **websiteURL** parameter is one of the outputs that the AWS Static Website component provides. Here is an example screenshot of the .yaml file.
```
cnunciato, line 77: Little thing, but I'd probably instruct them to replace the contents of the file here:

```suggestion
We need to add a few lines of code to the Pulumi.yaml file. In the `resources` section, the code references the AWS Static Website component and sets the **sitePath** property to the location of the **dist** build directory. We also want to know the URL of the site once it’s deployed. That’s what the `outputs` section  does. The **websiteURL** parameter is one of the outputs that the AWS Static Website component provides. Replace the contents of `Pulumi.yaml` with the following code:
```

cnunciato (commented):
Nice! Just a few minor suggestions, other than the Node version thing, which we should probably figure out.

themes/default/content/blog/deploy-aws-static-website-with-angular/index.md
sean1588: ahh hmm... thanks for pointing that out!!  ok going to mess around with this a bit and see if I can find a way around it or see if I get rid of whatever dependency is relying on python. If all else fails, maybe I will add a note about the node version support. But going to try not to go that route as it might be awkward telling the user they need to use a old version of node that is a couple versions behind the current

themes/default/content/blog/deploy-aws-static-website-with-angular/index.md
sean1588: thank you! yeah totally meant `node run build`.

themes/default/content/blog/deploy-aws-static-website-with-angular/index.md
```diff
@@ -0,0 +1,116 @@
...
+    - yaml
+    - configuration
+    - components
+    - static-websites
+    - angular
+---
+
+In this blog post, we’re going to use some Angular framework components to assemble a static website and then use Pulumi and its AWS Static Website component to deploy it to AWS. The website is for a café called the Pulumi Café. It will contain two pages, one an About page and the other a Menu page, as well as some navigational pieces.
+
+To follow this example, you need to have both Angular and Pulumi installed. (Here’s a link to the [Pulumi installation instructions]({{< relref "/docs/get-started/aws/begin" >}}).) You’ll also need an AWS account.
```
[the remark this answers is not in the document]
sean1588: I'll go ahead and link them at the bottom of the document. We had decided to not include them here in order to try and keep the user's attention and not have them leave in the middle of the post.

themes/default/content/blog/deploy-aws-static-website-with-angular/index.md
sean1588: sounds good. I'll add a note there for that.

themes/default/content/blog/deploy-aws-static-website-with-angular/index.md
sean1588: I think I've spent way too much time at this point like 5 hours trying to convert the app over to a version of angular which would be compatible with the node 16. I'm just going to update the readme in the Github repo with the build instructions and say it is only compatible with node 14 as I don't think it is a good use of my time to keep messing around with this as I have already lost an entire day of work fighting with it. I think it is probably better if I just create a very basic angular app with this instead of having a side nav and needing to deal with material and all the dependencies that come with it, but I'll update the github repo and the screenshot in the blog post later when I do that. I think we should just ship this the way it is if that sounds good to you. @cnunciato

cnunciato: Sounds good @sean1588, thanks for the valiant effort!

sean1588: wow you are fast! lol. I actually just pushed the commit to take out the build command. so might need to wait a few mins for the preview link to generate.

cnunciato (approved):
Merge when ready. 🚀

PR #1843, opened 2022-08-18
Fix CLI binary download link for Windows

praneetloke:
h/t to @aureq for spotting this one while we were looking at something else.


praneetloke:
I noticed another problem on the same page that needs some investigation. The links under the [Other Installation Methods](https://www.pulumi.com/docs/get-started/install/#other-installation-methods) for Windows don't render correctly. If you know what the problem is, I am happy to edit this PR with the fix (or feel free to edit it yourself as well. I gave maintainers access to edit this PR.)

image

susanev (approved):
i think we should merge this to get this fixed and then move the rest of Praneet's feedback to an issue

PR #1850, opened 2022-08-19
log is a method in context

nagkumar:
context in java has a method log() not as public field


t0yv0 (approved):
You are right! Thanks a lot for sending the correction.

PR #1852, opened 2022-08-19
Explicitly set GCP Project via `pulumi config`

guineveresaenger:
We don't, at least in pulumi-gcp, use gcloud config, so we need to ask people to set Project the Pulumi way.

### DIFF
--- a/themes/default/layouts/shortcodes/configure-gcp.html
+++ b/themes/default/layouts/shortcodes/configure-gcp.html
@@ -1,9 +1,10 @@
 When developing locally, we recommend that you install the [Google Cloud SDK](https://cloud.google.com/sdk/install) and then [authorize access with a user
-account](https://cloud.google.com/sdk/docs/authorizing#authorizing_with_a_user_account). If `gcloud` is not configured to interact with your Google Cloud project, set it with the
-`config` command using the project's ID:
+account](https://cloud.google.com/sdk/docs/authorizing#authorizing_with_a_user_account). 
+
+Next, Pulumi requires default application credentials to interact with your Google Cloud resources, so run `auth application-default login` command to obtain those credentials:
 
 <div class="highlight">
-    <pre class="chroma"><code class="language-bash" data-lang="bash"><span class="no-select">$ </span>gcloud config set project your-gcp-project-id</code></pre>
+    <pre class="chroma"><code class="language-bash" data-lang="bash"><span class="no-select">$ </span>gcloud auth application-default login</code></pre>
     <div class="copy-button-container">
         <pulumi-tooltip
             ><span class="tooltip-target"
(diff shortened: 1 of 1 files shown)


themes/default/layouts/shortcodes/configure-gcp.html
```diff
@@ -13,10 +14,10 @@
     </div>
 </div>
 
-Next, Pulumi requires default application credentials to interact with your Google Cloud resources, so run `auth application-default login` command to obtain those credentials:
+To configure Pulumi to interact with your Google Cloud project, set it with the `config` command using the project's ID:
```
stack72, line 17: project, set it with the `pulumi config` command

themes/default/layouts/shortcodes/configure-gcp.html
```diff
@@ -13,10 +14,10 @@
...
 
 <div class="highlight">
-    <pre class="chroma"><code class="language-bash" data-lang="bash"><span class="no-select">$ </span>gcloud auth application-default login</code></pre>
+    <pre class="chroma"><code class="language-bash" data-lang="bash"><span class="no-select">$ </span>pulumi config set gcp:project your-gcp-project-id</code></pre>
```
stack72, line 20: We also support 3 environment variables so it doesn't need to be via config

stack72 (changes requested):

themes/default/layouts/shortcodes/configure-gcp.html
guineveresaenger: I go into this further below in this doc, but it's in this PR and waiting on publish:
https://github.com/pulumi/registry/pull/1369

guineveresaenger: Added a snippet for this

themes/default/layouts/shortcodes/configure-gcp.html
```diff
@@ -13,10 +13,23 @@
...
+    <div class="copy-button-container">
+        <pulumi-tooltip
+            ><span class="tooltip-target"
+                ><button class="copy-button"><i class="far fa-copy copy text-xl"></i></button></span
+            ><span class="tooltip-content" role="tooltip"><span slot="content">Copy</span></span></pulumi-tooltip
+        >
+    </div>
+</div>
+
+You may also set your GCP Project via environment variable (listed in order of precedence): - `GOOGLE_PROJECT` - `GOOGLE_CLOUD_PROJECT` - `GCLOUD_PROJECT` - `CLOUDSDK_CORE_PROJECT`
```
susanev, line 29: would it make sense to make this an actual list?

susanev (approved):

themes/default/layouts/shortcodes/configure-gcp.html
guineveresaenger: Yes! It would! `prettier` seems to have taken over here? This is weird!

guineveresaenger:
the preview looks good now! thank you for the catch, @susanev !

stack72 (approved):

PR #1855, opened 2022-08-20
fix headings and clean up troubleshooting page

susanev:
i dont think this page has been reviewed as a whole in a long time, as there were some heading level issues and misalignment between various sections. this pr is cleaning up the page to make it a bit more usable. feedback welcome!

edits have been made to `docs/support/troubleshooting`

**important note**: a number of things have been moved around, so reading the files changed on github is somewhat difficult. i recommend first reviewing the current page, then the preview, ahead of reviewing files changed.

table-of-contents

### DIFF
--- a/themes/default/content/docs/support/troubleshooting.md
+++ b/themes/default/content/docs/support/troubleshooting.md
@@ -243,8 +243,9 @@ introduced in Traefik 1.7.0.
 
 Asynchronous calls are the default in `@pulumi/pulumi>=2.0.0` and the below only applies to programs using the `1.x` SDK.
 
-The warning occurs when invoking a resource function synchronously while also using
-[an explicit provider object]({{< relref "/docs/intro/concepts/resources#providers" >}}) that isn't yet ready to use.
+The warning occurs when invoking a resource function synchronously while also using an
+[explicit provider object]({{< relref "/docs/intro/concepts/resources#providers" >}}) that isn't yet ready to use.
+
 For example:
 
 ```ts
@@ -259,12 +260,9 @@ const parent = new SomeResource("name", { provider });
 const ids = aws.ec2.getSubnetIds(..., { parent });
 ```
 
-This warning may be benign. However, if you are experiencing crashes or hangs in Pulumi (especially in Node.js version 12.11.0 and
-above) and you see this warning, then it is likely that this is the source.
+This warning may be benign. However, if you are experiencing crashes or hangs in Pulumi (especially in Node.js version 12.11.0 and above) and you see this warning, then it is likely that this is the source.
 
-A warning is issued so as to not break existing code that is functionality properly. However, the root cause of this problem
-pertains to undefined behavior in the Node.js runtime, so apparently-working code today may begin crashing or hanging tomorrow. As such,
-we recommend updating your code In a future version, Pulumi *may* be updated to throw instead of producing a warning when this happens.
+The root cause of this problem pertains to undefined behavior in the Node.js runtime,
 It is recommended that Pulumi apps be updated to prevent breakage.
 
 To address the issue update your app to use one of the following forms:
@@ -283,13 +281,12 @@ const ids = pulumi.output(aws.ec2.getSubnetIds(..., { parent }));
 ```
 
 This is the preferred way to solve this issue. In this form all resource function calls will always execute asynchronously,
-returning their result through a `Promise<...>`.  The result of the call is then wrapped into an `Output` so it can easily be
+returning their result through a `Promise<...>`. The result of the call is then wrapped into an `Output` so it can easily be
 passed as a resource input and to make it [simple to access properties]({{< relref "/docs/intro/concepts/inputs-outputs#lifting" >}}) off of it.
 
-If you do not want to change all calls to be `async` (perhaps because only one is encountering a problem), you can alternatively
-update only specific problematic calls to be asynchronous like so:
-
 #### Invoke the resource function asynchronously
+
+Use this method to update only specific problematic calls to be asynchronous.
 
 ```ts
 const ids = pulumi.output(aws.ec2.getSubnetIds(..., { provider, async: true })); // or
@@ -300,10 +297,9 @@ In this form, the `async: true` flag is passed in which forces `getSubnetIds` to
 of the call is then wrapped into an `Output` so it can easily be passed as a resource input and to make it
 [simple to access properties]({{< relref "/docs/intro/concepts/inputs-outputs#lifting" >}}) off of it.
 
-Sometimes, however, this approach is not possible because the call to the resource function happens a layer deeper (possibly in a
-component not under your control).  In that case, we recommend the solution in the next section:
-
 #### Register the provider first
+
+If the problem exists in a layer deeper (e.e.g, a component not under your control), use this solution.
 
 ```ts
 const provider = new aws.Provider(...);
@@ -334,13 +330,10 @@ const stackReference = new StackReference("...", { name: otherResource.outputVal
 const val = stackReference.getOutputSync("outputName");
 ```
 
-This warning may be benign. However, if you are experiencing crashes or hangs in Pulumi (especially in Node.js version 12.11.0 and
-above) and you see this warning, then it is likely that this is the source.
+This warning may be benign. However, if you are experiencing crashes or hangs in Pulumi (especially in Node.js version 12.11.0 and above) and you see this warning, then it is likely that this is the source.
 
-Currently, a warning is issued so as to not break existing code that is functionality properly. However, the root cause of this problem
-pertains to undefined behavior in the Node.js runtime, so apparently-working code today may begin crashing or hanging tomorrow. As such,
-we recommend updating your code. In a future version, Pulumi *may* be updated to throw instead of producing a warning when this happens.
-It is recommended that Pulumi apps be updated to prevent breakage.
+A warning is issued so as to not break existing code that is functionality properly. However, the root cause of this problem
+pertains to undefined behavior in the Node.js runtime, so apparently-working code today may begin crashing or hanging tomorrow. As such, we recommend updating your code. It is recommended that Pulumi apps be updated to prevent breakage.
 
 There are only two ways supported to avoid this issue:
 
@@ -402,10 +395,7 @@ can happen:
 - The CLI process is killed by your operating system while performing an update.
 - The CLI crashes when performing an update.
 
-This error means that the Pulumi engine initiated an operation but was not able to
-see if this operation was successful. Because of this, the Pulumi engine has no way of knowing
-whether or not the operations it initated completed successfully. This means that resources
-may have been created that Pulumi does not know about.
+This error means that the Pulumi engine initiated an operation and was not able to determine if the operation completed successfully. This means that resources may have been created that Pulumi does not know about.
 
 To fix this situation, you should first cancel the last update..
 
@@ -418,7 +408,7 @@ The currently running update for 'interruptedstack' has been canceled!
 If `pulumi cancel` fails with `error: [400] Bad Request: the update has already completed`, you can safely ignore
 that error and continue with the next step.
 
-You should then export and import your stack. This will clear your state's stack of all pending operations.
+You should then export and import your stack. This will clear the stacks's pending operations.
 
 ```bash
 $ pulumi stack export | pulumi stack import
@@ -449,8 +439,7 @@ At this point your stack should be valid, up-to-date, and ready to accept future
 ## Manually Editing Your Deployment {#editing-your-deployment}
 
 Sometimes the only recourse for fixing a stack that is unable to do deployments is to edit the
-deployment directly. It is possible to do this, though it is a tactic of last resort. It is a goal of Pulumi
-to never require users to edit their state directly. We would love to hear about the issues you are experiencing
+deployment directly. We would love to hear about the issues you are experiencing
 that you can't resolve, both so we can assist you in fixing your stack and also to fix the issues in Pulumi
 that made it impossible for you to recover your stack in any other way.
 
@@ -460,13 +449,12 @@ make changes to your stack is that the stack's existing state has been corrupted
 are a variety of ways that a stack's state could be corrupted, but in almost all cases it is possible
 to manually edit the stack's existing state to fix the corruption.
 
-Note that this is an advanced operation and should be an absolute last resort.
+This is an advanced operation and should be an absolute last resort. We recommend you check-in with the [Pulumi Community Slack](https://slack.pulumi.com) first before editing your snapshot.
 
-If you intend to unprotect or delete a resource, consider using the [`pulumi state`]({{< relref "/docs/reference/cli/pulumi_state" >}}) command to
-do so instead of editing your state directly. `pulumi state` also makes fixes to your state but without
+If you intend to unprotect or delete a resource, consider using the [`pulumi state`]({{< relref "/docs/reference/cli/pulumi_state" >}}) command instead of editing your state directly. `pulumi state` makes fixes to your state without
 requiring you to edit the JSON representation of your stack's current state.
 
-To get a JSON representation of your stack's current state, you can export your current stack
+To get a JSON representation of your stack's current state, export your current stack
 to a file:
 
 ```bash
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/support/troubleshooting.md
```diff
@@ -105,9 +96,39 @@ To view a trace locally navigate to the [Jaeger UI](http://localhost:16686/searc
...
+### 409 Conflict: Another update is currently in progress. {#conflict}
+
+Run `pulumi cancel` to cancel the update.
 
-### [409] Conflict: Another update is currently in progress. {#conflict}
+{{% notes type="warning" %}}
+Warning! If you cancel another person's update, their update will fail immediately.
+{{% /notes %}}
 
 One of the services that `pulumi.com` provides is *concurrency control*. The service will allow
```
MitchellGerdisch, line 133: I wonder if instead of saying "One of the services that `pulumi.com` provides ..." we should say something like 
"One of the services that the [Pulumi Service]({{< relref "/docs/intro/pulumi-service/" >}}) provides ..."

themes/default/content/docs/support/troubleshooting.md
```diff
@@ -11,68 +10,60 @@ aliases:
...
 
 Verbose logging of the internals of the Pulumi engine and resource providers can be enabled by
 passing the `-v` flag to any `pulumi` CLI command. Pulumi emits logs at log levels between `1` and
-`11`, with `11` being the most verbose.
+`11`, with `11` being the most verbose. At log level 10 or below, Pulumi will avoid intentionally exposing any *known* credentials. At log level 11, Pulumi will intentionally expose some known credentials to aid with debugging, so these log levels should be used only when absolutely needed.
 
 By default, logs are written to the top-level temp directory (usually `/tmp` or the value of
-`$TMPDIR`). The `--logtostderr` flag can be added to write logs to `stderr` instead, for easier
-access.  Use the flag `--logflow` to apply the same log level to resource providers.
+`$TMPDIR`). The `--logtostderr` flag can be used to write logs to `stderr` instead.
```
jkodroff, line 34:
```suggestion
`$TMPDIR`). The `--logtostderr` flag can be used to write logs to stderr instead.
```

Debatable, but I would not backtick it FWIW.

themes/default/content/docs/support/troubleshooting.md
```diff
@@ -11,68 +10,60 @@ aliases:
...
+Use the flag `--logflow` to apply the same log level to resource providers.
 
 {{% notes type="warning" %}}
 Enabling verbose logging may reveal sensitive information (tokens, credentials...) that is provided from
-your execution environment to your cloud provider and which Pulumi may not even itself be aware of. It is
-recommended that you audit the saved logs and redact any information before sharing the logs. At log level 10
-or below, Pulumi will avoid intentionally exposing any *known* credentials. At log level 11, Pulumi will
-intentionally expose some known credentials to aid with debugging, so these log levels should be used only
-when absolutely needed.
+your execution environment to your cloud provider and which Pulumi may not be aware of. Before sharing the logs, audit and redact the sensitive information.
```
jkodroff, line 39:
```suggestion
your execution environment directly to your cloud provider, and which Pulumi may not be aware of. Before sharing the logs, be careful to audit and redact any sensitive information.
```

themes/default/content/docs/support/troubleshooting.md
```diff
@@ -105,9 +96,39 @@ To view a trace locally navigate to the [Jaeger UI](http://localhost:16686/searc
 
 ## Common Problems
 
-This section covers a few problems that can arise when working with Pulumi.
+### 403 error fetching plugin
+
+This error commonly occurs when using an arm64 based processor while using an older version of a provider that does not support arm64. It's not possible to upgrade the packages as your state will still be locked to the old version of the provider.
```
jkodroff, line 101:
```suggestion
This error commonly occurs when using an ARM64-based processor while using an older version of a provider that does not support ARM64. It is not possible to upgrade the providers as your state file requires the version of the provider used at the time the resources were created.
```

(At least I think that's the implication...)

themes/default/content/docs/support/troubleshooting.md
```diff
@@ -105,9 +96,39 @@ To view a trace locally navigate to the [Jaeger UI](http://localhost:16686/searc
...
+
+The method for fiing this depends on whether you are using an Intel based computer.
```
jkodroff, line 103:
```suggestion
The method for fixing this issue depends on whether you are using an Intel based processor:
```

themes/default/content/docs/support/troubleshooting.md
```diff
@@ -105,9 +96,39 @@ To view a trace locally navigate to the [Jaeger UI](http://localhost:16686/searc
...
+
+#### Intel based computer
```
jkodroff, line 105:
```suggestion
#### Intel based processor
```

themes/default/content/docs/support/troubleshooting.md
```diff
@@ -105,9 +96,39 @@ To view a trace locally navigate to the [Jaeger UI](http://localhost:16686/searc
...
+
+1. Open your Pulumi program on a non-arm64 based computer.
+1. Update your packages (pip / nuget / npm / go) and run `pulumi up`.
+1. Once the update is complete, open the new, updated Pulumi program on your arm64-based system.
+
+#### Non-Intel based computer
```
jkodroff, line 111:
```suggestion
#### Non-Intel based processor
```

themes/default/content/docs/support/troubleshooting.md
```diff
@@ -105,9 +96,39 @@ To view a trace locally navigate to the [Jaeger UI](http://localhost:16686/searc
...
+
+1. Remove Pulumi - if you're using Homebrew, `brew remove pulumi` or `rm -rf ~/.pulumi`
```
jkodroff, line 113: Aren't these separate things? `brew remove pulumi` would remove the Pulumi executable (`/opt/homebrew/bin/pulumi`), but `rm -rf ~/.pulumi` would remove plugins, templates, etc.

themes/default/content/docs/support/troubleshooting.md
```diff
@@ -520,43 +362,140 @@ value defined elsewhere in your application. If the value is known, it can be co
 If the stack-reference-name truly is dynamic and cannot be known ahead of time to supply directly into the app, then this
 approach will not work, and the only way to workaround the issue is to follow the steps in [Use getOutput/requireOutput](#use-getoutput).
 
-## 403 HTTP error fetching plugin
+### Cannot connect to the Pulumi Service
```
MitchellGerdisch, line 365: I think this section "Cannot connect to the Pulumi Service" should be moved up to follow (or maybe precede) the "Nothing happens due to network proxy" since they are both kind of networking related. It also puts the two topics related to pre-Pulumi 2.0 issues at the end.

themes/default/content/docs/support/troubleshooting.md
```diff
@@ -120,30 +141,28 @@ a lease to the stack that you are attempting to update. There are two reasons wh
 1. Somebody else is currently updating the stack. If you are working on a stack with more than one collaborator, it could
 be that your collaborators have initiated an update without your knowledge. You can confirm this by visiting the Pulumi
 web console and seeing who initiated the most recent update.
-2. You were just updating the stack, but the Pulumi command-line tool crashed in the middle of the update.
+
+2. You were updating the stack, but the Pulumi CLI crashed in the middle of the update.
```
jkodroff, line 145:
```suggestion
1. You were updating the stack, but the Pulumi CLI crashed in the middle of the update.
```

MitchellGerdisch (commented):
Nice work! This update makes the page cleaner and easier to parse.

themes/default/content/docs/support/troubleshooting.md
```diff
@@ -406,17 +256,14 @@ const parent = new SomeResource("name", { provider });
 const ids = aws.ec2.getSubnetIds(..., { parent });
 ```
 
-This warning may be benign. However, if you are experiencing crashes or hangs in Pulumi (especially in Node.js version 12.11.0 and
-above) and you see this warning, then it is likely that this is the source.
+This warning may be benign. However, if you are experiencing crashes or hangs in Pulumi (especially in Node.js version 12.11.0 and above) and you see this warning, then it is likely that this is the source.
```
jkodroff, line 259:
```suggestion
This warning may be benign. However, if you are experiencing crashes or hangs in Pulumi (especially in Node.js version 12.11.0 and above) and you see this warning, then it is likely that the condition described above is the root cause.
```

themes/default/content/docs/support/troubleshooting.md
```diff
@@ -447,10 +293,9 @@ In this form, the `async: true` flag is passed in which forces `getSubnetIds` to
 of the call is then wrapped into an `Output` so it can easily be passed as a resource input and to make it
 [simple to access properties]({{< relref "/docs/intro/concepts/inputs-outputs#lifting" >}}) off of it.
 
-Sometimes, however, this approach is not possible because the call to the resource function happens a layer deeper (possibly in a
-component not under your control).  In that case, we recommend the solution in the next section:
+#### Register the provider first
 
-### Register the provider first
+If the problem exists in a layer deeper (e.e.g, a component not under your control), use this solution.
```
jkodroff, line 298:
```suggestion
If the problem exists in a layer deeper (e.g, a component not under your control), use this solution.
```

themes/default/content/docs/support/troubleshooting.md
```diff
@@ -481,17 +326,14 @@ const stackReference = new StackReference("...", { name: otherResource.outputVal
 const val = stackReference.getOutputSync("outputName");
 ```
 
-This warning may be benign. However, if you are experiencing crashes or hangs in Pulumi (especially in Node.js version 12.11.0 and
-above) and you see this warning, then it is likely that this is the source.
+This warning may be benign. However, if you are experiencing crashes or hangs in Pulumi (especially in Node.js version 12.11.0 and above) and you see this warning, then it is likely that this is the source.
```
jkodroff, line 329: Same suggestion as above.

themes/default/content/docs/support/troubleshooting.md
```diff
@@ -520,43 +362,140 @@ value defined elsewhere in your application. If the value is known, it can be co
...
+```bash
+$ pulumi stack export | pulumi stack import
+warning: removing pending operation 'creating' on '...' from snapshot
+Import successful.
+```
 
-* The URL that the Pulumi CLI uses to connect to the Service is `https://api.pulumi.com`. (It does not use `https://app.pulumi.com`, so if you want to view the console, you'll need enable that as well.)
-* All access goes over HTTPS via port 443.
+For every warning that this command prints out, you should verify with your cloud provider whether or not this
+operation was successful. If the operation was successful, and a resource was created, you should delete that
```
jkodroff, line 416:
```suggestion
operation was successful. If the operation was successful, and a resource was created, you should delete the
```

themes/default/content/docs/support/troubleshooting.md
```diff
@@ -520,43 +362,140 @@ value defined elsewhere in your application. If the value is known, it can be co
...
+Sometimes the only recourse for fixing a stack that is unable to do deployments is to edit the
+deployment directly. We would love to hear about the issues you are experiencing
+that you can't resolve, both so we can assist you in fixing your stack and also to fix the issues in Pulumi
+that made it impossible for you to recover your stack in any other way.
+
+The Pulumi engine uses both your program and your stack's existing state to make decisions about what
+resources to create, read, update, or delete. The most common problem that makes it impossible to
+make changes to your stack is that the stack's existing state has been corrupted in some way. There
+are a variety of ways that a stack's state could be corrupted, but in almost all cases it is possible
+to manually edit the stack's existing state to fix the corruption.
```
jkodroff, line 446:
```suggestion
to manually edit the stack's state to fix the issue.
```

themes/default/content/docs/support/troubleshooting.md
```diff
@@ -520,43 +362,140 @@ value defined elsewhere in your application. If the value is known, it can be co
...
+
+This is an advanced operation and should be an absolute last resort. We recommend you check-in with the [Pulumi Community Slack](https://slack.pulumi.com) first before editing your snapshot.
```
jkodroff, line 448:
```suggestion
This is an advanced operation and should be an absolute last resort. We recommend you check-in with the [Pulumi Community Slack](https://slack.pulumi.com) first before editing your state file.
```

themes/default/content/docs/support/troubleshooting.md
```diff
@@ -520,43 +362,140 @@ value defined elsewhere in your application. If the value is known, it can be co
...
+requiring you to edit the JSON representation of your stack's current state.
+
+To get a JSON representation of your stack's current state, export your current stack
+to a file:
+
+```bash
+$ pulumi stack export --file state.json
+```
+
+This file contains a lot of information. At the top-level, this JSON object has two fields:
```
jkodroff, line 460: A brief search did not reveal an obvious source for detailed info on the Pulumi state file, but if there is one, we should link to it in this section.

jkodroff (changes requested):
Well that was epic, and is a _major_ improvement!

Since some suggestions are typos, requesting changes, but I don't expect all grammatical or technical suggestions to be accepted - author's perogative.

jkodroff (approved):

themes/default/content/docs/support/troubleshooting.md
susanev: @jkodroff we should avoid directional words, because they are not accessible. i can link to the previous section though if you'd like.

themes/default/content/docs/support/troubleshooting.md
susanev: @jkodroff i recently learned that snapshot is slightly different than state file, so would like to leave this as is. but i will be adding snapshot to the glossary soon. also, snapshot is mentioned a bunch more times in this file.

themes/default/content/docs/support/troubleshooting.md
susanev: sure will do! was feeling a bit funny about this being at the bottom anyway.

themes/default/content/docs/support/troubleshooting.md
jkodroff: @susanev This is helpful information. If you have a source handy, can you send me some guidelines on how to refer to previous information that refers to e.g. the preceding paragraph?

The main thing I was trying to address was the use of "this" and "that" with no surrounding context. I _think_ they're called "unqualified articles", and I was taught that they should be avoided in technical writing, but a Google search only turns up how journalists shouldn't write about science, which is obviously not what we're talking about here.

MitchellGerdisch (approved):

themes/default/content/docs/support/troubleshooting.md
susanev: @jkodroff i have no idea what _above_ means here? sry can you be more specific?

themes/default/content/docs/support/troubleshooting.md
susanev: i also could not find one

themes/default/content/docs/support/troubleshooting.md
susanev: we should link to the section we are referencing. i think i put this in the style guide. but maybe im making that up.
some sources that follow this model.
- https://styleguide.mailchimp.com/writing-for-accessibility/#header-3-avoid-directional-language
- https://uxdesign.cc/ux-copy-for-accessibility-2927491cadf8 "→ Avoid using directional language.
Above, below, left, right… Far more often than you might think, these words are meaningless and misinterpreted by the users (and not just those who are blind or visually impaired). Users with screen readers won’t have right or left as a reference."
- https://developers.google.com/style/accessibility "Don't use directional language to orient the reader, such as above, below, or right-hand side. This type of language doesn't work well for accessibility or for localization reasons."

susanev: ive clarified the language and removed the last reference to `that`. let me know if it makes sense.

susanev: i think imma go ahead and merge tonight, happy to come back with a separate pr if i got it wrong.

PR #1856, opened 2022-08-22
Change pull-request.yaml to run on ubuntu-latest

jkodroff:
Ubuntu 18 had a scheduled brownout that caused the workflow not to run. I've never experienced this with other workflows (all of which use `ubuntu-latest`). If we're intentionally using a pinned version of Ubuntu, please refuse this PR, let me know why, and I'll submit another PR to comment as to why we're using v18 and under what conditions we can upgrade.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -11,7 +11,7 @@ jobs:
     env:
       GOPATH: ${{ github.workspace }}/go
     name: Build and deploy preview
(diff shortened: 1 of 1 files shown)


cnunciato (approved):
Should be fine, yep. Thanks!

cnunciato:
@jkodroff Looks like we're still on 18.04 in /docs and /registry also. I'll get those updated as well. Thanks for the nudge.

PR #1865, opened 2022-08-24
Add Fn::Split to YAML Reference

aq17:
Add `Fn::Split` to YAML docs


cnunciato (approved):
🙌 Thanks!

BTW, are you using a formatter for those Markdown tables? If so, which?

aq17:
> 🙌 Thanks!
> 
> BTW, are you using a formatter for those Markdown tables? If so, which?

I have VSCode's `Markdown All in One` extension installed

PR #1867, opened 2022-08-25
Update get.md

jkodroff:


themes/default/content/docs/intro/concepts/resources/get.md
```diff
@@ -7,7 +7,12 @@ menu:
     weight: 7
 ---
 
-You can use the static `get` function, which is available on all resource types, to look up an existing resource’s ID. The `get` function is different from the `import` function. The difference is that, although the resulting resource object’s state will match the live state from an existing environment, the resource will not be managed by Pulumi. A resource read with the `get` function will never be updated or deleted by Pulumi during an update.
+You can use the static `get` function, which is available on all resource types, to look up an existing resource that is not managed by Pulumi. The `get` function is different from [the `import` CLI command]({{< ref "/docs/reference/cli/pulumi_import" >}}): `pulumi import` is used bring an existing resource under management by Pulumi. `get` is used to allow the attributes of an existing resource to be used within a Pulumi program. A resource read with the `get` function will never be updated or deleted by Pulumi during an update.
```
susanev, line 10:
```suggestion
You can use the static `get` function, which is available on all resource types, to look up an existing resource that is not managed by Pulumi. The `get` function is different from the [`import` CLI command]({{< ref "/docs/reference/cli/pulumi_import" >}}): `pulumi import` is used to bring an existing resource under management by Pulumi. `get` is used to allow the attributes of an existing resource to be used within a Pulumi program. A resource read with the `get` function will never be updated or deleted by Pulumi during an update.
```

themes/default/content/docs/intro/concepts/resources/get.md
```diff
@@ -7,7 +7,12 @@ menu:
...
+
+Two values are passed to the `get` function:
+
+1. The logical name Pulumi will use to refer to the resource.
+1. The physical ID that the resource has in the target cloud.
```
susanev, line 15:
```suggestion
Two values are passed to the `get` function:

1. Logical name Pulumi will use to refer to the resource.
1. Physical ID that the resource has in the target cloud.
```

dropping the `the`'s will make this a bit easier to scan

susanev (approved):

themes/default/content/docs/intro/concepts/resources/get.md
jkodroff: Agreed, but I think it reads worse for non-scanners without "the". Would it be within our style guidelines if we bolded them?

```markdown
1. The **logical name** Pulumi will use to refer to the resource.
1. The **physical ID** that the resource has in the target cloud.
```

jkodroff:
@susanev Having trouble verifying that my link (`relref`) to the CLI is correct. It looks like all the CLI links 404 in the generated site, even when clicking the links in the LH nav.

Please advise.

susanev:
@jkodroff the cli links 404 in previews because the cli docs do not get built until it gets to docs. i think there's directions in a readme somewhere about how to generate the cli docs locally if you want to try that, but let me take a look at the link.

susanev:
link looks good to me, but i honestly mess up links a lot so you might wanna have someone else verify as well.

jkodroff:
Verified the link is correct after deployment.

Issue #1506, opened 2022-05-13
Learning page update - typo in 'making a stack configurable'


pulumi-TH:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

### Affected area/feature

On the learning page (https://www.pulumi.com/learn/building-with-pulumi/stack-outputs/
) in the section “Making a stack configurable” the `pulumi config set` is wrong. The variables are each indicated with “_”, but in the code - or the previous lesson - the variables are camelCase.
This causes the `pulumi up` to fail with reference to missing variable definitions.

susanev:
@nimbinatus is this something you can take a look at?

nimbinatus:
Yep, this is something I can fix. We must have missed it in the PulumiUp updates

PR #1869, opened 2022-08-25
fix(var): fix all variables for typescript; remove seed container ref

nimbinatus:
fixes https://github.com/pulumi/pulumi-hugo/issues/1506


katcosgrove (approved):
lgtm!

PR #1870, opened 2022-08-25
Update state & login instructions for AWS profiles

AaronFriel:
Fixes https://github.com/pulumi/pulumi/issues/9555

### DIFF
--- a/themes/default/content/docs/intro/concepts/state.md
+++ b/themes/default/content/docs/intro/concepts/state.md
@@ -167,7 +167,7 @@ $ pulumi login s3://<bucket-name>
 ```
 
 {{% notes type="info"%}}
-As of Pulumi CLI v3.33.1, instead of specifying the AWS Profile using the `AWS_PROFILE` environment variable, add `region=us-east-1` (or any other region), `awssdk=v2` and `profile=` followed by the profile name to the query string. On Linux and macOS, this will need to be quoted to escape the '&' operators in the shell as below:
+As of Pulumi CLI v3.33.1, instead of specifying the AWS Profile, add `awssdk=v2` along with the region and profile to the query string. The URL should be quoted to escape the shell operator `&`, and used as follows:
 
 ```sh
 pulumi login 's3://<bucket-name>?region=us-east-1&awssdk=v2&profile=<profile-name>'
--- a/themes/default/content/docs/intro/concepts/state.md
+++ b/themes/default/content/docs/intro/concepts/state.md
@@ -167,7 +167,11 @@ $ pulumi login s3://<bucket-name>
 ```
 
 {{% notes type="info"%}}
-As of Pulumi CLI v3.33.1, it's possible to specify AWS profile from the shared config file `~/.aws/config` in the URL query parameter, i.e. `s3://<bucket-name>?profile=main`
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/intro/concepts/state.md
```diff
@@ -167,7 +167,12 @@ $ pulumi login s3://<bucket-name>
 ```
 
 {{% notes type="info"%}}
-As of Pulumi CLI v3.33.1, it's possible to specify AWS profile from the shared config file `~/.aws/config` in the URL query parameter, i.e. `s3://<bucket-name>?profile=main`
+As of Pulumi CLI v3.33.1, instead of specifying the AWS Profile using the `AWS_PROFILE` environment variable, add `region=us-east-1` (or any other region), `awssdk=v2` and `profile=` followed by the profile name to the query string. On Linux and macOS, this will need to be quoted to escape the '&' operators in the shell as below:
```
Frassle, line 170: Does gocloud have any docs about the params supported here? Can we point to them?

susanev: this is a lot to read all at once, do we need to have the content in backticks? or can use say profile, region, version, etc?

susanev (approved):

themes/default/content/docs/intro/concepts/state.md
AaronFriel: @susanev Depends on what you mean by "need"? The CLI expects verbatim query parameters. If a user typos, they see an error message from the underlying library:

```
error: problem logging in: unable to open bucket s3://my-bucket-c989529?regio=us-east-1&awssd=v2&prof=qa: open bucket s3://my-bucket-c989529?awssd=v2&prof=qa&regio=us-east-1: unknown query parameter "awssd"
```

The second sentence guards against user error, without quotes the shell will run 3 commands and our program only gets partial output:

```
$ pulumi login s3://my-bucket-c989529?region=us-east-1&awssdk=v2&profile=qa
[1] 30587
[2] 30588
[2]+  Done                    awssdk=v2

$ error: problem logging in: unable to check if bucket s3://my-bucket-c989529?region=us-east-1 is accessible: blob (code=Unknown): InvalidToken: The provided token is malformed or otherwise invalid.
        status code: 400, request id: [redacted], host id: [redacted]

[1]+  Exit 255                pulumi login s3://my-bucket-c989529?region=us-east-1
```

susanev: sry i was not clear, i meant the sentence where you you list out examples in backticks, `AWS_PROFILE`, `regision=us-east-1`, it makes that sentence difficult to read. what about...

```suggestion
As of Pulumi CLI v3.33.1, instead of specifying the AWS Profile add the region, profile, and profile name to the query string. On Linux and macOS, this will need to be quoted to escape the '&' operators in the shell as below:
```

and since you have an example right below it i don't think you need the backticked example in that sentence. i don't feel strongly though, so feel free to go forward as you have it if you'd like.

AaronFriel: I'll change that slightly to call out `awssdk=v2` in backticks as it must be used verbatim, and the others by name.

AaronFriel:
@susanev thanks for your notes!

PR #1885, opened 2022-08-31
Add options to Fn::Invoke docs

aq17:
Updates YAML reference for `Fn::Invoke` based on [change here](https://github.com/pulumi/pulumi-yaml/pull/284)

### DIFF
--- a/themes/default/content/docs/reference/yaml/_index.md
+++ b/themes/default/content/docs/reference/yaml/_index.md
@@ -287,6 +287,7 @@ Calls a function from a package and returns either the whole object or a single 
 | ------------- |---|-------------| -----|---|
 | `Function`    | string | Yes | No | Name of a function to call. |
 | `Arguments`   | map[string]Expression | Yes | Yes | Arguments to pass to the expression, each key is a named argument. |
+| `Options`   	| [Invoke Options](#invoke-options) | No | No | Options for the provider calling the function. |
 | `Return`      | string | No | No | If the function returns an object, a single key may be selected and returned instead with its name. |
 
 ```yaml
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/reference/yaml/_index.md
```diff
@@ -300,11 +301,32 @@ variables:
...
       Return: id
 ```
 
 The expression `${AmazonLinuxAmi}` will return the AMI ID returned from the [`aws:getAmi`](https://www.pulumi.com/registry/packages/aws/api-docs/getami/) function.
 
+#### Invoke Options
+
+The value of the `options` property of an Fn::Invoke is an object with the following properties.
+
+The  `parent` and`provider` values permit expressions which must use interpolation syntax to reference resources by name. For example:
```
AaronFriel, line 315:
```suggestion
The  `parent` and `provider` values permit expressions which must use interpolation syntax to reference resources by name. For example:
```

themes/default/content/docs/reference/yaml/_index.md
```diff
@@ -300,11 +301,32 @@ variables:
             values: ["amzn-ami-hvm-*-x86_64-ebs"]
         owners: ["137112412989"]
         mostRecent: true
+      Options:
+		Version: 5.9.0
```
AaronFriel, line 305:
```suggestion
        Version: 5.9.0
```

themes/default/content/docs/reference/yaml/_index.md
```diff
@@ -300,11 +301,32 @@ variables:
...
+```yaml
+    options:
+      provider: ${myEksProvider}
+      parent: ${someParentResource}
+```
+
+| Property                  | Type         | Description |
+| ------------------------- |--------------|-------------|
+| `parent`                  | Expression   | No | Yes | Parent specifies a parent for the resource |
+| `provider`                | Expression   | No | Yes | Provider specifies an explicitly configured provider, instead of using the default global provider |
```
AaronFriel, line 326:
```suggestion
| `parent`                  | Expression   | Parent specifies a parent for the resource |
| `provider`                | Expression   | Provider specifies an explicitly configured provider, instead of using the default global provider |
```

AaronFriel (approved):
Looks good to me with a few suggestions

PR #1892, opened 2022-09-01
Add resources under management APIs to REST docs

mchaynes:
Add documentation for our RUM APIs to https://pulumi.com/docs/reference/service-rest-api/#resources-under-management-rum.

Specifically looking for feedback on how I formatted these different responses. I added an example for each `granularity` since the response shape is slightly different depending on which granularity you choose. For example, `granularity=weekly` adds a `"weekNumber"` key to the response object. There might be a better way to represent this, very open to other ideas! 

Related To: https://github.com/pulumi/service-requests/issues/87

### DIFF
--- a/themes/default/content/docs/reference/service-rest-api/_index.md
+++ b/themes/default/content/docs/reference/service-rest-api/_index.md
@@ -2146,3 +2146,164 @@ Timestamp,Name,Login,Event,Description,SourceIP,RequireOrgAdmin,RequireStackAdmi
 2021-04-11T23:09:25Z,First Last,user1,Member Role Changed,"Changed organization role for ""user3"" to admin",192.168.10.11,true,false,false
 2021-04-11T21:09:52Z,First Last,user1,Secret Decrypted,"Decrypted secret value for stack ""demo-aws-ts-webserver/dev-user1"" (cipher text suffix: ""tbpiX4c="")",192.168.10.11,false,false,false
 ```
+
+## Resources Under Management (RUM)
+
+<!-- ###################################################################### -->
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/reference/service-rest-api/_index.md
```diff
@@ -2146,3 +2146,164 @@ Timestamp,Name,Login,Event,Description,SourceIP,RequireOrgAdmin,RequireStackAdmi
...
+
+### Get Resource Count Summary
+
+```http
+GET /orgs/{organization}/resources/summary?granularity=<granularity>&lookbackDays=<days>
+```
+
+#### Example (hourly)
+
+{{% notes "info" %}}
```
caseyyh, line 2162: Oh what does this do? It's not rendering right on the github rich diff :(

caseyyh (commented):
>  I added an example for each granularity since the response shape is slightly different depending on which granularity you choose. 

I'm not sure this is strictly necessary - a CURL is fairly easy to test out, and as long as there's some callout that the response shape might be different, I can probably just run a test example on my own terminal if I'm a developer looking to use this endpoint. (In fact, I usually do that rather than trust documentation, because we all know how fabulously up-to-date docs tend to be.) -- just my two cents.

As an alternate suggestion, I see elsewhere that we have a `Parameters` table for input -- maybe here we can have an `Output Parameters` table that indicates whether a parameter will be included in the response object for XYZ granularity. It might be more concise than outlining all the different examples, just a thought!

EvanBoyle (approved):

PR #1893, opened 2022-09-01
Update home hero and ctas

zchase:
This PR contains changes to the home page hero as described in: https://github.com/pulumi/marketing/issues/416

Tied to this theme PR: https://github.com/pulumi/theme/pull/200

NOTE: I will resolve the merge conflict in `go.mod` before emerging but need it now to have a "correct" preview because of the dependent theme PR.

### DIFF
--- a/themes/default/content/_index.md
+++ b/themes/default/content/_index.md
@@ -3,11 +3,10 @@ title: Pulumi - Universal Infrastructure as Code
 meta_desc: Pulumi's open source infrastructure as code SDK enables you to create, deploy, and manage infrastructure on any cloud, using your favorite languages.
 
 hero:
-  title: [ "Universal", "Infrastructure as Code" ]
+  title: [ "Ship", "Infrastructure", "Faster" ]
   description: |
-    Define infrastructure on any cloud using familiar programming languages.
-    For developers, infrastructure teams, and everybody in between.
-  cta_text: Get Started
+    Deliver infrastructure with high velocity and scale through software engineering.
+  cta_text: Start Free
 
 overview:
   title: [ "Your cloud.", "Your language.", "Your way." ]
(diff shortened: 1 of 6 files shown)


themes/default/content/_index.md
```diff
@@ -3,11 +3,10 @@ title: Pulumi - Universal Infrastructure as Code
...
+  title: [ "Ship", "Infrastructure", "Faster" ]
   description: |
-    Define infrastructure on any cloud using familiar programming languages.
-    For developers, infrastructure teams, and everybody in between.
-  cta_text: Get Started
+    Deliver infrastructure with high velocity and scale through software engineering.
+  cta_text: Start Free
 
 overview:
   title: [ "Your cloud.", "Your language.", "Your way." ]
```
aaronkao, line 12: title: [ "", "Universal", "Infrastructure as Code" ]

aaronkao (approved):

themes/default/content/_index.md
aaronkao, line 12: title: [ "", "Universal", "Infrastructure as Code" ]

susanev:
@zchase @aaronkao it may be worth considering all the other places we use the "Get Stared" language throughout the marketing, docs, and registry content. as only changing the label here is going to create misalignment with the rest of the content. i don't think yall should block as from what i understand this is an experiment, but if it ends up being successful we should go back and align the rest of the language. might be worth having an issue tracking that work so that we don't forget about it.

cnunciato (commented):
FWIW, with the title-casing and left-justification, I read as three separate words -- SHIP. INFRASTRUCTURE. FASTER. -- than as the phrase that's probably intended. Something more like this might be worth considering:

PR #1895, opened 2022-09-01
Fix Pulumi Challenge CTA image

cnunciato:
The social/meta image for https://www.pulumi.com/challenge/startup-in-a-box/  isn't isn't getting picked up at the moment because it resolves to a 404:

image

The reason the specified (i.e., page-relative) location of the image doesn't work is because we currently only support [Hugo Page resources](https://gohugo.io/content-management/page-resources/) in the blog; all others are presumed to exist under `static`, so require root-relative paths. So this change moves the image into `static` (alongside the other Challenge-related image) and updates the content-file reference to match.


cnunciato:
To verify this works:

image

http://pulumi-hugo-origin-pr-1895-b49ad405.s3-website.us-west-2.amazonaws.com/images/challenge/challenge_cta.png

zchase (approved):

PR #1899, opened 2022-09-02
Update the documenation for what backend is

Frassle:
Looking at adding a JSON schema file for this project format as well, see https://github.com/pulumi/pulumi/blob/2db79a54cfc01edbfea51f28324bf96e437cf630/sdk/go/common/workspace/project.json.

Maybe we could start auto-gening this page at some point from that schema?


susanev (commented):
thank you!! 🎉

susanev:
>Maybe we could start auto-gening this page at some point from that schema?

yes pls!!

susanev (approved):

PR #1902, opened 2022-09-02
move dev docs to folder

susanev:
we have 3 dev docs files now, so thought it may help things if we started a folder
i wasn't sure on the folder name, would `dev-docs` be better?


cnunciato (commented):
I think I'd choose `docs` over `dev-docs` (as these generally aren't dev-related), but at the same time, something tells me putting "docs" at the top level like this might look to some folks like that's where our docs-related content lives, and confuse them. So instead I might suggest something more general, like `guide` or `contributing`. Looks like GitHub uses `contributing` for its version of this, which it links to from CONTRIBUTING.md: https://github.com/github/docs. Up for that?

susanev:
will maybe come back to this in some made up future

PR #1907, opened 2022-09-05
fix pulumi service docs nav order

susanev:
## overview
we are starting to collect some ux debt on the service docs, this attempts to organize things in a logical order that our users may be able to better understand.

i started this convo in [internal slack](https://pulumi.slack.com/archives/C8A7QG1CZ/p1661217840351789)

i will mention that i have usability concerns that we have token information spread out across 3 separate docs pages, and that we have role information spread out across 3 separate docs pages as well. this requires folks to read all these pages to understand how these things work together. for now, though im trying to do the smallest change to reduce some usability issues.

super open to feedback on the order!!

## before
before-nav

## after
after-nav


kimberleyamackenzie (approved):
:shipit:

komalali (approved):

PR #1908, opened 2022-09-06
fix misleading template config mention

### LINKED ISSUE #1906: template options docs content is misleading
convo from https://github.com/pulumi/pulumi/issues/10581#issuecomment-1237309264

should fix up this content until the follow on work is complete

susanev:
fixes #1906 

the config option for templates only asks for that config when this template is used as part of `pulumi new`. it does not auto-apply this config for all stacks within a project.

I've tried a pass at more clear words, also open to removing this from the docs. not sure whats best.

this came from a issue report over in pulumi/pulumi: https://github.com/pulumi/pulumi/issues/10581


susanev:
@gsuess also interested in ur feedback on this!

PR #1909, opened 2022-09-06
fix(aws): bucket param input type

andridns:
parameter `bucket` type should be `str`. I believe `bucket.id` is more appropriate here than `bucket` which is of `Bucket` type


viveklak (approved):

viveklak:
> parameter `bucket` type should be `str`. I believe `bucket.id` is more appropriate here than `bucket` which is of `Bucket` type

Thanks for the PR! Could you also make the same change to the other languages?

andridns:
Thanks @viveklak . I've pushed the same fix to more AWS docs, and some to GCP - I'm not sure about other languages as I'm more familiar with Python

viveklak (approved):

cnunciato (approved):
Thanks @andridns!

PR #1921, opened 2022-09-08
Fix a syntax issue in the resource options docs

cnunciato:
When I use the syntax currently quoted in the docs, I get an error:

```
error: Error registering resource [record]: Pulumi.yaml:119,18-32: resource option dependsOn value must be a list of resources;
```

This seems to be what's needed.

### DIFF
--- a/themes/default/content/docs/intro/concepts/resources/options/dependson.md
+++ b/themes/default/content/docs/intro/concepts/resources/options/dependson.md
@@ -78,7 +78,8 @@ resources:
   res2:
     type: MyResource
     options:
-      dependsOn: ${res1}
+      dependsOn: 
+        - ${res1}
 ```
 
 {{% /choosable %}}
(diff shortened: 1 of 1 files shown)


justinvp (approved):
LGTM. I assume specifying the list like that is preferred over doing it in one line like: `dependsOn: [${res1}]` (I haven't tried this, so don't know if it works).

cnunciato:
@justinvp Good question -- I wasn't sure, so I tried it, and it doesn't seem to work (at least as of now):

```
error: could not get cloud url: could not load current project: \
    invalid YAML file: yaml: line 126: did not find expected ',' or ']'
```

I do see this array notation elsewhere in the YAML docs, so I'm assuming it's okay.

justinvp:
Thanks for confirming, @cnunciato, and for the doc fixes!

PR #1922, opened 2022-09-09
Add nodejs binaries pkg blog post

danielrbradley:


themes/default/content/blog/nodejs-binaries-with-pkg/index.md
```diff
@@ -0,0 +1,131 @@
...
+- nodejs
+- publishing
+- providers
+
+meta_image: "meta.png"
+---
+
+It's well known that you can write Pulumi programs in loads of languages, but what about the providers behind the scenes which handle communication with various cloud services?
+
+Providers are downloaded and run in many different platforms (e.g. Linux, macOS, Windows) and environments – where we can’t assume an end-user is going to have specific runtimes installed – including Node.js. Most of our providers are authored in Go – which makes it simple to produce native, static binaries for all modern operating systems. However, in this post we’re going to look at how we can write a component provider in Typescript and package it using vercel/pkg to achieve the same portability.
```
lblackstone, line 20: I'd add a link to vercel/pkg. I wasn't previously familiar with that project.

themes/default/content/blog/nodejs-binaries-with-pkg/index.md
```diff
@@ -0,0 +1,131 @@
...
+Historical note: we used to do a similar process using the [nexe](https://github.com/nexe/nexe) project, but there’s been no releases since 2017 and therefore no support for newer versions of Node.js and we therefore consider this package as unmaintained at this point in time.
+
+## Setting up pkg
+
+### 1 - Install
+
+Pkg is distributed as an npm package which can be installed into your “devDependencies” using:
+
+```bash
+npm install -D  pkg
```
lblackstone, line 31:
```suggestion
npm install -D pkg
```

themes/default/content/blog/nodejs-binaries-with-pkg/index.md
```diff
@@ -0,0 +1,131 @@
...
+
+### 3 - Execute
+
+The main inputs that pkg needs is:
+
+* The entry point to your program for packaging
+* The target machine to build for
+* The output path to write the finished binary
+
+Here’s an example of building the project in the current directory using node v16 for macOS ARM architectures:
```
lblackstone, line 68: Is `v16` correct? The command looks like `v18`

themes/default/content/blog/nodejs-binaries-with-pkg/index.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Unsigned MacOS ARM binaries
+
+When trying to use the binaries produced by our CI, we found that [the binaries weren’t runnable on MacOS ARM architectures](https://github.com/pulumi/pulumi-awsx/issues/850) - and were forcibly killed by the operating system.
+
+This led me down quite a bit of a rabbit hole investigating signing of binaries, but it was actually resolved by simply installing the ‘ldid’ tool on our linux CI environments. Sometimes reading the warnings carefully can save you lots of time! The ldid source is available via git://git.saurik.com/ldid.git and binaries are available from various sources. Our solution was to use the [“Install Ldid” GitHub Action](https://github.com/marketplace/actions/install-ldid) to install the ldid binary and add it to the PATH in our CI workflow.
```
lblackstone, line 91:
```suggestion
This led me down quite a bit of a rabbit hole investigating signing of binaries, but it was actually resolved by simply installing the ‘ldid’ tool on our linux CI environments. Sometimes reading the warnings carefully can save you lots of time! The ldid source is available via `git://git.saurik.com/ldid.git` and binaries are available from various sources. Our solution was to use the [“Install Ldid” GitHub Action](https://github.com/marketplace/actions/install-ldid) to install the ldid binary and add it to the PATH in our CI workflow.
```

themes/default/content/blog/nodejs-binaries-with-pkg/index.md
```diff
@@ -0,0 +1,131 @@
...
+tags:
+- engineering
+- nodejs
+- publishing
+- providers
+
+meta_image: "meta.png"
+---
+
+It's well known that you can write Pulumi programs in loads of languages, but what about the providers behind the scenes which handle communication with various cloud services?
```
lblackstone, line 18: I'd suggest fleshing out the intro section more, and setting some context. This is an engineering deep dive that some readers might opt to skip, so it would be good to cover the high-level points in the first couple of paragraphs, and then transition into the technical details after that.

The questions in my mind right now are:
* What is a provider?
* What is a component provider?
* Why might you want to write a provider in Typescript?

We probably have existing content that covers most of this in detail, so a few links could go a long way here.

themes/default/content/blog/nodejs-binaries-with-pkg/index.md
```diff
@@ -0,0 +1,131 @@
...
+
+Providers are downloaded and run in many different platforms (e.g. Linux, macOS, Windows) and environments – where we can’t assume an end-user is going to have specific runtimes installed – including Node.js. Most of our providers are authored in Go – which makes it simple to produce native, static binaries for all modern operating systems. However, in this post we’re going to look at how we can write a component provider in Typescript and package it using vercel/pkg to achieve the same portability.
+
+Historical note: we used to do a similar process using the [nexe](https://github.com/nexe/nexe) project, but there’s been no releases since 2017 and therefore no support for newer versions of Node.js and we therefore consider this package as unmaintained at this point in time.
+
+## Setting up pkg
```
lblackstone, line 24: Need more of a transition here. What is this section teaching us?

lblackstone (commented):
You've got a lot of good content here, but I'd suggest working on the intro/conclusion and transitions between sections to tie it all together.

danielrbradley:
> You've got a lot of good content here, but I'd suggest working on the intro/conclusion and transitions between sections to tie it all together.

@lblackstone Thanks for the in-depth review - really appreciated! 🙇 

I've rewritten the intro and added a bit more of an outro. Let me know what you think!

themes/default/content/blog/nodejs-binaries-with-pkg/index.md
```diff
@@ -0,0 +1,131 @@
...
+```
+
+Step-by-step:
+
+1. Define a makefile target for each output file we want.
+2. Set the TARGET variable to the pkg target we’ll want to use for each output.
+3. A wildcard rule invokes pkg build using the TARGET defined above
+4. `$@` is the current makefile target - the specific binary being built
+5. The wildcard rule depends on the node_modules which is also a makefile target - to ensure packages have been restored before the build starts
+6. Define a phony `bins` target to build all bin targets
```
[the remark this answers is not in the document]
danielrbradley: Added a brief recap and sign-off 👍

themes/default/content/blog/nodejs-binaries-with-pkg/index.md
danielrbradley: I've made it a little less about providers, because it's not really about them specifically. Hopefully the reworked intro covers more of of the "Why Typescript" and "Why binary packages" part.

themes/default/content/blog/nodejs-binaries-with-pkg/index.md
```diff
@@ -0,0 +1,151 @@
+---
+title: "Node.js Native Binary Compilation Using vercel/pkg"
```
lblackstone, line 2: I noticed that you use both `vercel/pkg` and `Vercel/pkg` in this post. Which is correct?

themes/default/content/blog/nodejs-binaries-with-pkg/index.md
```diff
@@ -0,0 +1,151 @@
...
+date: 2022-09-13
+meta_desc: "Using Vercel pkg to create native binaries for nodejs providers"
```
lblackstone, line 4: Oh, here's another difference. We should pick one and use the same terminology consistently.

themes/default/content/blog/nodejs-binaries-with-pkg/index.md
```diff
@@ -0,0 +1,151 @@
...
+tags:
+- engineering
+- nodejs
+- publishing
+- providers
+
+meta_image: "meta.png"
+---
+
+In Pulumi's engineering department we build and distribute our tools as native binaries as there's no local dependencies needed. Go provides a great story for building self-contained binaries that target modern operating systems.
```
lblackstone, line 18:
```suggestion
In Pulumi's engineering department, we often build and distribute tools as native binaries to avoid the need for additional dependencies on user machines. Most of these tools are written in Go, which has good support for building self-contained binaries that target modern operating systems.
```

themes/default/content/blog/nodejs-binaries-with-pkg/index.md
```diff
@@ -0,0 +1,151 @@
...
+
+However, it's possible to also author component providers using other languages & runtimes including: Node.js, Python and .NET. However, when it comes to distributing your provider plugin, we'd still like to be able to avoid the user needing to install the required runtime in order to use the provider plugin. In this article we're looking at how we can package a Node.js program into a standalone distributable.
```
lblackstone, line 20:
```suggestion
While other Pulumi-supported languages like Node.js, Python, and .NET require additional runtime dependencies, it's possible to bundle dependencies with the program. In this article, we'll show you how to do that for a Node.js program.
```

lblackstone (approved):

PR #1924, opened 2022-09-09
Document the `nodeargs` runtime option in Pulumi.yaml

justinvp:
https://github.com/pulumi/pulumi/pull/8655 added support for a `nodeargs` runtime option for Node.js projects, but we never explicitly documented it. This PR does that.


iwahbe (approved):
LGTM

PR #1929, opened 2022-09-11
fix css filename in startup challenge

susanev:
noticed that we ask folks to name this file `styles.css` but then in later content (including the html file) we use `style.css` so if folks copy directly the css wont actually work.

if anyone prefers that i keep this line as styles and change the other 2 lines to styles, happy to do so, let me know


cnunciato (approved):
Thanks!

Issue #822, opened 2021-11-04
slot machine weirdness


susanev:
i noticed some slot machine weirdness today
that im not sure is known, so writing up an issue

- sometimes, it just static doesn't rotate at all when you freshly load the page
- sometimes it rotates a few times then stops entirely

the screenshots show the single row without a hint at the other 2 rows and no rotating
and two rows where it rotates a few times and then stopped 

<img width="897" alt="Screen Shot 2021-11-04 at 8 00 21 AM" src="https://user-images.githubusercontent.com/5489125/140339155-5f9530fa-8be7-43bf-b296-6ac2d5af76c8.png">
<img width="789" alt="Screen Shot 2021-11-04 at 8 00 05 AM" src="https://user-images.githubusercontent.com/5489125/140339159-dcb60bb4-bc89-4325-a5ca-eaa37914a2ef.png">

susanev:
@zchase i noticed some slot machine weirdness today, no urgency

gthuang:
I'm seeing this issue too. For me, it only spins a few times and then stops on the bottom image that @susanev posted above.

PR #1940, opened 2022-09-14
Update home page slot machine component

zchase:
Fixes: https://github.com/pulumi/pulumi-hugo/issues/822
Part of https://github.com/pulumi/marketing/issues/416

This PR updates the home page slow machine component based on the new mockup. I also moved up the benefits section because it looked weird having the code example present twice so quickly.

Note: There is follow up PR in `theme` to randomize the order of slot items on each page load: https://github.com/pulumi/theme/pull/258

### DIFF
--- a/themes/default/content/_index.md
+++ b/themes/default/content/_index.md
@@ -15,7 +15,7 @@ overview:
   logos:
     languages:
       - /logos/tech/typescript.svg
-      - /logos/tech/dot-net.svg
+      - /logos/tech/dotnet.svg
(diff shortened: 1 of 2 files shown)


aaronkao:
So this changes the animation of the slot machine. Do we think its better than the previous animation? I view it as a bit distracting because there is a ton of flashing going on above the fold.

sean1588 (approved):

aaronkao (approved):

PR #1952, opened 2022-09-16
Use full height for cards

cnunciato:
Little fix to make sure cards with wrapping content stay aligned horizontally.

Before:

image

After:

image


nimbinatus (approved):
LGTM

PR #1956, opened 2022-09-16
Fix errant cloud references

cnunciato:
Looks like these references to Azure and AWS slipped through in the GCP template page. 😬


nimbinatus (approved):
sorry I missed these, too

PR #1957, opened 2022-09-17
Move links out of the header area

cnunciato:
Looking closer at the content at the top of this page:

image

... I'd like to propose that we pull the links to AWS, Azure, etc., out of the header area and into the body of the page instead, for a couple of reasons:

* Right now this reads as though clicking "AWS" may lead me to more information about how to "deploy a static website on AWS", which isn't the case.
* In general, links that pull users away from the page should be limited to the lower 2/3 or so of the page. (We have a note about this [in our style guide](https://github.com/pulumi/pulumi-hugo/blob/master/STYLE-GUIDE.md#content-design) in fact, which I believe was originally suggested by @SaraDPH.)

I know these links are meant to encourage traffic to the pages they link to, so I don't  want to lose that -- I'd just prefer they weren’t so prominent. Moving them down a bit seems an okay trade-off SEO-wise. The main thing I'm interested in is just making sure we aren't directing people away from the content they're most likely here for (and that we want them to engage with).

### DIFF
--- a/themes/default/content/templates/static-website/_index.md
+++ b/themes/default/content/templates/static-website/_index.md
@@ -1,7 +1,7 @@
 ---
 title: Static Website Templates
 layout: overview
-description: Pulumi program templates are the fastest way to deploy static websites on [AWS](/aws), [Azure](/azure), or [Google Cloud Platform](/gcp). Templates come with predefined [infrastructure as code](/what-is/what-is-infrastructure-as-code) so you can get started instantly.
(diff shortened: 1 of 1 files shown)


lindsaymarolich:
Agreed -- sorry when I suggested adding them I didn't realize it was in the subheader!

lindsaymarolich (approved):

cnunciato:
@lindsaymarolich no worries, thanks for taking a look!

PR #1964, opened 2022-09-21
Update index.md

cnunciato:


cnunciato:
Verified this fix, merging proactively to unblock the pipeline z

PR #1975, opened 2022-09-23
Add some space above the headings on template overview cards

cnunciato:
Makes padding consistent on the section and overview pages and adds a one-pixel border around cards for definition. (We do this also in Registry.) 

image


scottslowe (changes requested):
LGTM

scottslowe (approved):
Sorry, clicked the wrong button! LGTM

PR #1982, opened 2022-09-26
add an FAQ item about migrating stacks

jaxxstorm:


susanev (commented):
@jaxxstorm shouldnt this be under the pulumi service heading on line 43?

jaxxstorm:
I think a case can be made for both, I'm happy to put it wherever you think is best. Having it under the "organization" section makes sense to me.

susanev (approved):

PR #1987, opened 2022-09-27
Fix broken link, tweak a few words

cnunciato:
The link to Transferring Stacks was broken, so this fixes that and makes a few tweaks to better comply with the [style guide](https://github.com/pulumi/pulumi-hugo/blob/master/STYLE-GUIDE.md#links).


susanev (approved):
thank you for fixing the here link, i was gonna come back and fix that later but am so glad you were able to do it now!!

PR #1991, opened 2022-09-29
Add markdownlint customizations

cnunciato:
Messing around with adding some custom lint rules to help prevent common build-breaking pitfalls like relrefs with trailing slashes, etc.

### DIFF
--- a/scripts/lint/lint-markdown.js
+++ b/scripts/lint/lint-markdown.js
@@ -265,6 +265,50 @@ const opts = {
         // Allow indentation in unordered lists.
         MD007: false,
     },
+    customRules: [
+        {
+            names: ["relref-anchor-trailing-slash"],
+            description: "Hugo relref to anchor has trailing slash.",
+            information: new URL("https://gohugo.io/functions/relref/"),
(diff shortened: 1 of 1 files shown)


cnunciato:
Interestingly, while writing these custom rules does work nicely, what I'm finding (and am somewhat surprised by) is that rules like "relrefs shouldn't  have trailing slashes" aren't really correct, as relrefs with trailing slashes do seem to work -- sometimes.

The currently failing PR job shows this:

```
/home/runner/work/pulumi-hugo/pulumi-hugo/themes/default/content/docs/guides/crosswalk/kubernetes/playbooks.md:
Line 13: Hugo relref to anchor has trailing slash. ["/some/path/#anchor" should be "/some/path#anchor"].
Line 13: Hugo relref has trailing slash. ["/some/path/" should be "/some/path"].
```

The check for trailing slashes is failing as expected, which is great -- but the actual relrefs (which are in the code today) obviously work. So just checking for things like trailing slashes by itself won't be enough, unfortunately.

cnunciato:
Closing this as no longer necessary as of #2071.

PR #2006, opened 2022-10-03
Fix refresh option in pulumi project file example

geekflyer:
Follow up from https://github.com/pulumi/pulumi/issues/10824


larry-aptos (approved):

susanev (approved):
thank you!

PR #2018, opened 2022-10-06
Added post on using secrets with lambda functions

pierskarsenbarg:


themes/default/content/blog/safe-lambda-secrets/index.md
```diff
@@ -0,0 +1,545 @@
...
+    - serverless
+---
+
+The subject of how to make use of secrets in Lambda Functions comes up a fair bit and although there seems to be a lot of discussion on where you _should_ store them, the one thing that comes up is that **you should never store the plain text values of secrets in the Lambda Functinon's environment variables**. One such discussion I was having with a customer, made me think about how it should be possible to take the secrets that you've got on your [stack config file](https://www.pulumi.com/docs/intro/concepts/config/) and then use them to configure your Lambda Function, with the plain text values going into the Function's environment variables and the encrypted secret values going into AWS' Secrets Manager.
+
+<!--more-->
+
+## Getting started
+
+I'm going to assume you've already got Pulumi and AWS set up, you're logged into your backend and you've already created a project and stack. I'm also using the Pulumi Service as my [secrets provider]({{< relref "/docs/intro/concepts/secrets/#configuring-secrets-encryption" >}}).
```
pierskarsenbarg, line 19:
```suggestion
I'm going to assume you've already got Pulumi and AWS set up, you're logged into your backend, and you've already created a project and stack. I'm also using the Pulumi Service as my [secrets provider]({{< relref "/docs/intro/concepts/secrets/#configuring-secrets-encryption" >}}).
```

themes/default/content/blog/safe-lambda-secrets/index.md
```diff
@@ -0,0 +1,545 @@
...
+
+{{% notes type="info" %}}
+Although Lambda Functions are free to deploy and you get a generous allowance as part of AWS' free tier, storing and accessing secrets is not. You can view the cost on the [AWS Secrets Manager Pricing page](https://aws.amazon.com/secrets-manager/pricing/).
```
pierskarsenbarg, line 22:
```suggestion
Although Lambda Functions are free to deploy, and you get a generous allowance as part of AWS' free tier, storing and accessing secrets is not. You can view the cost on the [AWS Secrets Manager Pricing page](https://aws.amazon.com/secrets-manager/pricing/).
```

themes/default/content/blog/safe-lambda-secrets/index.md
```diff
@@ -0,0 +1,545 @@
...
+import * as pulumi from "@pulumi/pulumi";
+import * as aws from "@pulumi/aws";
+```
+
+### A warning about accessing the configuration
+
+Before we continue, it's worth pointing out the following:
+
+{{% notes type="warning" %}}
+On `pulumi up`, secret values are decrypted and made available in plaintext at runtime. These may be read through any of the standard `pulumi.Config` getters. While it is possible to read a secret using the ordinary non-secret getters, this is almost certainly not what you want. Use the secret variants of the configuration APIs instead, since this ensures that all transitive uses of that secret are themselves also marked as secrets.
```
pierskarsenbarg, line 80:
```suggestion
On `pulumi up`, secret values are decrypted and made available in plaintext at runtime. These may be read through any of the standard `pulumi.Config` getters. While it is possible to read a secret using ordinary non-secret getters, this is almost certainly not what you want. Use the secret variants of the configuration APIs instead since this ensures that all transitive uses of that secret are themselves also marked as secrets.
```

themes/default/content/blog/safe-lambda-secrets/index.md
```diff
@@ -0,0 +1,545 @@
...
+Since we're using structured configuration to store the values, we can access them in a strongly-typed way. We're going to create an interface to hold the structure:
+
+```typescript
+interface LambdaConfig {
+    envvars: string[];
+    secrets: pulumi.Output<string>[];
+}
+```
+
+You'll notice that the property names match the dictionary names in the stack config and that the secrets property is of type `pulumi.Output<string>[]` instead of `string[]`. This is because they're Pulumi secrets and we want to handle them in the correct way.
```
pierskarsenbarg, line 96:
```suggestion
You'll notice that the property names match the dictionary names in the stack config and that the secrets property is of type `pulumi.Output<string>[]` instead of `string[]`. This is because they're Pulumi secrets, and we want to handle them in the correct way.
```

themes/default/content/blog/safe-lambda-secrets/index.md
```diff
@@ -0,0 +1,545 @@
...
+Next we'll access the config and store the values in a variable:
+
+```typescript
+const config = new pulumi.Config();
+const lambdaconfig = config.requireObject<LambdaConfig>("lambdawithsecrets");
+```
+
+### Storing the secret values
+
+As I said in the beginning, we're going to store the secrets from the stack config in AWS Secrets Manager and then the ARN of each of the secrets will be stored in the Lambda Function's environment variables.
```
pierskarsenbarg, line 107:
```suggestion
As I said in the beginning, we're going to store the secrets from the stack config in AWS Secrets Manager, and then the ARN of each of the secrets will be stored in the Lambda Function's environment variables.
```

themes/default/content/blog/safe-lambda-secrets/index.md
```diff
@@ -0,0 +1,545 @@
...
+    environment: {
+        variables: lambdaEnvVars
+    },
+    timeout: 15
+})
+
+export const lambdaName = lambda.name;
+```
+
+There's nothing complicated about the function. We're using the role from earlier, setting the environment variables to be the object we created just above and pointed the `code` input at a local folder called `app` which we'll look at next. The lambda name being exported at the end is so that we can use the Pulumi CLI to get this value and invoke the function using the AWS CLI.
```
pierskarsenbarg, line 244:
```suggestion
There's nothing complicated about the function. We're using the role from earlier, setting the environment variables to be the object we created just above and pointed the `code` input at a local folder called `app`, which we'll look at next. The lambda name being exported at the end is so that we can use the Pulumi CLI to get this value and invoke the function using the AWS CLI.
```

themes/default/content/blog/safe-lambda-secrets/index.md
```diff
@@ -0,0 +1,545 @@
...
+Outputs:
+    lambdaName: "lambdaWithSecrets-ca62cf8"
+
+Resources:
+    + 10 created
+
+Duration: 24s
+```
+
+Let's see what this has done. If we go to look at the Lambda Function in the console, click on the `Configuration` tab and select `Environment variables` you'll see the two plaintext environment variables and their values along with the two secret environment variables containing the ARNs to the relative secret in the Secrets Manager. Since we know what the environment variable key is (it's the uppercase version of what's in the stack config), it means we can reliably use that same name in the function code (the two lines where we call the `getSecretForLambda` function).
```
pierskarsenbarg, line 369:
```suggestion
Let's see what this has done. If we go to look at the Lambda Function in the console, click on the `Configuration` tab and select `Environment variables` you'll see the two plaintext environment variables and their values along with the two secret environment variables containing the ARNs to the relative secret in the Secrets Manager. Since we know the environment variable key (the uppercase version of what's in the stack config), we can reliably use that same name in the function code (the two lines where we call the `getSecretForLambda` function).
```

MitchellGerdisch (approved):
Very nice write up.

PR #2027, opened 2022-10-10
Update GitHub app permissions screenshot

blampe:


susanev:
@blampe do you mind updating this screenshot so its the same view our users will see instead of what we see as admins of the app? that app settings link i do not believe will be visible to folks. i am also happy to do this if you don't mind me pushing a commit.

themes/default/content/docs/guides/continuous-delivery/github-app.md
```diff
@@ -47,10 +47,10 @@ configure it to only be used by certain repositories.
 
 ![Configuration Page](/images/docs/github-app/org-configuration.png)
 
-The Pulumi GitHub application does not have access to your source code. It will only report
-status on pushes / pull requests that happen for repositories it is configured to access. You can
-also uninstall the GitHub application at any time without impacting your stacks or
-other Pulumi-managed resources.
+The GitHub app will report Stack operations on GitHub commits and pull requests
```
susanev, line 50: why are we droping Pulumi from before GitHub?

blampe: Typo, re-added.

blampe:
> @blampe do you mind updating this screenshot so its the same view our users will see instead of what we see as admins of the app? that app settings link i do not believe will be visible to folks. i am also happy to do this if you don't mind me pushing a commit.

NP I took care of this.

susanev (approved):

PR #2028, opened 2022-10-10
azure static website blog post

sean1588:


themes/default/content/blog/deploy-azure-static-website/index.md
```diff
@@ -0,0 +1,205 @@
...
+* Uploads the website contents located in the specified directory
+* Optionally can configure a CDN, custom domain, and certs.
+
+## Using the component to deploy a static website
+
+Let’s walk through a simple example of using this component to deploy a static website. Today, we are going to provision a storage account, upload the website contents, and wire it up to an Azure CDN.
+
+In order to follow along with this example you will need to have the following:
+
+* Pulumi CLI installed on your machine
```
zchase, line 33: I think adding some links how to install Pulumi and set up Azure would be helpful.

themes/default/content/blog/deploy-azure-static-website/index.md
```diff
@@ -0,0 +1,205 @@
...
+      sitePath: "../site"
+outputs:
+  websiteURL: ${web.cdnURL}
+```
+
+{{% /choosable %}}
+
+{{% /chooser %}}
+
+This code creates a new instance of the Website component and sets a couple of properties. `sitePath` is the location where your website contents live and `withCDN` is set to true to let the component know that you would like to provision a CDN. It also exports the `cdnURL` property so you can see where the site can be accessed. That’s all that is needed and now we are ready to deploy!
```
zchase, line 190: I think you should expand a bit on what is expected to be in `sitePath`. For example I think you used React here so for example you could say the component expects the output of a react build, not the path the entire application. I think users might make an assumption the component builds it for them.

zchase (approved):
🎉

PR #2035, opened 2022-10-11
Remove vestigial column values in Resource Options table

### LINKED ISSUE #2034: YAML reference has lost descriptions in resource options table
https://github.com/pulumi/pulumi-hugo/pull/1714 removed table headings from https://github.com/pulumi/pulumi-hugo/blame/master/themes/default/content/docs/reference/yaml/_index.md, without removing cells in the rows. The result is that some descriptions are replaced with `No` (originally from the "Required" column).

squaremo:
A previous PR (#1714) removed columns from the resource options table, but left a few rows with too many values, which crowded out the descriptions. This removes the extra values so the description column has all the descriptions.

Fixes #2034.


cnunciato (approved):
Thank you!

PR #2036, opened 2022-10-11
Move theme back into this repo

zchase:
Fixes: https://github.com/pulumi/theme/issues/105


kimberleyamackenzie (approved):
I'm super on board with this change, thank you @zchase for doing this work!  

A couple thoughts -
- Before we merge, can we do some updating/adding of read me docs to make the new flows clearer?  
- Could we add a script or single command that builds everything to make the dev flow easier?
- I'm not 100% clear on how to use the assets.html file, which has the note `Use this partial to pull the JS and CSS bundles into your project.` - could you clarify?  Is this for if I want to pull the theme styles into a project in a separate repo?

I'm super happy to pitch in on any of the above if you don't have the bandwidth, just let me know.

Approving so I don't block for another long chunk of time, but it would also be cool to get a preview of this going for folks to poke around in before it's merged.

sean1588 (approved):
thanks for doing this @zchase. the changes look good to me. I guess we can double check and verify everything once this gets in by turning off the auto-merge workflow like you suggested.

Issue #2041, opened 2022-10-12
Updates to CED Page


lindsaymarolich:
We have a few changes to make to update Cloud Engineering Days:
1/ Replace the 'Hackathon' tile with 'Case Studies' and the text below with: 'Learn from Pulumi customers on how they scaled up their infrastructure velocity and reliability with Pulumi. Infrastructure experts from companies like Fauna, Panther Labs, and Altana AI will present best practices for using Pulumi and infrastructure as code to increase velocity and even build their product platforms.' Please change button to a 'Register' button like the others
2/ Please remove Kat and David from the workshops presenter images
3/ Please remove the section about the Pulumi Community Hackathon (right below 'Join the Conversation')
4/ Please add partner logos to the section where Hackathon used to be. This can just be a white background with logos equally distributed horizontally across the page (I will add the logos shortly)
5/ Please remove 'Hackathon Launch' from the agenda and update 'Workshop Day' to the 12PM slot instead of 1PM
6/ It was noticed that sometimes the images of people look stretched at different screen sizes, can we prevent this?
<img width="428" alt="Screen Shot 2022-10-12 at 2 31 38 PM" src="https://user-images.githubusercontent.com/69173062/195453272-07c5f863-3680-488a-8cd0-3072a49b97a4.png">

shughes26:
https://github.com/pulumi/pulumi-hugo/pull/2040 this PR covers:
1 (halfway)/ removes hackathon tile (not replaced yet)
2/ removes Kat and David from presenters
3/ removes hackathon section
5/ removes hackathon launch from the agenda
6/ fixes stretchy images

will open a separate PR for:
1 (the rest of it)/ 'Case Studies'
5/ move up workshop day timing
4/ partner logos

wamazonsmith-web:
Add Customer speakers in the Pulumi Demo's and best practices:
Wondering if there is a way to pull in the customer logos so they feature prominently?
Dan Swartz, Principal Software Engineer, Altana AI
Dennis Webb, Staff Software Engineer, Panther

Also adding to Schedule at a Glance:
Can we add the customer logos in the relevant spots?
Within Achieving Results with Pulumi Customers
Add in the two above
![danswartz](https://user-images.githubusercontent.com/75398972/195645915-47cad009-627d-405c-be11-b125b4d1f307.jpeg)

![](https://files.slack.com/files-pri/TC2U317C7-F046AQ4FXB6/dscf9081.jpg)
[...]

PR #2044, opened 2022-10-13
CED updates

shughes26:
https://github.com/pulumi/pulumi-hugo/issues/2041
- CED updates - add speakers, case studies, partner logos
- update CED workshop presenters to match current info


lindsaymarolich:
Looks great! A few minor things; could we please add the Microsoft logo to the partner logo section and constrain them all to the same height? It would also be great if they could be distributed horizontally rather than stacked (shown below). In the mockup below I also added the 'Our Partners' section header, and centered the partner logos on the case studies tile so they align with the workshops tile:

Screen Shot 2022-10-13 at 4 01 34 PM


Microsoft-logo_rgb_c-gray


Thank you!

lindsaymarolich (approved):

PR #2045, opened 2022-10-13
List languages in template frontmatter

cnunciato:
This change adds a new shortcode, `{{< templates/pulumi-new >}}`, that reads from template frontmatter to build the language chooser and `pulumi new` commands. Also adjusts the right-hand nav to use that same list to build the links to the templates repository. 

cc @gthuang as this makes a few tweaks to the templates archetype, which now expects that list of languages, along with the name to use for the Pulumi project directory (with defaults included of course).


nimbinatus (approved):
LGTM

PR #2049, opened 2022-10-14
feat: points python docs at new static links

kpitzen:


config/_default/menus.yml
```diff
@@ -99,7 +99,7 @@ reference:
 
   - name: Python
     parent: Pulumi SDK
-    url: /docs/reference/pkg/python/pulumi
+    url: /docs/reference/pkg/python/providers/pulumi.html
```
justinvp, line 102: Why is it under `/providers` now? Seems odd since `pulumi` isn't a provider. Can we fix that?

susanev:
@kpitzen did you still want to get this in?

kpitzen:
> @kpitzen did you still want to get this in?

I do, but I think I want to chat with @justinvp about:
1) Making these docs available in a more intuitive location
2) Making sure _all_ docs are being successfully generated

I appreciate the bump, though!  More to come soon

susanev:
@kpitzen @justinvp how are we doing on this?

PR #2050, opened 2022-10-14
Add templates as options on getting started landing page

zchase:
This PR adds the newly shipped templates as options in the Getting Started Guide. The main goal here is increase awareness and usage of the templates, and by offering them as options on the getting started page we should get more visibility.

**Hypothesis**

By presenting users more use case specific options in the getting started guide we will increase the CTR of users to these templates/guides and in turn increase the conversion rate of the website resulting in new users.

getting-started-options-tempaltes

### DIFF
--- a/themes/default/content/docs/get-started/_index.md
+++ b/themes/default/content/docs/get-started/_index.md
@@ -28,7 +28,7 @@ Select one of the following options to get started:
 <div class="tiles flex-wrap justify-center items-stretch mt-4">
     <div class="pb-4 md:pr-4 md:w-1/2">
         <div class="block rounded shadow border border-gray-300 p-3 h-full flex flex-col">
-            <h4><i class="fas fa-folder text-blue-400 pr-2"></i>Simple App</h4>
+            <h4><i class="fas fa-folder text-blue-400 pr-2"></i>Starter App</h4>
             <p>Create a static website with an S3 bucket.</p>
             <div class="flex flex-grow items-end">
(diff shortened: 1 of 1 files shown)


susanev:
@zchase can you sync up with @anita-trimbur we are calling the current one "Starter" in the service UI instead of "Simple". I think it would be good to align these things. Can y'all decide together on a direction?

zchase:
@susanev I'll change it to "Starter" to align with y'all, "Simple" was just what I picked in the moment. Thanks for letting me know!

aaronkao:
I think we should wait to roll this out until after some of this is addressed https://github.com/pulumi/devrel-team/issues/425. My original thinking was that Getting Started eventually would just use one of the template as part of its own flow. This level of choice I think might be more confusing. 

The short term change we could make is adding this on the Next Steps page - https://www.pulumi.com/docs/get-started/aws/next-steps/

aaronkao:
Discussed in MBR on 10/18/2022. Good to ship and test. Tracking metrics https://docs.google.com/spreadsheets/d/1Z-xuAlLlRPvG6ksL8jcHVAHHjD1Y_SZAROYrptZyRcA/edit

sean1588 (approved):
Just one comment about the starter app. Other than that lgtm.

sean1588:
> I do think it might be worth considering removing the links to pulumi/examples for now (since that experience is so much less guided than /templates), but I'll leave that to you.

plus one on that

PR #2052, opened 2022-10-14
Webhooks doc fixes

meagancojocar:
- fixed contact us link going to about us
- wording clean up

### DIFF
--- a/themes/default/content/docs/intro/pulumi-service/webhooks.md
+++ b/themes/default/content/docs/intro/pulumi-service/webhooks.md
@@ -18,10 +18,10 @@ To try it out, start a [trial](https://app.pulumi.com/site/trial) now.
 {{% /notes %}}
 
 Pulumi Webhooks allow you to notify external services of events
-happening within your Pulumi organization or stack. For example,
+happening within your Pulumi organization. For example,
 you can trigger a notification whenever a stack is updated.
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/intro/pulumi-service/webhooks.md
```diff
@@ -18,10 +18,10 @@ To try it out, start a [trial](https://app.pulumi.com/site/trial) now.
 {{% /notes %}}
 
 Pulumi Webhooks allow you to notify external services of events
-happening within your Pulumi organization or stack. For example,
+happening within your Pulumi organization. For example,
```
susanev, line 21: why are we dropping stack from here? there are also stack webhooks

shughes26 (approved):

themes/default/content/docs/intro/pulumi-service/webhooks.md
meagancojocar: This line isn't about the types of webhooks its about the types of events. The clause "happening within your organization" covers both `stack_update`, `stack_preview` and `stack` events. They all happen within an organization. Then later in the docs we break down how `stack` events are only delivered to organization-based webhooks.

susanev: hmm, I'm not sure I follow, but if it makes sense to y'all feel free to merge.

PR #2054, opened 2022-10-16
clarify stack name in actions docs

susanev:
okay i got totally hung up on this tonight; as pulumi was creating things in my individual account or trying to, when the stack really existed in an org. im not sure if this is the best way to make this clear. super open to other ideas and suggestions!

the downside is that this wont just automatically work for folks if they copy and paste this directly, they will need to change this value. 

we could try working it into the prose somehow.

edit: i realized that this may frustrate folks that are using their own backend since they don't have the org concept. i do still think its important to figure out how to include this info somewhere on this page.

### DIFF
--- a/themes/default/content/docs/guides/continuous-delivery/github-actions.md
+++ b/themes/default/content/docs/guides/continuous-delivery/github-actions.md
@@ -93,7 +93,7 @@ jobs:
       - uses: pulumi/actions@v3
         with:
           command: preview
-          stack-name: dev
+          stack-name: org-name/stack-name # When using an individual account, only use stack-name.  
         env:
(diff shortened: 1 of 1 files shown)


guineveresaenger (commented):
nit: I think you missed one on [line 96](https://github.com/pulumi/pulumi-hugo/pull/2054/files#diff-012b98f573034a88bde60bc3b26463cba207de706b42c4b0b6bb249d7f1d9091L96)

My suggestion would be, and I know this is controversial, to add a comment in the yaml example like so:

```
stack-name: org-name/stack-name #for individual accounts, only use stack-name
```

RobbieMcKinstry (approved):
I added a commit that adopts @guineveresaenger 's feedback. LGTM; I'm happy to merge.

PR #2056, opened 2022-10-17
Fix incorrect argument name

lblackstone:


squaremo (approved):
This is indeed the correct option name, it all checks out -- thanks Levi 🍓

PR #2064, opened 2022-10-20
Add a page listing all blog tags

cnunciato:
Adds a page that lists all blog tags alphabetically. This is helpful mainly for reference, as blog-post authors frequently want to know what tags we've already defined so they can use them.

Also fixes a bug with the sidebar logic. (It now works as intended -- not sure what happened there, but in any case, should be good now!)

### DIFF
--- a/themes/default/content/blog/tag.md
+++ b/themes/default/content/blog/tag.md
@@ -1,0 +1,7 @@
+---
+title: "Pulumi Blog: All tags"
+layout: tags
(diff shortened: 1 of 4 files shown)


susanev:
@cnunciato did you need any help with this?

SaraDPH (approved):

cnunciato:
Happy to address any feedback as a follow up.

PR #2067, opened 2022-10-20
Refresh runtime versions in the CI guides

ringods:
Customer reported the runtime versions were quite outdated. Updated these to reflect the current runtime versions better.


nimbinatus (approved):
LGTM

PR #2071, opened 2022-10-21
Replace relrefs with plain Markdown links

cnunciato:
Given the size and composition of our website today, we’re no longer able to use Hugo relrefs as effectively as we used to. This change removes all Hugo relrefs in Markdown files, replaces them with regular Markdown links, and adds a linting rule to prevent content authors from using relrefs in Markdown files in this repo going forward.

What this means is that in the Markdown files where you write this today:

```
[Some link title]({{< relref "/some/path" >}})
```

you can just write this from now on instead:

```
[Some link title](/some/path/)
```

As a feature of Hugo, relrefs are great in that they help content authors avoid writing bad links by failing the build when a link target can’t be found. For us, however, they’re problematic in that the links we often write point to content that doesn’t exist in the current module, so we have to suppress relref failures in development builds (essentially negating their usefulness entirely). This often leads to downstream failures in integration builds because of syntax errors or other Hugo-idiosyncratic reasons, slowing us down and leaving people feeling confused and frustrated.

The upshot of this change is that by suppressing relref and letting all links through unchecked, we may occasionally ship a page with a broken link, but we’ll no longer fail the build for random reasons that don’t make sense to Hugo non-experts and that force us to react in the moment to unbreak the pipeline. Overall, it will improve our authoring and shipping velocity — and pragmatically, since we crawl the full site every day for bad links anyway (and alert ourselves on them), we should be able to keep broken links under control by keeping an eye on what breaks and fixing things up on a rolling basis just as we do today.

Note that this change doesn’t prevent using relrefs elsewhere; we’ll still use them in layout files, can use them in other modules, etc. It only applies to this module, as this is the one where most content authoring happens. 

Lastly, since this PR is too big to review closely, I’ve set up a PR in the docs repo that uses this branch to build full site previews, which I’ve been testing with the link checker to guard against any regressions: 

https://github.com/pulumi/docs/pull/8133

If a scan of the changes aligns with what you’d expect, and you agree with the change in principle, your ✅ would be appreciated; there’s definitely no need to review every change here, as all of these links are well covered by our automated testing. Thank you!


scottslowe:
This changeset is **HUGE** and will be difficult for anyone to review within a reasonable time period. What automated tools, if any, do we have to verify the links that were affected by the change?

cnunciato:
@scottslowe See my note at the bottom of the PR description. There’s a link there to a PR in pulumi/docs that I’ve been using to run local tests using the same link checker we run against the production website every day.

scottslowe:
@cnunciato Thanks for the clarification! On that basis, the subset of changes I reviewed seem reasonable, so I'll go ahead and approve the PR.

zchase (approved):
Seems reasonable to me. Some links don't have a trailing slash now and I wondering if that is going to cause a bunch of SEO warnings. Does hugo automatically append the trailing slash or will have to fix the missing ones?

cnunciato:
@zchase it does, and good catch. I added the trailing slashes in https://github.com/pulumi/pulumi-hugo/pull/2071/commits/9d842866951135ab7e6cc9acf3fe3d39f105ef1e and we'll probably want to add some linting for this as well.

themes/default/content/docs/intro/languages/go.md
```diff
@@ -20,22 +20,22 @@ The fastest way to get up and running is to choose from one of the following Get
 
 <div class="tiles mt-4">
     <div class="flex-1 pb-4 md:mr-4">
-        <a class="tile p-4" href="{{< relref "/docs/get-started/aws" >}}?language=go">
+        <a class="tile p-4" href="/docs/get-started/aw/?language=go">
```
cnunciato, line 23:
```suggestion
        <a class="tile p-4" href="/docs/get-started/aws/?language=go">
```

themes/default/content/partner/aws.md
```diff
@@ -12,7 +12,7 @@ hero:
         SDK enables you to build, deploy, and manage your AWS infrastructure faster
         and with more confidence, using modern programming languages and software engineering
         practices. Leverage the full expressivity of general-purpose languages ([TypeScript/JavaScript](/docs/intro/languages/javascript/),
-        [Python](/docs/intro/languages/python/), [Go](/docs/intro/languages/go/), [C#](/docs/intro/languages/dotnet/), [Java](/docs/intro/languages/java/))
+        [Python](/docs/intro/languages/python/), [Go](/docs/intro/languages/go/), [C#](/docs/intro/languages/dotnet/), [Java](/docs/intro/languages/java//)
```
cnunciato, line 15:
```suggestion
        [Python](/docs/intro/languages/python/), [Go](/docs/intro/languages/go/), [C#](/docs/intro/languages/dotnet/), [Java](/docs/intro/languages/java/)
```

cnunciato (commented):
One last pass through..

cnunciato:
Ok @zchase -- on this:

>Some links don't have a trailing slash now and I wondering if that is going to cause a bunch of SEO warnings.

.. I went through this PR last night and noticed many links still don't have trailing slashes; unfortunately my regex fu wasn't perfect here. My understanding is that tools like Screaming Frog can crawl a full site and produce reports for things like this, though, so if that's possible, I'd be happy to work off such a report to address these as follow ups (and then add some linting to look out for them going forward). Is generating a report like that something you could help with?

PR #2101, opened 2022-10-24
Update to template page helper

gthuang:
Fixes to catch up the standard page template with the latest changes


cnunciato (approved):
Oh, I'm sorry! I totally thought I'd done these. Thanks for keeping on top of it.

PR #2111, opened 2022-10-26
Apply Coding Standards to Pulumi Challenge (Startup in a Box)

phillipedwards:
- Remove unused variables.
- Remove call to `this.RegisterOutputs` - this is misleading; it is not required (nor needed here) and causes confusion.
- Remove private class properties where not needed.
- Removed constructor and output type - both are not used.
- Formatting...


phillipedwards:
I figured since the Pulumi Challenge is front and center to our `kubecon` both I might as well give the code some love.

phillipedwards:
On second thought I imagine the previous code matches the video, so I'm going to close this issue :(

PR #2120, opened 2022-10-26
remove from workshop from pulumi tv

sean1588:


themes/default/content/resources/getting-started-with-iac-aws/index.md
```diff
@@ -10,7 +10,7 @@ featured: false
 pre_recorded: false
```
SaraDPH, line 11: I think that's where it has to be **true**

SaraDPH (approved):

PR #2121, opened 2022-10-26
Resolve broken internal/external links

tofuni:
Resolved handful of broken internal and external links across /blogs/ and /docs/ as part of technical SEO maintenance.


zchase:
This has been moved to this PR so we can build a preview: https://github.com/pulumi/pulumi-hugo/pull/2213

PR #2153, opened 2022-10-29
Project config blog post

Frassle:


Zaid-Ajaj (approved):
The section looks nice and short for people to quickly start using hierarchical config. Links to github comments could be replaced by a short summary but I also like that we link the feature to more context about what the users wanted and what we want to do next

themes/default/content/blog/project-config-mvp/index.md
```diff
@@ -0,0 +1,97 @@
...
+    - config
+---
+
+One of our most up-voted feature requests (with 78 thumbs ups) is to [support hierarchical config](https://github.com/pulumi/pulumi/issues/2307). We're happy to announce that we've now released the first part of plans to support this feature.
+
+Pulumi will now allow you to set configuration values in your `Pulumi.yaml` file, using the given value as a default for all stacks in the project. While we expect even this first level of support will be incredibly useful to many people we also want to assure you that we have many more plans in place to make this feature even better.
+
+### Project level configuration values
+
+Pulumi has always had the concept of [configuration per stack](/docs/intro/concepts/config/), and the changes for project level configuration haven't fundamentally changed that. You can still set config for stacks via the the `pulumi config` CLI command, or by editing the `Pulumi.<stack>.yaml` file. Project level config just gives you the possibility to set default values for all stacks in your `Pulumi.yaml`.
```
gthuang, line 20:
```suggestion
Pulumi has always had the concept of [configuration per stack](/docs/intro/concepts/config/), and the changes for project level configuration haven't fundamentally changed that. You can still set config for stacks via the `pulumi config` CLI command, or by editing the `Pulumi.<stack>.yaml` file. Project level config just gives you the ability to set default values for all stacks in your `Pulumi.yaml`.
```

themes/default/content/blog/project-config-mvp/index.md
```diff
@@ -0,0 +1,97 @@
...
+```yaml
+name: project
+runtime: go
+config:
+    obj:
+        value:
+            foo: bar
+```
+
+This is to allow the engine to distinguish between values and schemas, which will show below.
```
gthuang, line 52:
```suggestion
This allows the engine to distinguish between values and schemas, which we will show below.
```

themes/default/content/blog/project-config-mvp/index.md
```diff
@@ -0,0 +1,97 @@
...
+As you can see this allows setting default configuration values per-stack for configuring the engine or providers, but you can also set default values for project configuration. For example maybe you have a default count of the number of virtual machines to create:
+
+```yaml
+name: project
+runtime: go
+config:
+    vm-count: 2
+```
+
+The only caveat to mention about this system, and it's related to the next section where we're going to discuss specifications, is that if you have object values they must be nested under a key `value`. So if you want to set the key `obj` to the object value `{ foo: "bar" }` you would need to write it as such:
```
gthuang, line 41:
```suggestion
The only caveat to mention about this system is that if you have object values they must be nested under a key `value`. So if you want to set the key `obj` to the object value `{ foo: "bar" }` you would need to write it as such:
```

themes/default/content/blog/project-config-mvp/index.md
```diff
@@ -0,0 +1,97 @@
...
+    obj:
+        value:
+            foo: bar
+```
+
+This is to allow the engine to distinguish between values and schemas, which will show below.
+
+### Configuration specification
+
+A related feature to defining values for config at the project level is [being able to define a schema for the config](https://github.com/pulumi/pulumi/issues/1052). This allows you to define what configuration settings you expect in your `Pulumi.yaml`, the engine can then use that information to validate configuration for every stack.
```
gthuang, line 56:
```suggestion
A related feature to defining values for config at the project level is [being able to define a schema for the config](https://github.com/pulumi/pulumi/issues/1052). This allows you to define what configuration settings you expect in your `Pulumi.yaml` so that the engine can use that information to validate configuration for every stack.
```

themes/default/content/blog/project-config-mvp/index.md
```diff
@@ -0,0 +1,97 @@
...
+
+This is done using a very similar declaration to JSON schema, or our Pulumi JSON schema (the schemas from resource provider that drive our code generators). Below is a small example of some of the things you can declare:
```
gthuang, line 58:
```suggestion
This is done using a very similar declaration to JSON schema, or our Pulumi JSON schema (the schemas from resource providers that drive our code generators). Below is a small example of some of the things you can declare:
```

themes/default/content/blog/project-config-mvp/index.md
```diff
@@ -0,0 +1,97 @@
...
+        description: an array of subnets to create
+        items:
+            type: string
+    vaultKey:
+        type: string
+        description: which vault key to use for this stack
+        secret: true
+```
+
+These are declared in the same config block as the other project level config shown through this post. You can mix the two together, to declare your project settings and set up defaults for providers:
```
gthuang, line 77:
```suggestion
These are declared in the same config block as the other project level config shown through this post. You can combine the two by declaring your project settings and defining defaults for providers together:
```

themes/default/content/blog/project-config-mvp/index.md
```diff
@@ -0,0 +1,97 @@
...
+config:
+    aws:region: us-east-2
+    createVpc:
+        type: boolean
+        default: false
+```
+
+### YAML
+
+When we released our support for simple Pulumi programs in YAML we had to add some support for configuration with it. This was done via the [`configuration` key](https://www.pulumi.com/docs/reference/yaml/#configuration) in the `Pulumi.yaml` file. We'll be deprecating support for that key now, as YAML will instead use the new standard project configuration.
```
gthuang, line 89:
```suggestion
When we released our support for simple Pulumi programs in YAML we added some support for configuration with it. This was done via the [`configuration` key](https://www.pulumi.com/docs/reference/yaml/#configuration) in the `Pulumi.yaml` file. We'll be deprecating support for that key now, as YAML will instead use the new standard project configuration.
```

themes/default/content/blog/project-config-mvp/index.md
```diff
@@ -0,0 +1,97 @@
...
+
+```yaml
+config:
+    aws:region: us-east-2
+    createVpc:
+        type: boolean
+        default: false
+```
+
+### YAML
```
gthuang, line 87: perhaps update the title to call out the deprecation change more?

Frassle, line 87:
```suggestion
### Deprecation of 'configuration' in YAML
```
?

PR #2159, opened 2022-10-31
Kubernetes web app template page

gthuang:
Adding landing page for new k8s deployment+service template.


gthuang:
@cnunciato how do we change the right-side section that says "This template is also available for:"? It currently lists the kubernetes cluster templates, but it doesn't make sense to do that for this page.

themes/default/content/templates/kubernetes/webapp/index.md
```diff
@@ -0,0 +1,99 @@
...
+
+When the deployment completes, Pulumi exports the following [stack output](/docs/intro/concepts/stack#outputs) values:
+
+deploymentName
+: The name of your new Kubernetes Deployment.
+
+serviceName
+: The name of your new Kubernetes Service.
+
+Output values like these are useful in many ways, most commonly as inputs for other stacks or related cloud resources. The computed `serviceName`, for example, can be used from the command line to open the newly deployed web application in your favorite web browser:
```
scottslowe, line 55: Unfortunately, you can't use the computed `serviceName` output in this way. Although it is exposed as a Service, it's not consumable using that URL.

themes/default/content/templates/kubernetes/webapp/index.md
```diff
@@ -0,0 +1,99 @@
...
+
+Projects created with the Kubernetes Application template expose the following [configuration](/docs/intro/concepts/config) settings:
+
+namespace
+: The name of the namespace to be created in your existing cluster. Defaults to `default`.
+
+replicas
+: The number of replicated Pods to be created in your new Deployment. Defaults to `1`.
+
+webserverconfig
```
scottslowe, line 71: This isn't exposed as a configuration setting.

themes/default/content/templates/kubernetes/webapp/index.md
```diff
@@ -0,0 +1,99 @@
...
+: The ConfigMap that stores your application's configuration. By default, it is configured for Nginx.
+
+webserverdeployment
```
scottslowe, line 74: Same here---this isn't exposed as a configuration setting.

scottslowe (changes requested):
Looks great overall, just a few minor changes needed. Thanks for putting this together!

scottslowe:
@gthuang Regarding the use of `serviceName`: we _could_ put in the text that users could port-forward to that Service for testing/verification. I can provide more details if you all think that might be helpful.

themes/default/content/templates/kubernetes/webapp/index.md
```diff
@@ -52,28 +52,18 @@ deploymentName
...
-
-```bash
-$ open $(pulumi stack output serviceName)
-```
+Output values like these are useful in many ways, most commonly as inputs for other stacks or related cloud resources.
 
 ## Customizing the project
 
-Projects created with the Kubernetes Application template expose the following [configuration](/docs/intro/concepts/config) settings:
+You can change the behavior of the Deployment by passing values to it and an example is already included in the template. Projects created with the Kubernetes Application template expose the following [configuration](/docs/intro/concepts/config) settings:
```
scottslowe, line 59: I don't think I agree with this addition as it isn't really accurate.

scottslowe (changes requested):
One minor nit regarding an addition, but LGTM otherwise

scottslowe (approved):

cnunciato (approved):

PR #2161, opened 2022-11-01
Update template helper

gthuang:
remove the rel refs

### DIFF
--- a/themes/default/archetypes/templates/template/index.md
+++ b/themes/default/archetypes/templates/template/index.md
@@ -31,7 +31,7 @@ cloud:
 # The content below is meant help you get started and to serve as a guide to work by. Feel free to adjust it needed for your template.
 ---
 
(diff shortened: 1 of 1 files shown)


cnunciato (commented):
Ah! Sorry about that. I thought I got these. Thanks for the fixes.

I added trailing slashes as well for the links, which we should generally do now that we aren't using `relref` (which added them for us automatically). They aren't technically required because the browser will figure things out, but when we do leave them out, the server responds with a 302 redirect that _does_ append a trailing slash -- so putting the trailing slash in just prevents the browser from having to do that. (So to an end user, things just feel a bit snappier.) Preventing that 302 is also slightly better for SEO.

At some point we'll add linting for this also, but for now, just something to keep in mind when you can remember it. Thank you!

cnunciato (approved):
@gthuang Looks good to me! Thanks for doing this. Feel free to merge when you like.

PR #2168, opened 2022-11-02
fix up static website imgs

susanev:
Screen Shot 2022-11-01 at 7 55 28 PM
Screen Shot 2022-11-01 at 7 55 21 PM
Screen Shot 2022-11-01 at 7 55 17 PM


susanev:
@anita-trimbur what do you think of these? do you have a better idea for how we could visually represent the static website components?

PR #2206, opened 2022-11-09
Add precommit hook for file size

sean1588:
This PR adds a file size check to the pre-commit hook to warn users about committing large files since we are starting to get close to hitting a size limit with our hugo build. See https://github.com/pulumi/docs/issues/8220#issuecomment-1309325148

I set this to warn for files greater than 1MB, but open to other suggestions. I only added this to warn for newly committed files as a pre-commit hook since this would continue fail for existing large files. Users can also get around the hook by committing with `--no-verify` in the event they really need to add such a large file.

image

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,3 +2,5 @@
 . "$(dirname -- "$0")/_/husky.sh"
 
 yarn run lint-staged
+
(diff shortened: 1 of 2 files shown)


susanev (approved):

nimbinatus (approved):
+1, LGTM!

PR #2211, opened 2022-11-10
reduce image size for animated .gif files and convert to .webp

aureq:
- ⏬ reduce image size for animated `.gif` files and convert to `.webp`
- ⬆️ update image references inside `.md` files accordingly
- ✅ some `.gif` haven't been converted because the `.webp` is bigger
- 🐛 related to https://github.com/pulumi/docs/issues/8220

## List of images that can't be optimized further (`.webp` > `.gif`)
theme/default/static/images/challenge/Step5.gif
theme/default/static/images/challenge/Step3.gif
theme/default/static/images/challenge/Step1.gif
theme/default/static/images/challenge/Step4.gif
theme/default/static/images/challenge/Step2.gif
theme/default/content/blog/kubernetes-server-side-apply/upsert.gif
theme/default/content/blog/organization-access-tokens/audit-logs-org-tokens.gif
theme/default/content/blog/automation-api/dbmigration.gif
theme/default/content/blog/automation-api/halloumi.gif
theme/default/content/blog/automation-api/http.gif
theme/default/content/blog/automation-api/debugging.gif
theme/default/content/blog/automation-api/vmgr_add.gif
theme/default/content/blog/pulumi-release-notes-74/scim.gif
theme/default/content/blog/building-your-first-serverless-app-using-only-javascript/code-completion.gif
theme/default/content/blog/refactoring-iac/refactor.gif
theme/default/content/blog/day-2-kubernetes-migrating-eks-nodegroups-with-zero-downtime/eks-migration.gif
theme/default/content/blog/day-2-kubernetes-migrating-eks-nodegroups-with-zero-downtime/k8s-cluster-workload.gif
theme/default/content/blog/getting-started-with-k8s-part2/kubectl-pod-status.gif
theme/default/content/blog/introducing-pulumi-registry/registry.gif
theme/default/content/blog/data-science-on-demand-spinning-up-a-wallaroo-cluster-is-easy-with-pulumi/tty-fast.gif

## List of images that have been optimized and converted to `.webp`
themes/default/content/blog/organization-access-tokens/nav-org-tokens.gif
themes/default/content/blog/announcing-enum-support/ts-enum.gif
themes/default/content/blog/announcing-enum-support/csharp-enum.gif
themes/default/content/blog/announcing-enum-support/python-enum.gif
themes/default/content/blog/announcing-enum-support/go-enum.gif
themes/default/content/blog/stack-readme/stack-readme.gif
[...]

PR #2212, opened 2022-11-10
Revert "reduce image size for animated .gif files and convert to .webp"

susanev:
Reverts pulumi/pulumi-hugo#2211


susanev:
admin merging; because this does not work on all machines

PR #2217, opened 2022-11-11
Add YAML to secrets concepts page

aq17:


themes/default/content/docs/intro/concepts/secrets.md
```diff
@@ -226,7 +226,7 @@ dbPassword                 [secret]
 
 Similarly, if our program attempts to print the value of `dbPassword` to the console-either intentionally or accidentally-Pulumi will mask it out:
 
-{{< chooser language "javascript,typescript,python,go,csharp,java" >}}
+{{< chooser language "javascript,typescript,python,go,csharp,yaml,java" >}}
```
iwahbe, line 229: We generally have the order https://github.com/pulumi/pulumi-hugo/blob/f46e8a6cbde8c5385f70015d9cd466b99d113c37/themes/default/content/docs/intro/concepts/config.md#L218

I think we should be consistent here.

iwahbe (approved):
1 small thing to fix. Otherwise LGTM

PR #2221, opened 2022-11-14
Add clarification on installation step

komalali:


blampe (approved):

susanev (commented):
@komalali could we make this a warning instead of a note?

PR #2223, opened 2022-11-14
Update the pricing page

zchase:


isaac-pulumi (approved):
Changed 15 members to 10 members in the comparison table

Issue #1604, opened 2022-06-09
Write a concept page for update plans


Frassle:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Write a concept page for update plans, explaining what they are, what the format of the files is, and how to use them in different cases such as manually or via CI/CD pipelines.

Frassle:
https://github.com/pulumi/home/issues/1684 might get folded into this

PR #2230, opened 2022-11-15
Update plans concept page

Frassle:
Fixes https://github.com/pulumi/pulumi-hugo/issues/1604

### DIFF
--- a/themes/default/content/docs/intro/concepts/plans.md
+++ b/themes/default/content/docs/intro/concepts/plans.md
@@ -1,0 +1,176 @@
+---
+title: "Plans"
+meta_desc: Learn about Pulumi update plans and how they can be used.
+menu:
+  intro:
+    parent: concepts
+    weight: 10
+
+aliases:
+  - /updateplans/
+---
+
+{{% notes type="warning" %}}
+Update plans are currently in experimental preview and will only show up in `--help` if the environment variable `PULUMI_EXPERIMENTAL` is set to `true`.
+{{% /notes %}}
+
+The result of previews can be saved to an _update plan_, this can then be used during a later update to
+constrain the update only to the operations that were planned.
+
+Constraining updates to plans may be useful in organizations with strict approval processes.
+
+## Generate an update plan
+
+To generate an update plan run `pulumi preview --save-plan=plan.json`. The plan will be written to the `plan.json` file.
+
+## Use an update plan
+
+To use an update plan run `pulumi up --plan=plan.json`, this will constrain the update to only the operations that were saved to the plan.
+
+## Limitations
+
+Update plans can only record information that is available at preview time. This means there are some program
+constructs that while possible to run correctly at update time, can't be recorded at preview time.
+
+The most common one is inputs set by the output of other resources that are being created. While the plan will
+record that the program wants to set the input field, it can't record the value which it will be set with and
+simply records `unknown`.
+
+The next most common issue is creating resources inside an [`apply`
+function](https://www.pulumi.com/docs/intro/concepts/inputs-outputs/#apply). If the value for the apply is
+unknown at preview time than the entire apply function will not run. This results in the plan not recording
+anything for the resources created inside the apply and reporting an error at update time when they try to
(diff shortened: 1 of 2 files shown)


themes/default/content/docs/intro/concepts/plans.md
```diff
@@ -0,0 +1,125 @@
+---
+title: "Plans"
+meta_desc: Learn about Pulumi update plans and how they can be used.
+menu:
+  intro:
+    parent: concepts
+    weight: 10
+---
```
justinvp, line 8: To get the shorter URL, I think it's just a matter of adding an alias here, like we do here:

https://github.com/pulumi/pulumi-hugo/blob/d7a8112c4f57cf0990982061d28b0b0f7dafb1aa/themes/default/content/docs/get-started/_index.md?plain=1#L11-L15

```
aliases:
  - /updateplans/

```

themes/default/content/docs/intro/concepts/plans.md
```diff
@@ -0,0 +1,134 @@
...
+
+aliases:
+  - /updateplans/
+---
+
+{{% notes type="warning" %}}
+Update plans are currently in experimental preview and will only show up in `--help` if the environment variable `PULUMI_EXPERIMENTAL` is set to `true`.
+{{% /notes %}}
+
+The result of previews can be saved to an _update plan_, this can then be used during a later update to constrain the update only to the operations that were planned.
```
jkodroff, line 17: Can we throw a sentence or two in here for reasons why people would want to use this feature - scenarios where this would be useful in practice? e.g. "Constraining an update to a plan may be useful in organizations with strict approval processes."

themes/default/content/docs/intro/concepts/plans.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Generate an update plan
+
+To generate an update plan run `pulumi preview --save-plan=plan.json`. The plan will be written to the `plan.json` file.
+
+## Use an update plan
```
jkodroff, line 23: We should also:

1. Clearly note the limitations (here or wherever it makes most sense to the author) of plan files, namely that they will not include resources created in an apply. 
2. Give some advice on how to handle this scenario, since unapplied changes in the main branch are not good. (Keep generating plans until there are no additional changes? Would we need to modify the GHA etc. to give a good workflow if people are using plan files?)

themes/default/content/docs/intro/concepts/plans.md
```diff
@@ -0,0 +1,134 @@
...
+
+To use an update plan run `pulumi up --plan=plan.json`, this will constrain the update to only the operations that were saved to the plan.
+
+## Format
+
+The format of update plans is currently experimental and subject to change. The most up to date description of the format is the [code](https://github.com/pulumi/pulumi/blob/master/sdk/go/common/apitype/plan.go).
```
jkodroff, line 29: If this includes deprecation in full, we should explicitly state this as well just so folks know how much they can reasonably expect to rely on the experimental feature. If we have an official policy on experimental feature non/support, it would be good to link here. If we don't have one, would be good to write one (when we have time).

Frassle: > If we have an official policy on experimental feature non/support, it would be good to link here.

Yeh that would be good! Raised https://github.com/pulumi/pulumi-hugo/issues/2239

themes/default/content/docs/intro/concepts/plans.md
```diff
@@ -0,0 +1,176 @@
...
+
+Constraining updates to plans may be useful in organizations with strict approval processes.
+
+## Generate an update plan
+
+To generate an update plan run `pulumi preview --save-plan=plan.json`. The plan will be written to the `plan.json` file.
+
+## Use an update plan
+
+To use an update plan run `pulumi up --plan=plan.json`, this will constrain the update to only the operations that were saved to the plan.
```
mikhailshilkov, line 28: Should we add a section that mentions the new CLI option?

PR #2235, opened 2022-11-15
Resize and optimize some GIFs

cnunciato:
This is the result of running Gifsicle on all GIFs greater than 5MB, resizing them to 800px wide (the width at which they're rendered in the blog) and running Gifsicle's built-in optimizer. Wins back about 120MB worth of space.

For reference, the command was: 

```bash
find themes/default/content/**/*.gif -type f -size +5M | xargs gifsicle --resize-width=800 --optimize=3 --batch
```

Also adds a bit to the blogging README showing how to do this for individual files going forward.

Fixes https://github.com/pulumi/docs/issues/8220 ... I think?

### DIFF
--- a/BLOGGING.md
+++ b/BLOGGING.md
@@ -162,7 +162,7 @@ For best results, we suggest the following specs for the `meta_image`, largely b
 
 Remember to replace the `meta_image` placeholder (or remove the property altogether and delete the placeholder `meta.png` file) before submitting your post.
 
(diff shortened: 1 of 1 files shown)


susanev:
might be worth having @meagancojocar do quality checks on a few of these to make sure they still display well.

zchase (approved):

cnunciato:
I spot-checked all of these and they all look pretty good to me, so merging for now. If something looks less than great, though, we can always come back and re-run Gifsicle on individual files with different settings.

PR #2240, opened 2022-11-16
Update plan prompt blog

Frassle:


themes/default/content/blog/experimental-update-plan-prompt/index.md
```diff
@@ -0,0 +1,46 @@
...
+  details
+```
+
+The `no` option is still the default choice, and the `yes` option is still immediately above it, so the change will hopefully align with user habits. However, there is a top option added, which suggests running an update with a plan built during the preview.
+
+Each interactive preview now calculates an Update Plan. If you choose the `yes, using Update Plans` option, the Update Plan will be automatically passed to the update command and applied during the deployment. Any inconsistencies between the preview and the update will be reported back to the user.
+
+## Try Update Plans
+
+Note! Update Plans are in public preview. We recommend using it only in non-critical, non-production scenarios during the preview period.
```
mikhailshilkov, line 42: Could you please format this as a note? https://github.com/pulumi/pulumi-hugo/blob/master/themes/default/content/blog/announcing-public-preview-update-plans/index.md?plain=1#L36

mikhailshilkov (approved):

PR #2242, opened 2022-11-16
Add team & org access tokens to pricing page

gthuang:
Adding team & org access tokens to the feature comparison table.

PR #2248, opened 2022-11-17
fix link failing pulumi/docs link check

susanev:
its currently failing on
>ERROR 2022/11/17 01:44:26 [en] REF_NOT_FOUND: Ref " /docs/intro/pulumi-service/organization-access-tokens/" from page "pricing/_index.md": page not found

### DIFF
--- a/themes/default/layouts/page/pricing.html
+++ b/themes/default/layouts/page/pricing.html
@@ -1467,7 +1467,7 @@
                 <div class="compare-plans-table-row-container">
                     <div class="compare-plans-table-row-item">
                         <span
-                            ><a class="compare-table-link" href="{{ relref . " /docs/intro/pulumi-service/organization-access-tokens/" }}"
+                            ><a class="compare-table-link" href="{{ relref . " /docs/intro/pulumi-service/organization-access-tokens" }}"
                                 >Organization and Team Access Tokens</a
                             ></span
                         >


themes/default/layouts/page/pricing.html
```diff
@@ -1467,7 +1467,7 @@ <h2 class="text-center">Compare editions</h2>
                 <div class="compare-plans-table-row-container">
                     <div class="compare-plans-table-row-item">
                         <span
-                            ><a class="compare-table-link" href="{{ relref . " /docs/intro/pulumi-service/organization-access-tokens/" }}"
+                            ><a class="compare-table-link" href="{{ relref . " /docs/intro/pulumi-service/organization-access-tokens" }}"
```
joeduffy, line 1470: Is the leading space intentional?

`" /docs/..."`

susanev: im not sure, they are like that throughout this page.

themes/default/layouts/page/pricing.html
```diff
@@ -1467,7 +1467,7 @@ <h2 class="text-center">Compare editions</h2>
                 <div class="compare-plans-table-row-container">
                     <div class="compare-plans-table-row-item">
                         <span
-                            ><a class="compare-table-link" href="{{ relref . " /docs/intro/pulumi-service/organization-access-tokens/" }}"
```
squaremo, line 1470: [This PR](https://github.com/pulumi/pulumi-hugo/pull/2242) has the trailing `/` but passes CI: https://github.com/pulumi/pulumi-hugo/actions/runs/3483734518/jobs/5827540138#step:7:60 (that's the place that the downstream CI fails). Is there a switch used there that's not used here, because here pages are expected to be missing, perhaps?

squaremo:
> its currently failing

More info: Looks like the failure is downstream https://github.com/pulumi/docs/actions/runs/3484599929/jobs/5829326415#step:9:1318. It is not obvious, but though there are some green checks since then, it hasn't succeeded at the _same_ workflow.

themes/default/layouts/page/pricing.html
squaremo: The relref here should fail linting too, since https://github.com/pulumi/pulumi-hugo/pull/2071, but is in layouts/ so isn't linted :-/

squaremo (approved):
I'm going to approve this, on the basis that trailing slashes in relref are a known problem, and this single, quick fix (probably) unblocks the site build. Thanks @susanev 🍎 

As a follow-up, I suggest changing the links here to plain links, and moving this page out of layout/ if possible. I'll file issues for those.

squaremo:
(I rewrote the commit message to include an explanation for the change. In general I wouldn't do that, but this PR plainly needed expediting.)

themes/default/layouts/page/pricing.html
squaremo: The leading spaces, and the newlines inside elements, look weird but don't affect the output -- links still work so far as I can see in the preview build.

At some point it would be good to er, regularise the markup in here (and elsewhere?). I think some of it is `yarn prettier` thinking that angle brackets dangling in space are better than long lines. The leading spaces might just be harmless copypasta.

squaremo:
This is the downstream PR which includes this fix as well as the content that had started to pile up: https://github.com/pulumi/docs/pull/8244.

PR #2249, opened 2022-11-17
Change advice on relref in line with #2701

squaremo:


README.md
```diff
@@ -67,13 +67,13 @@ This repository is a [Hugo module](https://gohugo.io/hugo-modules/) that doubles
...
 
-If you want to link to a page that exists on https://pulumi.com but not in this repository, just use the page's **relative path** with a [Hugo `relref`](https://gohugo.io/content-management/shortcodes/#ref-and-relref) in the usual way, and we'll make sure all links resolve properly at build-time. For example, to link to the [Digital Ocean Droplet](https://www.pulumi.com/docs/reference/pkg/digitalocean/droplet/) page (a page that doesn't exist in this repository but that would exist in an integration build), you'd use:
+If you want to link to a page that exists on https://pulumi.com but not in this repository, use the page's **relative path** in a Markdown or HTML link. For example, to link to the [Digital Ocean Droplet](https://www.pulumi.com/docs/reference/pkg/digitalocean/droplet/) page (a page that doesn't exist in this repository, but that would exist in an integration build), you'd use the Markdown
```
susanev, line 70: its not relative anymore is it?

susanev (approved):

README.md
squaremo: It's not a relative _path_ no that's true. But it wasn't before either -- the path in the example is exactly the same. The term used should be "path relative to the root". I'll correct that.

squaremo:
Thanks @susanev for reviewing 🎉  @cnunciato Would you also check you're happy with the language around (not using) relrefs -- thanks!

README.md
```diff
@@ -67,13 +67,13 @@ This repository is a [Hugo module](https://gohugo.io/hugo-modules/) that doubles
 
 Because of this, many of the links you follow when browsing around on the development server (to paths underneath `/docs/reference` for example) will fail to resolve because their content files are are checked into a different repository &mdash; most likely https://github.com/pulumi/docs. When we build the Pulumi website, we merge this module along with any others into a single build artifact, but when you're working within an individual module like this one, you may find you're unable to reach certain pages or verify the links you may want to make to them.
```
cnunciato, line 68:
```suggestion
Because of this, many of the links you follow when browsing around on the development server (to paths underneath `/docs/reference` for example) will fail to resolve because they originate from another repository &mdash; most likely https://github.com/pulumi/docs or https://github.com/pulumi/registry. When we build the Pulumi website, we merge this module along with any others into a single build artifact, but when you're working within an individual module like this one, you may find you're unable to reach certain pages or verify the links you may want to make to them.
```

README.md
```diff
@@ -67,13 +67,13 @@ This repository is a [Hugo module](https://gohugo.io/hugo-modules/) that doubles
...
 
-If you want to link to a page that exists on https://pulumi.com but not in this repository, just use the page's **relative path** with a [Hugo `relref`](https://gohugo.io/content-management/shortcodes/#ref-and-relref) in the usual way, and we'll make sure all links resolve properly at build-time. For example, to link to the [Digital Ocean Droplet](https://www.pulumi.com/docs/reference/pkg/digitalocean/droplet/) page (a page that doesn't exist in this repository but that would exist in an integration build), you'd use:
+If you want to link to a page that exists on https://pulumi.com but not in this repository, use the page's path **starting at the root** in a Markdown or HTML link. For example, to link to the [Digital Ocean Droplet](https://www.pulumi.com/docs/reference/pkg/digitalocean/droplet/) page (a page that doesn't exist in this repository, but that would exist in an integration build), you'd use the Markdown
```
cnunciato, line 70: Since the the example applies only to Markdown:

```suggestion
If you want to link to a page that exists on https://pulumi.com but not in this repository, use the page's path **starting at the root** in a Markdown or HTML link. For example, in a Markdown file, to link to the [Digital Ocean Droplet](https://www.pulumi.com/docs/reference/pkg/digitalocean/droplet/) page (a page that doesn't exist in this repository, but that would exist in an integration build), you'd use:
```

README.md
```diff
@@ -67,13 +67,13 @@ This repository is a [Hugo module](https://gohugo.io/hugo-modules/) that doubles
...
 
-```
-{{< relref /docs/reference/pkg/digitalocean/droplet >}}
+```markdown
+[Digital Ocean Droplet](/docs/reference/pkg/digitalocean/droplet)
```
cnunciato, line 73: I'd also add a trailing slash here (Hugo relrefs did this automatically): 

```suggestion
[Digital Ocean Droplet](/docs/reference/pkg/digitalocean/droplet/)
```

This is a seemingly small but relevant thing for SEO and overall user experience in that without it, the server will respond to requests for `/droplet` with a redirect to `/droplet/` (the path where `index.html` actually lives). Including the trailing avoids that redirect dance by linking directly to the content, which keeps us from getting dinged on SEO.

We don't yet lint for this, but we totally can (at least in Markdown files) and should. (It's mentioned in the linked PR as well.)

cnunciato (approved):
Looks good to me, just a couple of little-ish suggestions -- thanks!

README.md
squaremo: Isn't this exactly the thing that caused lots of downstream CI failures when using `relref`? No wonder you wanted to get rid of relrefs!

susanev: i think the issue with / and hugo is that hugo adds them itself as part of relref and so doesn't like when you add them; could be wrong here though. so without relref we arent automatically adding them anymore, so folks need to do it manually.

cnunciato: Yeah, this is indeed pretty confusing. susan has it right, but the nuance is that these are actually two different things that just happen to look similar.

With Markdown of course, when you write, `[foo](/bar/)`, whatever you put between the parens gets written verbatim into an HTML tag: `<a href="/bar/>foo</a>`. So to get the trailing slash in the rendered HTML, you have to put the slash into the Markdown expression yourself.

When you write `{{< relref "/bar/" >}}`, though, you're not crafting a link, you're telling Hugo, "Find the page named `/bar/` and then render a URL for it here" -- and what sucks is that because Hugo's lookup rules are esoteric and strange, _sometimes_ (depending on where that page lives in the hierarchy, how other pages are named, etc.), that lookup works, even with a trailing slash in the lookup expression. When it does work, Hugo finds the page and returns a link to it that appends includes a trailing slash. When it doesn't find the page, it blows up.

Hopefully that explains this weirdness a bit better. It'd be great to just stop using these altogether, even in layouts, as it'll make everything more portable in general, but yeah -- growing pains, I guess. 😄

PR #2254, opened 2022-11-17
Bump Go to 1.19

cnunciato:
Updates to 1.19 (current).


zchase (approved):
Do we also need to bump the version in `go.mod`? It is 1.16 right now

PR #2274, opened 2022-11-29
Update ECS examples for AWSx 1.0

danielrbradley:
Adds Go, Java and YAML as well as adjusting for argument changes.

Removed example that doesn't align to the text as the text was talking about the classic `fromDockerBuild` which doesn't exist any more.


justinvp (approved):
LGTM

PR #2277, opened 2022-11-29
Add Monica to team page

mnlumi:
Add Monica to team page.


themes/default/data/team/team/monica-rodriguez.toml
```diff
@@ -0,0 +1,11 @@
+id = "monica-rodriguez"
+name = "Monica Rodriguez"
+title = "Senior Product Manager"
+company = "Pulumi"
+weight = 1
```
mnlumi, line 5: Is weight needed? Not everyone has that in their file.

susanev:
@mnlumi ur image is over a 1 mb, it does not need to be that big. id be happy to push a commit to reduce the size if you'd like; otherwise pls reduce the image size, thank you!!

mnlumi:
> @mnlumi ur image is over a 1 mb, it does not need to be that big. id be happy to push a commit to reduce the size if you'd like; otherwise pls reduce the image size, thank you!!

@susanev I've reduced the size. Thanks for looking at this!

susanev (approved):
welcome to pulumi!! 🎉

PR #2279, opened 2022-11-30
Add layout for AWS serverless template

cnunciato:
Noticed this page was missing the `template` layout setting (so losing the left and right nav, breadcrumb, styling, etc.):

image

This should fix it up.


scottslowe (approved):
LGTM

PR #2280, opened 2022-11-30
NodeJS warning

Frassle:
See https://github.com/pulumi/pulumi/issues/11488

### DIFF
--- a/themes/default/content/docs/intro/languages/javascript.md
+++ b/themes/default/content/docs/intro/languages/javascript.md
@@ -14,6 +14,14 @@ aliases: ["/docs/reference/javascript/"]
 
 Pulumi supports writing your infrastructure as code in any JavaScript language running on Node.js using any of the [Current, Active and Maintenance LTS versions](https://nodejs.org/en/about/releases/).
 
+{{% notes type="warning" %}}
+While we expect _most_ uses of Pulumi to work on 19.2 and higher our function serialization doesn't currently handle those versions due to internal V8 changes.
+
+See [the GitHub issue](https://github.com/pulumi/pulumi/issues/11488) for tracking this.
+
+Due to this we're not currently testing on 19.2 and as such don't consider it fully supported.
+{{% /notes %}}
+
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/intro/languages/javascript.md
```diff
@@ -14,6 +14,14 @@ aliases: ["/docs/reference/javascript/"]
 
 Pulumi supports writing your infrastructure as code in any JavaScript language running on Node.js using any of the [Current, Active and Maintenance LTS versions](https://nodejs.org/en/about/releases/).
 
+{{% notes type="warning" %}}
+Pulumi currently does not support NodeJS 19.2 and higher due to internal V8 changes that our function serialization doesn't currently handle.
```
lukehoban, line 18: Is it actually true that we don't support it at all, or only that closure serialization does not work on 19.2?

Frassle: It's just closure serialisation, but we're not testing against 19.2 untill we can split off all the tests that use closure serialisation from everything else (and it's a lot because so many of the node tests use dynamic providers as part of the tests)

kpitzen (approved):

themes/default/content/docs/intro/languages/javascript.md
Frassle: I wonder if it would help if I flip these sentences around a bit? Say that we expect 19.2 to work except for function serialization, then link the issue, then comment that we're not currently testing 19.2 so we don't consider it fully supported?

kpitzen: I like that additional bit of nuance to this note - it feels like it re-centers the issue on function serialization and support guarantees re: testing rather than sounding like we don't think people should use 19,2 with Pulumi.

+1

susanev:
@Frassle can we merge this? or is it blocked on something?

Frassle:
I've rephrased the warning, if folk think the new wording is fine then this can be merged.

PR #2288, opened 2022-12-02
Update cluster-services.md

orcutt989:
``` 
pulumi stack output clusterName
error: current stack does not have output property 'clusterName'
```

```
pulumi stack output cluster_name
works!!
```

### DIFF
--- a/themes/default/content/docs/guides/crosswalk/kubernetes/cluster-services.md
+++ b/themes/default/content/docs/guides/crosswalk/kubernetes/cluster-services.md
@@ -96,7 +96,7 @@ navigate to the logs in your region, and look for the following group.
 The cluster name can be retrieved from the [cluster stack][crosswalk-control-plane] output.
 
 ```bash
-$ pulumi stack output clusterName
+$ pulumi stack output cluster_name
 ```
 
 ### Worker Nodes and Pods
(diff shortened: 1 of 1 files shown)


danielrbradley (commented):
Hi, thanks for looking to keep these docs up-to-date.

Where are you seeing that the output should be `cluster_name` rather than `clusterName`? Taking a quick read of this guide, it seems to be referencing the `kubernetes-guides`, `aws/aws/04...` example which has the output:

```
export const clusterName = config.clusterName
```

https://github.com/pulumi/kubernetes-guides/blob/master/aws/04-cluster-services/index.ts#L12

Is there a second code location which outputs a different export name somewhere?

orcutt989:
All I know is my stack doesnt have a "clusterName" and I don't believe I ever customized it.

scottslowe:
@orcutt989 Are you able to share the code you're using, so we can try to figure out why `cluster_name` works for you instead of `clusterName`? Thanks!

scottslowe:
@orcutt989 Yes, that's it.

orcutt989:
OK my bad please close it up.

susanev:
no worries!! glad you got it working!! ❤️

scottslowe:
@orcutt989 No worries! We'll go ahead and close this issue, but we'll look into perhaps adding some wording that clarifies the use of the name. Thank you for taking time to open this issue!

PR #2293, opened 2022-12-05
feat(dev-adv): decommission the developer advocacy page and offboard Laura and Kat

nimbinatus:
Please do not merge until after 9 Dec 2022.


scottslowe (approved):
Approving these changes, and will raise a separate issue to fix the behavior of the GitHub button (which just reloads the current page---this is the behavior on the production site right now as well as in this PR).

PR #2296, opened 2022-12-06
Don't terrify our users reading about an ancient Node version.

ringods:
Signed-off-by: Ringo De Smet <ringo@de-smet.name>


susanev (approved):
😬 to that pr title

PR #2297, opened 2022-12-06
Fix TypeScript example for creating a stack README

chrsmith:
If you use the TypeScript example code (just exporting `readFileSync(...)`) it returns a `Buffer` type, which gets exported as a byte array. In order for this to work correctly, you need to convert that `Buffer` into a string.

With this change the snippet now matches the [blog post](https://www.pulumi.com/blog/stack-readme/#step-1) announcing the feature.


susanev:
hiiiiii Chris!!!

susanev (approved):

PR #2308, opened 2022-12-09
fix kube cert logo

susanev:
i have no idea where the original logo came from but its super broken, and was displaying all sorts of poorly in firefox, and prolly safari too, this fixes it up

in firefox this is the change
Screen Shot 2022-12-09 at 11 38 16 AM


mnlumi (approved):
Looks good, thanks for fixing!

susanev:
cc/ @anita-trimbur it may be good to work checking logos in firefox and safari for new projects to prevent this moving forward

PR #2310, opened 2022-12-09
Add docs for resolving k8s SSA field conflicts

lblackstone:
Partial fix for https://github.com/pulumi/pulumi-kubernetes/issues/2258

### DIFF
--- a/themes/default/content/registry/packages/kubernetes/how-to-guides/managing-resources-with-server-side-apply.md
+++ b/themes/default/content/registry/packages/kubernetes/how-to-guides/managing-resources-with-server-side-apply.md
@@ -960,3 +960,264 @@ resources:
 {{% /choosable %}}
 
 {{< /chooser >}}
+
+## Handle Field Conflicts on Existing Resources
+
+After a resource is deployed to the cluster, it is possible for other controllers to start managing it. Thanks
+to the power of SSA, it is possible to see a detailed record of which controller last changed every field for
+the resource. An additional benefit of this fine-grained tracking is the ability to deliberately resolve conflicts
+rather than inadvertently overwriting changes made by another controller. Let's look at some common examples of field
+conflicts, and how to resolve them.
+
+### Cert Manager
+
+Cert-manager is a common example of an operator that mutates other cluster resources. Consider the following sequence
+of events:
+
+1. Deploy the cert-manager operator.
+2. Deploy a `ValidatingWebhookConfiguration` resource "X" using Pulumi with the `cert-manager.io/inject-ca-from`
+annotation set.
+3. The cert-manager operator updates "X" by setting the `webhooks[*].clientConfig.caBundle` field(s). This field is now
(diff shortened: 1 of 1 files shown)


themes/default/content/registry/packages/kubernetes/how-to-guides/managing-resources-with-server-side-apply.md
```diff
@@ -960,3 +960,264 @@ resources:
...
+
+{{< chooser language "typescript,python,go,csharp,java,yaml" >}}
+
+{{% choosable language javascript %}}
+
+{{% /choosable %}}
+{{% choosable language typescript %}}
+
+```typescript
+const res = new MyResource("res",
```
squaremo, line 996: Should `MyResource` be `ValidatingWebhookConfiguration` to match the prose description above, or have I misunderstood that description?

themes/default/content/registry/packages/kubernetes/how-to-guides/managing-resources-with-server-side-apply.md
```diff
@@ -960,3 +960,264 @@ resources:
...
+            Transformations: []yaml.Transformation{
+                // Ignore changes that will be overwritten by the kruise-manager deployment.
+                func(state map[string]interface{}, opts ...pulumi.ResourceOption) {
+                    if state["kind"] == "ValidatingWebhookConfiguration" ||
+						state["kind"] == "MutatingWebhookConfiguration" {
+						ignoreChanges := pulumi.IgnoreChanges([]string{
+                            "metadata.annotations.template",
+                            "webhooks[*].clientConfig",
+                        })
+                        opts = append(opts, ignoreChanges)
```
squaremo, line 1161: This assigns to an argument, which doesn't seem like it would have any effect outside the func.

squaremo (changes requested):
Requested change: I think the Go example of supplying a transformation won't work. I'm not sure it's possible to make it work, without fixing the kubernetes.yaml package, so you may need to say it's not possible for now.

themes/default/content/registry/packages/kubernetes/how-to-guides/managing-resources-with-server-side-apply.md
lblackstone: I was trying to leave the example a bit more generic since it applies to resources other than the `ValidatingWebhookConfiguration`, but perhaps that is confusing paired with the concrete example in the prose?

themes/default/content/registry/packages/kubernetes/how-to-guides/managing-resources-with-server-side-apply.md
lblackstone: Good catch. It looks like the example that I copied this from may not be working right either. I'm updating these examples now with something that works.

lblackstone:
@squaremo I updated all of the examples and tested them locally to make sure they work as expected.

justinvp (approved):

PR #2335, opened 2022-12-18
Change from getId() to id() in Java

Weinell:
bucket.getId() was deprecated - bucket.id() is the new method.


t0yv0 (approved):
Thank you!

PR #2347, opened 2022-12-21
Fix smart quotes in code snip

### LINKED ISSUE #2346: docs/blog: Smartifying of code snippet leads to it not working
### What happened?

There doesn't seem to be an issue type for reporting problems with the docs, so I'm filing this as a bug.

I was reading through https://www.pulumi.com/blog/disable-default-providers/ and tried the provided code sample to disable all default providers:
```
pulumi:disable-default-providers: [“*”]
```

This ended up not working because the code snippet there has the `"` and `"` transmogrified to `“` and `”`. This is hard to spot and lead to a bit of a rabbit hole until I tried the `pulumi config set` invocation on https://www.pulumi.com/docs/intro/concepts/resources/providers/#disabling-default-providers and `git diff` finally revealed the problem to me.

### Steps to reproduce

* Copy the line from the blog post into `Pulumi.yaml`.
* `pulumi preview` and watch how it still works without passing through explicit provider(s)

### Expected Behavior

Snippet from the blog works. It's the first result that shows up in a Google search, instead of the docs.

### Actual Behavior

It does not disable all default providers.

### Output of `pulumi about`

_No response_

### Additional context

There may be other places in the blog where this could be happening. It doesn't seem to happen in the longer code snippets, so I guess the single line is being treated differently? It might also have been that whoever authored that line got "helped" by their OS.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

cnunciato:
Fixes #2346.

### DIFF
--- a/themes/default/content/blog/disable-default-providers/index.md
+++ b/themes/default/content/blog/disable-default-providers/index.md
@@ -310,7 +310,7 @@ If you want to enforce that no resource should use a default provider, you would
 configuration file:
 
 ```yaml
-pulumi:disable-default-providers: [“*”]
+pulumi:disable-default-providers: ["*"]
 ```
 
 Now that you can [disable the default provider][dis-def-prov], you don’t have to worry about all of the possible unexpected consequences


scottslowe (approved):
LGTM

PR #2349, opened 2022-12-21
Add YAML to “Building with Pulumi” pathway

scottslowe:
This PR adds YAML support to all topics within the "Building with Pulumi" pathway.


cnunciato (approved):
Looks great! I didn't have time to walk through it locally, but it all looks correct. Thanks for doing this!

PR #2353, opened 2022-12-22
Add a Deploy with Pulumi button option to template pages

cnunciato:
Fixes https://github.com/pulumi/devrel-team/issues/439.

### DIFF
--- a/themes/default/layouts/shortcodes/templates/pulumi-new.html
+++ b/themes/default/layouts/shortcodes/templates/pulumi-new.html
@@ -20,13 +20,31 @@
 
 <!-- prettier-ignore -->
 {{ range $language := $languages }}
+{{ $templateName := (printf "%s-%s" $prefix $language) }}
 <div>
     <pulumi-choosable type="language" values="{{ $language }}">
         <div class="highlight">
             <pre tabindex="0" class="chroma"><code class="language-bash" data-lang="bash"><span class="line"><span class="cl">$ mkdir {{ $dirname }} <span class="o">&amp;&amp;</span> <span class="nb">cd</span> {{ $dirname }}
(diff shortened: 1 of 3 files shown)


susanev:
@cnunciato i appreciate us adding the deploy with pulumi buttons, thank you!!

would you be at all willing to put this in the right-panel? or below the top-page heading, so its above the fold?

seems a bit weird to me that we are suggesting this as an option after they run the pulumi new command, am i missing something?

cnunciato:
@susanev Totally open to any suggestions y'all have here, yep. I like the right-hand side thing for sure -- the only concern there would be making it clear this is kind of an either-or situation in that you either deploy with the CLI using "these commands here" _or_ with the Service using this button). Also relevant that the actual URL changes depending on the language selection, too, so that's a slight wrinkle to keep in mind as well. But yes -- any suggestions y'all have with those things in mind would be awesome and much appreciated!

cnunciato:
> is it possible to send them to the new project landing page instead

Doesn't look like the console supports that as of now (or at least I'm not seeing any URL params change when I switch between the various options on that page).

cnunciato:
To be honest, I'm very much on the fence about adding these buttons at this point, given the experience we have now, since you still end up returning to the CLI to run commands -- you just run slightly different commands. So it may actually make more sense to hold off and add them later -- but curious to hear others' thoughts on it.

scottslowe:
I like the current placement. I think it makes sense to me, given the workflow involved and such, and the wording ("Alternatively, you can...") seems pretty natural. I'd vote for the current placement and treatment.

aaronkao:
Good with the placement here. Ditto on Scott's comment on wording

scottslowe (approved):
LGTM

PR #2360, opened 2022-12-30
Add missing "chooser" around a couple of JS and TS snippets

kasadaamos:
The code snippets of JS and TS under https://www.pulumi.com/docs/intro/concepts/inputs-outputs/#lifting appeared both at the same time on the page because they were missing the wrapping of `{{< chooser...`


scottslowe (approved):
Thanks for the PR!

PR #2367, opened 2023-01-04
fix syntax errors in YAML

fschlz:
"FileAsset" is not accepted needs to be "fileAsset" "block" need to be "Block"


scottslowe (approved):
Thank you for your contribution!

PR #2379, opened 2023-01-09
add docs on disconnecting identity providers

susanev:


isabeldepapel (approved):
Thanks for adding this!

Do we maybe want to add a note that if they switch to email / password (e.g. Pulumi, not GitHub, etc) and they have members that have only ever logged in using SSO, those members will be locked out and they'll need to contact us to get them fixed? I think this has caused a little bit of confusion in the past

PR #2381, opened 2023-01-09
get-started/*/create-project: Add prompt examples

abhinav:
In the Get Started documentation,
when we tell users to Create a New Project with `pulumi new`,
we include a handful of short paragraphs next to each other
with instructions.

This is not very skimmable:

- keywords like project name/description/stack name
  are hidden in the text
- the actual instruction of what to do is also just hidden
  in the text

For someone following along, it's better to have keywords
related to prompts they're seeing highlighted,
and to include a sample output they can follow along to.

This changes the create-project sections
of all four Get Started with Pulumi pages
in the following way:

- project name and description are bold
- change "name of a stack" to "a stack name", and bold "stack name"
- bold out any provider-specific configuration prompts
- add the output of `pulumi new` after each step;
  I've altered the output slightly to not be language-specific

Before this change, only the k8s tutorial included a sample prompt,
but that was at the top of the text rather than interspersed with it.

---

**Help wanted**:
One thing I'm unsure about is the GCP prompt.
It's unclear whether the project name is required or optional.

### DIFF
--- a/themes/default/content/docs/get-started/aws/create-project.md
+++ b/themes/default/content/docs/get-started/aws/create-project.md
@@ -80,11 +80,34 @@ The [`pulumi new`](/docs/reference/cli/pulumi_new) command creates a new Pulumi 
 
 After logging in, the CLI will proceed with walking you through creating a new project.
 
-First, you will be asked for a project name and description. Hit `ENTER` to accept the default values or specify new values.
+First, you will be asked for a **project name** and **project description**. Hit `ENTER` to accept the default values or specify new values.
 
-Next, you will be asked for the name of a stack. Hit `ENTER` to accept the default value of `dev`.
+```
+This command will walk you through creating a new Pulumi project.
+
+Enter a value or leave blank to accept the (default), and press <ENTER>.
+Press ^C at any time to quit.
+
+project name: (quickstart)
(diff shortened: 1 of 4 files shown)


susanev (approved):
thank you for your contribution here!!

a few things, we have a style guide if you haven't seen it yet
https://github.com/pulumi/pulumi-hugo/blob/master/STYLE-GUIDE.md

we also currently use bold as a way to communicate the exact label in the ui or cli, which i think all of your changes do, but would be good to double-check that is the case before merging this.

abhinav:
Thanks for the style guide link!

> we also currently use bold as a way to communicate the exact label in the ui or cli, which i think all of your changes do, but would be good to double-check that is the case before merging this.

That makes sense.
`**description**` is not an exact match for "project description", but it's preceded by "project name."
So just "description" there may be fine, but if not, I'm happy to change it to "project name and project description."

The per-provider configuration (region/etc.) is not an exact match because the user facing label takes the form `$provider:$config` (e.g. `aws:region`) which differs from the friendly name "AWS Region". I can drop the bold from these.

susanev:
@abhinav yah if you can align project description id appreciate it; and dropping the bolding of the config sounds good, thank you!!

PR #2392, opened 2023-01-11
update iac/aws workshop for february

shughes26:
https://github.com/pulumi/marketing/issues/678

### DIFF
--- a/themes/default/content/resources/getting-started-with-iac-aws/index.md
+++ b/themes/default/content/resources/getting-started-with-iac-aws/index.md
@@ -7,7 +7,7 @@ meta_desc: "In this workshop, you will learn the fundamentals of Infrastructure 
 featured: false
 
 # If the video is pre-recorded or live.
-pre_recorded: true
+pre_recorded: false
 
 # If the video is part of the PulumiTV series. Setting this value to true will list the video in the "PulumiTV" section.
 pulumi_tv: false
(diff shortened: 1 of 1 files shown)


SaraDPH:
@prischoi, what time does it show to you? 

@shughes26 I see that the file says 9 AM PT, but it is not converting to the correct TZ to me. It should say 11 AM CST.
image

SaraDPH (approved):
I'm approving because the file is correct, although TZ conversion doesn't seem to be happening correctly.

prischoi:
> @prischoi, what time does it show to you?
> 
> @shughes26 I see that the file says 9 AM PT, but it is not converting to the correct TZ to me. It should say 11 AM CST. image

I don't see this workshop at all on the Resources page? The only one I see is on 2/14.
Screenshot 2023-01-11 at 7 46 35 PM

prischoi:
I'm getting 11 AM EST when pulling up the page which I assume is incorrect since it should be 12 since the correct time is 11 AM CST?

Screenshot 2023-01-12 at 9 45 46 AM

SaraDPH:
Yep, I should be seeing 11 AM, and you would see 12 PM. @shughes26 do you know how to fix the TZ conversion?

shughes26:
@prischoi @SaraDPH I'm looking into it

shughes26:
@prischoi @SaraDPH https://github.com/pulumi/pulumi-hugo/pull/2398 turns out I'd used the wrong offset because of daylight savings

PR #2404, opened 2023-01-17
shortcode/secret-new: Fix Go sample

abhinav:
The pulumi-secret-new refers to "NewSecret" for Go, which doesn't exist.
This likely meant to refer to "ToSecret", which other documentation refers to. Fix the short code and make it a link to the API reference.


abhinav:
Will combine this with another related change.

Issue #2382, opened 2023-01-09
build: Preview sometimes gets rate limited


abhinav:
### What happened?

I submitted a pull request and the preview job failed with the following error:

```
./scripts/ci/pull-request.sh
# ...
ERROR 2023/01/09 19:41:15 Failed to get JSON resource "https://api.github.com/repos/pulumi/docs/contents/content/docs/reference/cli/pulumi.md": Failed to retrieve remote file: Forbidden, body: "{\"message\":\"API rate limit exceeded for 40.86.28.130. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)\",\"documentation_url\":\"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting\"}\n"
```

### Steps to reproduce

Maybe create many pull requests at the same time?

### Expected Behavior

Job finishes successfully and posts a preview to the PR.

### Actual Behavior

Job fails and no preview is posted.

### Output of `pulumi about`

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

susanev:
im not really sure what we can do about this. @sean1588 @cnunciato thoughts?

abhinav:
Unauthenticated requests get rate limited more heavily.
If we provide a GitHub API token for whatever is making that request, the rate limit is a lot higher.

sean1588:
yeah, i notice i've been running into this a bit more often as well.

> If we provide a GitHub API token for whatever is making that request, the rate limit is a lot higher.

thanks @abhinav. If using a github api token to make the request would bump up the rate limit we should probably do that and I imagine it should be fairly trivial to do.

susanev:
k, adding it to the iteration if we end up having time, otherwise lets fix it up next iteration

abhinav:
👍 

https://github.com/pulumi/pulumi-hugo/blob/master/themes/default/layouts/shortcodes/pulumi-command.html appears to be the perpetrator.
The `getJSON` function used there [supports headers](https://gohugo.io/templates/data-templates/#add-http-headers).
[...]

PR #2406, opened 2023-01-17
Use github token if available for api calls

zchase:
Fixes: https://github.com/pulumi/pulumi-hugo/issues/2382

This PR adds the GITHUB_TOKEN if it is available to the API call.


themes/default/layouts/shortcodes/pulumi-command.html
```diff
@@ -1,5 +1,12 @@
 {{/* Fetch the current version of the Pulumi CLI doc." suffix. */}}
-{{- $json := getJSON "https://api.github.com/repos/pulumi/docs/contents/content/docs/reference/cli/pulumi.md" }}
+{{ $headers := dict }}
+{{ $gitHubToken := os.Getenv "GITHUB_TOKEN" }}
+{{ if ne $gitHubToken "" }}
+    {{ $token := printf "Bearer %s" $gitHubToken }}
```
abhinav, line 5: optional suggestion: You can shorten this with `with` if you'd like:

```suggestion
{{ with (os.Getenv "GITHUB_TOKEN") }}
    {{ $token := printf "Bearer %s" . }}
```

Empty string evaluates to a non-truthy value in Go's template system.

abhinav (approved):

PR #2420, opened 2023-01-19
Add a line about using the right logo in blog posts

cnunciato:
Adds a line to BLOGGING.md to verify that logos are current. (Request from @SaraDPH.)


scottslowe (approved):
LGTM. No need to block on this, but should we provide more information on exactly where to find the correct image?

cnunciato:
> LGTM. No need to block on this, but should we provide more information on exactly where to find the correct image?

Good callout, I'll add a note about asking in Slack about that as well.

PR #2422, opened 2023-01-19
Fix bug where registry filter page shows previous input when navigating back

kimberleyamackenzie:
Fixes https://github.com/pulumi/registry/issues/1824

I was stumped on this for a bit, but it looks like the browser was trying to be helpful by autofilling the previous search term in the input when navigating back, so simply turning off autocomplete solves the issue.


susanev (approved):

sean1588 (approved):
:ship: :shipit: :rocket:

PR #2425, opened 2023-01-20
[_index.md] CDK assets are uploaded to appropriate services

ilyash-b:
... and not to CloudFormation (which was my  understanding of previous wording)


themes/default/content/docs/intro/vs/cloud-template-transpilers/aws-cdk/_index.md
```diff
@@ -31,7 +31,7 @@ Both Pulumi and CDK allow you to build and deploy infrastructure on AWS using fa
...
+First, CDK supports only AWS, whereas Pulumi supports over 60 cloud and SaaS providers, with more being added all the time. Second, while both Pulumi and CDK support languages like TypeScript, Python, Go, C#, and Java, the two tools are fundamentally different in how they interpret these languages and deploy resources. Unlike Pulumi, whose open-source engine understands these languages and communicates directly with cloud providers to deploy infrastructure, CDK is a transpiler (i.e., a [source-to-source compiler](https://en.wikipedia.org/wiki/Source-to-source_compiler)) that produces AWS Cloud Assembly, an intermediate format consisting of CloudFormation templates and assets. After the assets are uploaded to appropriate services (S3, ECR), CDK uses CloudFormation service, which acts as the deployment engine for provisioning resources.
```
cnunciato, line 34: "Assets" by itself sounds like it refers to something CloudFormation-specific, but my understanding is that in the context of CDK, "assets" essentially means "various other files", so I'd suggest keeping the "other" to eliminate that ambiguity. 

```suggestion
First, CDK supports only AWS, whereas Pulumi supports over 60 cloud and SaaS providers, with more being added all the time. Second, while both Pulumi and CDK support languages like TypeScript, Python, Go, C#, and Java, the two tools are fundamentally different in how they interpret these languages and deploy resources. Unlike Pulumi, whose open-source engine understands these languages and communicates directly with cloud providers to deploy infrastructure, CDK is a transpiler (i.e., a [source-to-source compiler](https://en.wikipedia.org/wiki/Source-to-source_compiler)) that produces AWS Cloud Assembly, an intermediate format consisting of CloudFormation templates and other assets. After the assets are uploaded to the appropriate services (S3, ECR), CDK uses the CloudFormation service, which acts as the deployment engine for provisioning resources.
```

cnunciato (commented):
Thanks for this clarification! Couple of tiny things I'd suggest, otherwise 👍.

cnunciato:
Hi @ilyash-b, just checking in on this. Would you like to make these changes? If you prefer, we could also commit them for you and then merge your PR. Up to you.

ilyash-b:
Looks good. Please go ahead on your side!

Regards,
Ilya
________________________________

cnunciato:
Thanks again @ilyash-b!

PR #2429, opened 2023-01-23
Configure figma design tokens

kimberleyamackenzie:
This PR will allow us to configure Figma design tokens, so that folks using Figma can have 1) easy access to the design tokens currently checked into source control, and 2) the ability to create PRs from Figma to update these tokens as needed.  This will help us keep designs + prod code in sync better and allow the design team to more easily make theme changes.

- [ ] Add a `tokens.json` file, which is what Figma considers the source of truth for design tokens
- [ ] Update our Tailwind config to read from the tokens file, so that can be the single source of truth
- [ ] There should be no user-facing impacts to this - basically in the preview, everything should look the exact same.  Please give a shout if you see any styles (particularly fonts and colors) that look unexpected or broken.


zchase (approved):
LGTM --- How is the tokens file update? Manually?

kimberleyamackenzie:
@zchase yeah the tokens file can just get updated manually, and figma is connected to it so that folks can generate PRs to do so straight from figma (or locally, both ways work).  i'll post some docs on how the figma part works after this merges.

anita-trimbur (approved):
i went through a bunch of pages and couldn't find anything! lgtm :)

PR #2214, opened 2022-11-10
Update title tags for docs pages

zchase:
This PR contains page-level optimizations for sections within Pulumi Docs (i.e. Intro to Pulumi, Tutorials, Support, etc) as part of on-page SEO efforts to improve DX on organic search.

Original: https://github.com/pulumi/pulumi-hugo/pull/2207

PR #2436, opened 2023-01-24
Shorten language nav items

cnunciato:
This change makes a few tweaks to the frontmatter params of our language pages in order to shorten the items in the TOC navigation and make the titles themselves a bit more consistent. 

Right now, these list items are pretty verbose, so they aren't quite as amenable to scanning as they used to be. With this change, we should be able to keep the desired SEO changes introduced with https://github.com/pulumi/pulumi-hugo/pull/2214 (as page titles and H1s won't be affected), but restore the navigation to its previous state.

Before:

image

After:

image

Hopefully this is acceptable; if the current UX is intentional, and it'd be significantly better to keep the longer titles in the navigation as well, let me know and I'll close. But this does seem like a better overall UX, and hopefully strikes a good  balance between UX and SEO.

### DIFF
--- a/themes/default/content/docs/intro/languages/dotnet.md
+++ b/themes/default/content/docs/intro/languages/dotnet.md
@@ -1,5 +1,7 @@
 ---
-title: "How to Use .NET (C#, VB, F#) for Pulumi"
+title: .NET (C#, VB, F#)
+title_tag: How to use .NET (C#, VB, F#) with Pulumi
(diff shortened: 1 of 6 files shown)


susanev:
i had a pr for this, but it sounded like Zack didn't want it. im supportive, but will leave it to yall to figure out.

scottslowe (approved):
LGTM

zchase (approved):

PR #2441, opened 2023-01-26
Updates team content for recent changes

kpitzen:


themes/default/data/team/team/paul-stack.toml
```diff
@@ -1,10 +0,0 @@
-id = "paul-stack"
-name = "Paul Stack"
-title = "Engineering Manager"
-weight = 1
-
-status = "inactive"
```
justinvp, line 6: I may not be up-to-date on how the current system works, but I thought we keep these files around but as "inactive" for cases like old blog posts authored by the inactive person.

kpitzen: Yeah, I think that's correct

justinvp (approved):

PR #2454, opened 2023-02-01
Fix a link in the Redshift post

cnunciato:
The post should link to the next post in the series, but instead links to itself.


scottslowe (approved):
Good to go

PR #2456, opened 2023-02-01
Clarify how to retrieve namespaced configuration

cnunciato:
Users often ask how to retrieve namespaced configuration values like `aws:region` and the like. This change adds some examples that show how to do that.


scottslowe (approved):
LGTM

PR #2459, opened 2023-02-01
Add dotnet to some pages

Frassle:


themes/default/content/learn/building-with-pulumi/stack-outputs/index.md
```diff
@@ -56,6 +56,19 @@ Note that you'll want to insert this just before the last `return nil` statement
 
 {{% /choosable %}}
 
+{{% choosable language csharp %}}
+
+```csharp
+return new Dictionary<string, object?>
+{
+    ["url"] = pulumi.Output.Format($"http://localhost:{frontendPort}")
```
Zaid-Ajaj, line 64: This should be just `Output.Format($"http://localhost:{frontendPort}")`, right? the dotnet SDK doesn't have a lower-case `pulumi` namespace

Frassle: herp yes, too much copy paste

Zaid-Ajaj (approved):

themes/default/content/learn/building-with-pulumi/stack-outputs/index.md
```diff
@@ -56,6 +56,19 @@ Note that you'll want to insert this just before the last `return nil` statement
...
+{{% choosable language csharp %}}
+
+```csharp
+return new Dictionary<string, object?>
+{
+    ["url"] = Output.Format($"http://localhost:{frontendPort}")
+};
+```
+
+Note that you'll want to insert this just before the last `return nil` statement.
```
Frassle, line 68: Op this needs fixing

PR #2478, opened 2023-02-07
cd/github: Fix language on code fences

abhinav:
A few code fences on the GitHub Actions page use 'diff'
even though they contain YAML.

This switches those fences to YAML.


susanev (approved):
thank you!

PR #2534, opened 2023-03-07
Remove dead links, handle missing URLs on the About page

cnunciato:
A couple of awards URLs are no longer valid and don't seem to have good replacements. This change removes them and updates the page template to handle records with no URLs. (The Awards page already seems to handle this.)


susanev (approved):
thank you!!

PR #2535, opened 2023-03-07
Document support for Gitlab releases to host plugin binaries

ringods:
Added the documentation for the `gitlab://` support for plugin hosting.

Implemented in https://github.com/pulumi/pulumi/releases/tag/v3.56.0


Frassle (approved):
Thanks @ringods!

PR #2554, opened 2023-03-10
update title tags to replace slashes with dots for registry pages

sean1588:
issue: https://github.com/pulumi/registry/issues/2184

### DIFF
--- a/themes/default/layouts/partials/head.html
+++ b/themes/default/layouts/partials/head.html
@@ -153,7 +153,7 @@
         {{ else if eq .Type "docs" }}
             <title>{{ $title }} | Pulumi Docs</title>
         {{ else if eq .Type "registry" }}
-            <title>{{ $title }} | Pulumi Registry</title>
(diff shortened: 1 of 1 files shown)


themes/default/layouts/partials/head.html
```diff
@@ -153,7 +153,7 @@
         {{ else if eq .Type "docs" }}
             <title>{{ $title }} | Pulumi Docs</title>
         {{ else if eq .Type "registry" }}
-            <title>{{ $title }} | Pulumi Registry</title>
+            <title>{{ replace $title "/" "." }} | Pulumi Registry</title>
```
susanev, line 156: google is cool with dots?

susanev (approved):

themes/default/layouts/partials/head.html
sean1588: i'm guessing it is since the resources in the other packages have dots in the title.

PR #2562, opened 2023-03-13
Add image requirement

komalali:


themes/default/content/docs/reference/deployments-rest-api/_index.md
```diff
@@ -91,6 +91,7 @@ The executor context defines information about the executor where the Pulumi ope
 
 Image requirements:
 
+* It must be a publicly available image.
```
pgavlin, line 94: Is this accurate? Like we can't use an ECR repo if there are associated--oh, do we not have anywhere to fill in the creds? The settings have space for them; maybe the UI doesn't...

susanev (approved):
@komalali my approval is mostly a stamp.

let me know if you need any help with this or anything.

themes/default/content/docs/reference/deployments-rest-api/_index.md
komalali: Yeh I don't believe we have anywhere to fill in the creds - does the settings API have accomadation for this? I've never seen it in the code.

pgavlin: > does the settings API have accomadation for this? I've never seen it in the code.

Well, I thought it did, but I was wrong. We can accommodate this in the workflow bits but it's not exposed by the deployment settings API. We should fix that.

PR #2571, opened 2023-03-16
Sarah/resources cleanup

shughes26:
Automatically remove workshops from the "upcoming" tab when they've occurred more than 24 hours in the past, and replace the registration form with "Recording coming soon"

This doesn't remove them from the "all" tab, so we can still see which workshops need videos added


SaraDPH (approved):

zchase (approved):
LGTM!

PR #2577, opened 2023-03-20
Make Zephyr ToC entries consistent

scottslowe:
This PR makes the ToC (table of contents) entries for the Zephyr posts consistent. It ensures that the ToC is a bulleted list in all three posts, and updates the hyperlinks accordingly to account for the publication of the third post.


cnunciato (approved):
⭐

PR #2581, opened 2023-03-21
Add rss feed for events

sean1588:
Adds the rss feed to surface events sorted by date desc.
image

### DIFF
--- a/themes/default/content/resources/_index.md
+++ b/themes/default/content/resources/_index.md
@@ -27,4 +27,6 @@ sections:
     - label: Whitepapers
       anchor: whitepapers
       icon: book
+
+outputs: ["html", "rss"]
 ---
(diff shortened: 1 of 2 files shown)


cnunciato (commented):
Nice! This is awesome.

I have two questions, both because of my somewhat-ignorance of the details of RSS. 

One is whether we should use at least something, like the event URL, for the `guid` field, as we do with the blog. This would (I think) allow us to update RSS items after we've served them to clients for the first time (to support things like corrections, etc.). 

Second is around dates. Do you know whether there's any RSS-standard way to handle things like this that are published before or after the dates of the events they refer to? I only ask because it's somewhat strange to see posts in my reader with future dates, like Pulumi Up.

sean1588:
thanks @cnunciato. yeah good call on adding the guid field. 

As far as the dates are concerned, i'm not including a `<pubDate>` field in the items list, just because we don't have any data around when it was actually published and yes would probably look strange in a reader. So I am using a field called `<date>`, just to indicate the date the event is occuring on, rather than when we added the event to the site. Though I am wondering what a reader will do with the date field since it is not part of the official specification. I am wondering if they will just render fields outside the  spec somehow...? I'll do a bit more digging and testing around this.

sean1588:
@cnunciato, I think I'm just going to have to go with the pubdate here. I think it may look weird in readers, but I don't think we have really any other options here that the specification would allow for.

cnunciato (approved):
@sean1588 Sounds good, and yep, I agree that's probably the best option for now, assuming we want to go with RSS. Thanks!

PR #2584, opened 2023-03-22
Update Docker post with video

aaronkao:


mnlumi:
@aaronkao Should there be any description/context around the youtube video link? Something like "Here's a walk-through of creating an image".

aaronkao:
Ok, just added some text

PR #2586, opened 2023-03-22
`[docs/intro/concepts/inputs-outputs]` document JSON output helpers

### LINKED ISSUE #2556: Update Output docs to mention the jsonStringify and jsonParse helpers
Update the Output docs at https://www.pulumi.com/docs/intro/concepts/inputs-outputs/ to discuss the new jsonStringify and jsonParse helpers with inline examples for each language.

Perhaps a new section like "Working with Outputs and JSON" similar to the existing "Working with Outputs and String"?

https://github.com/pulumi/pulumi-hugo/blob/master/themes/default/content/docs/intro/concepts/inputs-outputs.md

dixler:
This PR adds documentation for the JSON stringify/parse methods added to the SDKs.

Fixes #2556

### DIFF
--- a/themes/default/content/docs/intro/concepts/inputs-outputs.md
+++ b/themes/default/content/docs/intro/concepts/inputs-outputs.md
@@ -655,13 +655,12 @@ variables:
 
 Often in the course of working with web technologies, you encounter JavaScript Object Notation (JSON) which is a popular specification for representing data. Pulumi provides first-class helper functions for deserializing JSON string outputs into your language's native objects and serializing your language's native objects to JSON string outputs.
 
+### Converting Outputs to JSON
+
+You can natively represent the definition of an AWS Step Function State Machine and embed outputs from other resources then convert it to a JSON string.
+
 {{< chooser language "javascript,typescript,python,go,csharp" >}}
-
 {{% choosable language javascript %}}
-
-### Example Converting Outputs to JSON
-
-The following example allows you to natively represent the definition of an AWS Step Function State Machine and easily embed outputs from other resources in your stack.
 
 ```javascript
 const stateMachine = new awsnative.stepfunctions.StateMachine("stateMachine", {
@@ -673,12 +672,12 @@ const stateMachine = new awsnative.stepfunctions.StateMachine("stateMachine", {
         "States": {
             "Hello": {
                 "Type": "Task",
-                "Resource": helloFunction.arn,
+                "Resource": helloFunction.arn, // Pulumi Resource Output
                 "Next": "World",
             },
             "World": {
                 "Type": "Task",
-                "Resource": worldFunction.arn,
+                "Resource": worldFunction.arn, // Pulumi Resource Output
                 "End": true,
             },
         },
@@ -686,9 +685,123 @@ const stateMachine = new awsnative.stepfunctions.StateMachine("stateMachine", {
 });
 ```
 
-### Example Parsing JSON Outputs to Objects
+{{% /choosable %}}
+{{% choosable language typescript %}}
 
-The following is an example of Pulumi code that operates on a JSON object and removes all of the policy statements by replacing it with an empty list.
+```typescript
+const stateMachine = new awsnative.stepfunctions.StateMachine("stateMachine", {
+    roleArn: sfnRole.arn,
+    stateMachineType: "EXPRESS",
+    definitionString: pulumi.jsonStringify({
+        "Comment": "A Hello World example of the Amazon States Language using two AWS Lambda Functions",
+        "StartAt": "Hello",
+        "States": {
+            "Hello": {
+                "Type": "Task",
+                "Resource": helloFunction.arn, // Pulumi Resource Output
+                "Next": "World",
+            },
+            "World": {
+                "Type": "Task",
+                "Resource": worldFunction.arn, // Pulumi Resource Output
+                "End": true,
+            },
+        },
+    })
+});
+```
+
+{{% /choosable %}}
+{{% choosable language python %}}
+
+```python
+state_machine = aws_native.stepfunctions.StateMachine("stateMachine",
+    role_arn=sfn_role.arn,
+    state_machine_type="EXPRESS",
+    definition_string=pulumi.Output.json_dumps({
+        "Comment": "A Hello World example of the Amazon States Language using two AWS Lambda Functions",
+        "StartAt": "Hello",
+        "States": {
+            "Hello": {
+                "Type": "Task",
+                "Resource": hello_function.arn, # Pulumi Resource Output
+                "Next": "World",
+            },
+            "World": {
+                "Type": "Task",
+                "Resource": world_function.arn, # Pulumi Resource Output
+                "End": True,
+            },
+        },
+    })
+)
+```
+
+{{% /choosable %}}
+{{% choosable language go %}}
+
+{{% notes type="info" %}}
+The Pulumi Go SDK does not currently support serializing or deserializing maps with unknown values.
+It is tracked [here](https://github.com/pulumi/pulumi/issues/12460).
+
+The following is a simplified example of using `pulumi.JSONMarshal` in Go.
+{{% /notes %}}
+
+```go
+pulumi.JSONMarshal(pulumi.ToMapOutput(map[string]pulumi.Output{
+    "bool": pulumi.ToOutput(true),
+    "int":  pulumi.ToOutput(1),
+    "str":  pulumi.ToOutput("hello"),
+    "arr": pulumi.ToArrayOutput([]pulumi.Output{
+        pulumi.ToOutput(false),
+        pulumi.ToOutput(1.0),
+        pulumi.ToOutput(""),
+        pulumi.ToMapOutput(map[string]pulumi.Output{
+            "key": pulumi.ToOutput("value"),
+        }),
+    }),
+    "map": pulumi.ToMapOutput(map[string]pulumi.Output{
+        "key": pulumi.ToOutput("value"),
+    }),
+    // The following functionality is currently unsupported as myResource
+    // is an unknown value.
+    "unknown": myResource.ApplyT(func(res interface{}) (interface{}, error) {
+        return "Hello World!", nil
+    }),
+}))
+```
+
+{{% /choosable %}}
+{{% choosable language csharp %}}
+
+```csharp
+var stateMachine = Pulumi.Output.JsonSerialize(Output.Create(new {
+        Comment = "A Hello World example of the Amazon States Language using two AWS Lambda Functions",
+        StartAt = "Hello",
+        States = new Dictionary<string, object?>{
+        ["Hello"] = new {
+            Type = "Task",
+            Resource = helloFunction.Arn, // Pulumi Resource Output
+            Next = "World",
+        },
+        ["World"] = new {
+            Type = "Task",
+            Resource = worldFunction.Arn, // Pulumi Resource Output
+            End = true,
+        },
+    },
+}));
+```
+
+{{% /choosable %}}
+{{< /chooser >}}
+
+### Parsing JSON Outputs to Objects
+
+You can parse a JSON string into an object and then, inside of an Apply, manipulate the object to remove all of the policy statements.
+
+{{< chooser language "javascript,typescript,python,go,csharp" >}}
+{{% choosable language javascript %}}
 
 ```javascript
 const jsonIAMPolicy = pulumi.output(`{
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/intro/concepts/inputs-outputs.md
```diff
@@ -651,6 +651,217 @@ variables:
 
 {{% /choosable %}}
 
+## Working with Outputs and JSON {#outputs-and-json}
+
+Often in the course of working with web technologies, you encounter JavaScript Object Notation (JSON) which is a popular specification for representing data. Pulumi provides first-class helper functions for working with deserializing JSON string outputs into your language's native objects and serializing your language's native objects to JSON string outputs.
+
+Outputs that contain strings cannot be used directly in operations such as string concatenation. String interpolation lets you more easily build a string out of various output values, without needing {{< pulumi-apply >}} or [Output.all](/docs/reference/pkg/python/pulumi#pulumi.Output.all). You can use string interpolation to export a stack output, provide a dynamically computed string as a new resource argument, or even for diagnostic purposes.
+
+The following is an example of pulumi code that operates on a JSON object and removes all of the policy satements by setting it to an empty list.
```
ericrudder, line 660:
```suggestion
The following is an example of Pulumi code that operates on a JSON object and removes all of the policy satements by setting it to an empty list.
```

themes/default/content/docs/intro/concepts/inputs-outputs.md
```diff
@@ -651,6 +651,217 @@ variables:
...
+                "Type": "Task",
+                "Resource": worldFunction.arn,
+                "End": true,
+            },
+        },
+    })
+});
+```
+
+### Example Parsing JSON outputs to Objects
```
ericrudder, line 691:
```suggestion
### Example Parsing JSON Outputs to Objects
```

themes/default/content/docs/intro/concepts/inputs-outputs.md
```diff
@@ -651,6 +651,217 @@ variables:
...
```
ericrudder, line 752:
```suggestion
### Example Parsing JSON Outputs to Objects
```

themes/default/content/docs/intro/concepts/inputs-outputs.md
```diff
@@ -651,6 +651,217 @@ variables:
...
+                "Type": "Task",
+                "Resource": world_function.arn,
+                "End": True,
+            },
+        },
+    })
+)
+```
+
+### Example Parsing JSON outputs to Objects
```
ericrudder, line 792:
```suggestion
### Example Parsing JSON Outputs to Objects
```

themes/default/content/docs/intro/concepts/inputs-outputs.md
```diff
@@ -651,6 +651,217 @@ variables:
...
+{{% /choosable %}}
+{{% choosable language csharp %}}
+
+### Example Converting Outputs to JSON
+
+```go
+pulumi.JSONMarshal(outputValue)
+```
+
+### Example Parsing JSON outputs to Objects
```
ericrudder, line 853:
```suggestion
### Example Parsing JSON Outputs to Objects
```

themes/default/content/docs/intro/concepts/inputs-outputs.md
```diff
@@ -651,6 +651,327 @@ variables:
...
+{
+    // delete the policy statements.
+    policy.Statement = Pulumi.Output.Create(new List<StatementEntry> { });
+    return policy;
+})
+```
+
+For more details view the .NET documentation.
+
+{{% /choosable %}}
```
dixler, line 971: @Zaid-Ajaj @Frassle does this look idiomatic?

themes/default/content/docs/intro/concepts/inputs-outputs.md
```diff
@@ -651,6 +651,327 @@ variables:
...
+{{% choosable language go %}}
+
+### Example Converting Outputs to JSON
+
+{{% notes type="info" %}}
+The Pulumi Go SDK does not currently support Marshalling and Unmarshalling Maps with unknowns.
+
+This is being tracked [here](https://github.com/pulumi/pulumi/issues/12460).
+{{% /notes %}}
```
dixler, line 845: @Frassle is this a good summary of the current state of Go?

themes/default/content/docs/intro/concepts/inputs-outputs.md
```diff
@@ -651,6 +651,327 @@ variables:
...
+            "key": pulumi.ToOutput("value"),
+        }),
+    }),
+    "map": pulumi.ToMapOutput(map[string]pulumi.Output{
+        "key": pulumi.ToOutput("value"),
+    }),
+    // currently unsupported as myResource is unknown
+    "unknown": myResource.ApplyT(func(res interface{}) (interface{}, error) { return "Hello World!", nil }),
+}))
+```
```
dixler, line 868: I feel that this example could be more practical as I think people only work with `outputs` because the values are unknown.

themes/default/content/docs/intro/concepts/inputs-outputs.md
```diff
@@ -651,6 +651,327 @@ variables:
...
 
+## Working with Outputs and JSON {#outputs-and-json}
+
+Often in the course of working with web technologies, you encounter JavaScript Object Notation (JSON) which is a popular specification for representing data. Pulumi provides first-class helper functions for deserializing JSON string outputs into your language's native objects and serializing your language's native objects to JSON string outputs.
+
+{{< chooser language "javascript,typescript,python,go,csharp" >}}
+
+{{% choosable language javascript %}}
+
+### Example Converting Outputs to JSON
```
abhinav, line 662: We shouldn't put each example under a header like this.
It pollutes the TOC on the right:
image

It may be better to make these bold instead

```markdown
**Converting Outputs to JSON**
```

themes/default/content/docs/intro/concepts/inputs-outputs.md
```diff
@@ -651,6 +651,327 @@ variables:
...
+```
+
+For more information view the NodeJS documentation.
+
+{{% /choosable %}}
+{{% choosable language python %}}
+
+### Example Converting Outputs to JSON
+
+The following example allows you to natively represent the definition of an AWS Step Function State Machine and easily embed outputs from other resources in your stack.
```
abhinav, line 773: Repeating the header and the description between each chooser will become very hard to maintain.
Consider instead <header> <chooser with examples> <header> <chooser with examples>

```markdown
**Example converting Outputs to JSON**

{{< chooser ... >}}
...
{{< /chooser ... >}}

**Example parsing JSON Outputs to an object**

{{< chooser ... >}}
...
{{< /chooser ... >}}
```

At that point, if you wanted to go back to using headers, you could give them simpler/shorter names:

```markdown
### Serializing Outputs to JSON

{{< chooser ... >}}
...
{{< /chooser ... >}}

### Deserializing JSON Outputs

{{< chooser ... >}}
...
{{< /chooser ... >}}
```

abhinav (commented):
Haven't looked at the content yet but we should reconsider the structure a bit.
It's not rendering as intended.

themes/default/content/docs/intro/concepts/inputs-outputs.md
susanev: unfortunately, we should not use bold in place of a heading; we can instead have an h3 of Examples which will show up in the table of contents, then you can use h4s from all the various examples and they wont populate in the table of contents if that's concerning to yall.

susanev: ref: https://usability.yale.edu/web-accessibility/articles/headings#:~:text=Do%20not%20use%20bold%20instead,reader%20users%20will%20not%20benefit.

themes/default/content/docs/intro/concepts/inputs-outputs.md
Frassle: It's not just unknowns. Any nested outputs I think will trigger an error in Marshal because all output types are unmarshallable.

themes/default/content/docs/intro/concepts/inputs-outputs.md
```diff
@@ -651,6 +651,327 @@ variables:
...
+
+```csharp
+var stateMachine = Pulumi.Output.JsonSerialize(Output.Create(new {
+        Comment = "A Hello World example of the Amazon States Language using two AWS Lambda Functions",
+        StartAt = "Hello",
+        States = new Dictionary<string, object?>{
+        ["Hello"] = new {
+            Type = "Task",
+            Resource = helloFunction.Arn,
```
Frassle, line 919: Maybe worth a comment pointing out this is an output?

themes/default/content/docs/intro/concepts/inputs-outputs.md
Frassle: Looks ok to me.

themes/default/content/docs/intro/concepts/inputs-outputs.md
dixler: I ran this example on my machine and it worked when the "unknown" key was commented out.

themes/default/content/docs/intro/concepts/inputs-outputs.md
dixler: I think I can deduplicate it for the examples for `Deserializing JSON Outputs`.

However, `Serializing Outputs to JSON` has a different example in `Go` with a different description and an additional Note. I can deduplicate the Header, but maybe not the content.

@cnunciato Is there a solution to do the deduplication for this second case?

abhinav: Ah, I didn't realize the content varied for Go. Makes sense based on what you said while writing these docs.

dixler: Nvm. I have a solution that I think looks good.

themes/default/content/docs/intro/concepts/inputs-outputs.md
```diff
@@ -651,6 +651,339 @@ variables:
...
+    },
+}));
+```
+
+{{% /choosable %}}
+{{< /chooser >}}
+
+### Example Parsing JSON Outputs to Objects
+
+The following is an example of Pulumi code that operates on a JSON object and removes all of the policy statements by replacing it with an empty list.
```
abhinav, line 801: This example does a bit more than what the title suggests it should.
We just want a JSON-parsed `Output<object>` from an `Output<string>`.
Although I'm not certain how you would demonstrate "this is an object" without doing stuff with it, so I guess this is fine.

themes/default/content/docs/intro/concepts/inputs-outputs.md
```diff
@@ -651,6 +651,339 @@ variables:
...
+        {
+            "Sid": "VisualEditor1",
+            "Effect": "Allow",
+            "Action": "s3:*",
+            "Resource": "arn:aws:s3:::my-bucket"
+        }
+    ]
+}`);
+
+const policyWithNoStatements = pulumi.jsonParse(jsonIAMPolicy).apply(policy => {
```
abhinav, line 862: I think the main value being demonstrated here is that we have an `object` instead of a `string`.
Do you think adding type annotations here would help?

```typescript
const iamPolicy: Output<object> = pulumi.jsonParse(jsonIAMPolicy);
const policyWithNoStatements ...
```

(I'm not sure if "object" above is correct, FWIW.)

themes/default/content/docs/intro/concepts/inputs-outputs.md
```diff
@@ -651,6 +651,339 @@ variables:
...
+});
+```
+
+For more details [view the NodeJS documentation](/docs/reference/pkg/nodejs/pulumi/pulumi/).
+
+{{% /choosable %}}
+{{% choosable language python %}}
+
+```python
+json_IAM_policy = pulumi.Output.from_input('''
```
abhinav, line 875: This should be `json_iam_policy`. `snake_case` doesn't do capitalization unless it's `SCREAMING_SNAKE_CASE`.

themes/default/content/docs/intro/concepts/inputs-outputs.md
```diff
@@ -651,6 +651,339 @@ variables:
...
+    ]
+}`).(pulumi.StringInput)
+
+policyWithNoStatements := pulumi.JSONUnmarshal(jsonIAMPolicy).ApplyT(
+    func(v interface{}) (interface{}, error) {
+
+    // delete the policy statements
+    v.(map[string]interface{})["Statement"] = []pulumi.ArrayOutput{}
+    return v, nil
+})
```
abhinav, line 940: The function body needs another level of indentation.

```suggestion
        // delete the policy statements
        v.(map[string]interface{})["Statement"] = []pulumi.ArrayOutput{}
        return v, nil
    })
```

Also, shouldn't this be `[]interface{}`, or `[]ArrayOutput`?

abhinav (commented):
Prose and heading setup looks okay. Left some comments on code samples.

Suggested changes:

- Consider dropping the "Example" from the headers
- Instead of starting with "The following is an example of [..]", consider adding an intro before that. Something like "You can parse a JSON string inside an `Output` to get back a parsed object inside the `Output` that you can operate on." (Optionally, use `{{< pulumi-output >}}` instead of `Output` to get the [pulumi-output shortcode](https://github.com/pulumi/pulumi-hugo/blob/master/themes/default/layouts/shortcodes/pulumi-output.html) which changes to `Output`, `Output<T>`, etc. based on chosen language.

abhinav (approved):
LGTM

PR #2590, opened 2023-03-24
Remove references to swag

scottslowe:
This PR removes references to swag in the Pulumi Challenges. The only exception is the "One Quickstart to Rule Them All," which has swag embedded so much throughout the post that I left it in place (there was/is already a banner at the top indicating that no swag will be sent out).

Fixes https://github.com/pulumi/devrel-team/issues/483


cnunciato (approved):
Awesome, thank you!

PR #2616, opened 2023-03-26
Fix out of place parenthesis

yardenshoham:
There was no opening parenthesis to match


ericrudder (approved):
Thanks @yardenshoham!

PR #2621, opened 2023-03-28
fix converters

susanev:
fixes https://github.com/pulumi/pulumi-hugo/issues/2605
and makes this a bit better

opted for making the button full-width as the easiest fix for now without rewriting the css for this
open to feedback, but i think this is pretty slick

what this does
- stops hiding convert button on screens less than xl, lol
- makes the convert button full width on all screens
- stops hiding the tabs on smaller screens, that weird white space you see is for the scroll bar; eventually we can come back and make this better

we can absolutely do a lot better on this, was making the smallest change to fix the bug

before
Screenshot 2023-03-27 at 6 18 14 PM

after
Screenshot 2023-03-27 at 6 18 02 PM


susanev:
closing this, will come back with a few more improvements in a new pr

PR #2622, opened 2023-03-28
blog: Project-Scoped Stacks in Self-Managed Backends

justinvp:
Adds a new blog post intended to be published this Wednesday (3/29) on the same day the feature will be released.

Part of https://github.com/pulumi/pulumi/issues/12173

Related Issues/PRs:
- https://github.com/pulumi/pulumi/issues/2522
- https://github.com/pulumi/pulumi/pull/12437
- https://github.com/pulumi/pulumi/pull/12438

PR #2633, opened 2023-03-30
STYLE-GUIDE: Allow one-line-pargraph and sembr

abhinav:
Per discussion with @cnunciato,
change the style guide to allow for one-line pargraphs
as well as sembr paragraphs.
The only form explicitly forbidden is reflowing text.

Moved the rationale into a `<details>` section
with examples demonstrating value of each case.

### DIFF
--- a/STYLE-GUIDE.md
+++ b/STYLE-GUIDE.md
@@ -54,37 +54,57 @@ This bit of info is serious. If you missed it, bad things could happen.
 
 * Keep paragraphs short, rarely use 4 or more sentences in a single paragraph.
 
-* When writing paragraphs and long sentences,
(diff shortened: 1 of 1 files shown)


susanev:
where did yall discuss this? would be rad to use public slack channels when possible

susanev (approved):

abhinav:
> where did yall discuss this? would be rad to use public slack channels when possible

Apologies, the discussion was in a DM (which in hindsight, should have been moved to a public channel).
I don't want to paste @cnunciato's messages as-is without asking him, so I'll let him do that.
But yes, next time we'll move the discussion to a public channel first.

FWIW, here's the context that led to that chat: https://github.com/pulumi/pulumi-hugo/pull/2622#discussion_r1153216874

PR #2637, opened 2023-03-31
concepts/stacks references: explain OutputDetails

### LINKED ISSUE #2547: Update Stack Reference docs to mention GetOutputDetails
Related to https://github.com/pulumi/pulumi/issues/10839.

Update the existing documentation for Stack References to mention the new `GetOutputDetails` method in each language.

https://www.pulumi.com/docs/intro/concepts/stack/#stackreferences

https://github.com/pulumi/pulumi-hugo/blob/master/themes/default/content/docs/intro/concepts/stack.md

abhinav:
This adds a new section to the StackReference section,
explaining how to get values out of a stack reference.

It covers both getOutput and getOutputDetails,
with examples of when one is preferred over the other.

getOuput is primarily suggested when the result
is input to another resource (after some transforms)
and getOutputDetails is suggested when you need direct access.

Code samples for getOutput were adapted from existing cases.
For getOutputDetails, all code samples were manually verified
with a sample project locally.

Notes:

- This functionality does not currently work in Python
  because we do not provide the ability to use `await`
  in Python programs. Ref: pulumi/pulumi#12172
- Although this works in Java, that version of pulumi-java
  has not yet been released.
  We will not land this before releasing pulumi-java.
  Ref: pulumi/pulumi-java#1042

Resolves #2547

### DIFF
--- a/themes/default/content/docs/intro/concepts/stack.md
+++ b/themes/default/content/docs/intro/concepts/stack.md
@@ -700,6 +700,275 @@ as your current stack (i.e. when deploying the "staging" stack of the above prog
 from the infra project. Once you have that resource, you can fetch the `kubeConfig` output variable with the `getOutput`
 function. From that point onwards, Pulumi understands the inter-stack dependency for scenarios like cascading updates.
 
+### Reading outputs from stack references
+
+Stack references support two ways of reading outputs from the referenced stack:
+
+* `getOutput` returns an `Output` that provides gated access to the output value.
+  The output value can be accessed and transformed with methods like `Output.apply`.
+  This is useful when the output is used as an input to another resource.
+* `getOutputDetails` returns an `OutputDetails` object that provides direct access to the output value.
+  This is useful when you want to process the output directly in your code.
+
+As demonstration of **`getOutput`**,
+suppose that your referenced stack exports a `privateIp` output.
+You want to incorporate the IP address into the name of an S3 bucket object
+containing logs from that machine.
+
+{{< chooser language "javascript,typescript,python,go,csharp,java" >}}
+
+{{% choosable language javascript %}}
+
+```javascript
+const infra = new pulumi.StackReference(...);
+const ip = infra.getOutput("privateIp");
+const logKey = ip.apply(ip => `logs/${ip}.log`);
+const logFile = new aws.s3.BucketObject("log", {
+    // ...
+    key: logKey
+});
+```
+
+{{% /choosable %}}
+{{% choosable language typescript %}}
+
+```typescript
+const infra: StackReference = new pulumi.StackReference(...);
+const ip: Output<string> = infra.getOutput("privateIp");
+const logKey: Output<string> = ip.apply(ip => `logs/${ip}.log`);
+const logFile: aws.s3.BucketObject = new aws.s3.BucketObject("log", {
+    // ...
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/intro/concepts/stack.md
```diff
@@ -700,6 +700,275 @@ as your current stack (i.e. when deploying the "staging" stack of the above prog
...
+    const host = new aws.ec2.Instance(`bastion-${i}`, {
+        // ...
+        subnetId: subnet.id,
+    });
+    // ...
+}
+```
+
+Note that your Pulumi program must export a top-level `async` function
+to be able to use the `await` operator.
```
susanev, line 831: how important is this note? should it be using the shortcode?

themes/default/content/docs/intro/concepts/stack.md
```diff
@@ -700,6 +700,275 @@ as your current stack (i.e. when deploying the "staging" stack of the above prog
...
+
+See [Javascript Entrypoint](https://www.pulumi.com/docs/intro/languages/javascript/#entrypoint)
+for more information.
+
+{{% /choosable %}}
+{{% choosable language python %}}
+
+{{% notes "info" %}}
+This functionality is not currently supported in Python.
+Please vote on [pulumi/pulumi#12172](https://github.com/pulumi/pulumi/issues/12172)
```
susanev, line 876: perfect usage of please here!

susanev (approved):

themes/default/content/docs/intro/concepts/stack.md
abhinav: Now that I think about it, it's not _that_ important.
Yeah, that can drop the short code.

abhinav:
moving to do-not-merge and draft mode until pu-java is released

themes/default/content/docs/intro/concepts/stack.md
abhinav: 🤦  Derp, sorry, you were suggesting that this could use the shortcode.
I thought I'd used it and you were advising against it.

So, to expand on why it's not super important:
code will fail to compile loudly if they don't do this. It's mentioned in the Pulumi Getting Started docs for Typescript.
I replicated it here mostly because this API is one of the places where they'll actually need to use an `await` in the Pulumi program, so it's good to mention "FYI, this is what you'll need if it doesn't work."

themes/default/content/docs/intro/concepts/stack.md
ericrudder, line 876:
```suggestion
Progress is tracked on [pulumi/pulumi#12172](https://github.com/pulumi/pulumi/issues/12172)
```

ericrudder: given that this isn't lack of desire nor trivial, i think tracking is more appropriate than voting ...

abhinav:
Current dependencies on/for this PR:
* master
  * **PR #2637** <a href="https://app.graphite.dev/github/pr/pulumi/pulumi-hugo/2637" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/pulumi/pulumi-hugo/2637?utm_source=stack-comment).

susanev (approved):

PR #2639, opened 2023-04-01
fix broken links

susanev:


kpitzen (approved):

ericrudder:
i think the first entry was incorrectly modified ... it should read
something like:

Log queries are used to retrieve any log data from Azure Monitor. Whether
you’re analyzing data in the portal
<https://docs.microsoft.com/en-us/azure/azure-monitor/log-query/portals>
or configuring
an alert rule
<https://docs.microsoft.com/en-us/azure/azure-monitor/platform/alerts-metric>
to
be notified of a particular condition, you will use a query to specify the
data you want.

Note the word *and  *formatting changes.

---

for the ksonnet post, i believe the link should just be removed and have
the plain text "announced" ... the link currently shows is abt corp
structure, not OSS projects.

Issue #2718, opened 2023-04-18
Add public access provisions for Getting Started flows


cnunciato:
Recent AWS changes to default bucket permissions have affected our Getting Started flows, which rely on `acl: "public-read"` to make website buckets accessible publicly. On the Deploy the Changes step, users now see the following error: 

```
Diagnostics:
  aws:s3:BucketObject (index.html):
    error: 1 error occurred:
        * updating urn:pulumi:dev::quickstart::aws:s3/bucketObject:BucketObject::index.html: 1 error occurred:
        * Error uploading object to S3 bucket (my-bucket-4bb6c43): AccessControlListNotSupported: The bucket does not allow ACLs
        status code: 400
```  

We should add the appropriate changes to get these flows fixed.

PR #2719, opened 2023-04-19
Remove website steps from the AWS getting-started flow

cnunciato:
## Description

An AWS change rolled out today that affects the way bucket permissions work, breaking our AWS getting-started flow. 

[As of today](https://aws.amazon.com/blogs/aws/heads-up-amazon-s3-security-changes-are-coming-in-april-of-2023/), new S3 buckets can no longer accept an `acl` value of `public-read`, which our getting-started flow uses to demo creating a simple S3 website. As a result, the tutorial fails on the "modify program" step (which needs a public bucket and bucket-object for the website to work), so this change removes the website-related parts of that step and demos creating a simple text file instead.

Fixes https://github.com/pulumi/pulumi-hugo/issues/2718.

Note that this change does not actually fix the bug as written, as doing so would require several more lines of code and add unnecessary complexity to the tutorial.

## Checklist:


themes/default/content/docs/get-started/aws/create-project.md
```diff
@@ -112,19 +112,27 @@ Saved config
 > What are [projects](/docs/intro/concepts/project/) and [stacks](/docs/intro/concepts/stack/)? Pulumi projects and stacks let you organize Pulumi code. Consider a Pulumi _project_ to be analogous to a GitHub repo---a single place for code---and a _stack_ to be an instance of that code with a separate configuration. For instance, _Project Foo_ may have multiple stacks for different development environments (Dev, Test, or Prod), or perhaps for different cloud configurations (geographic region for example). See [Organizing Projects and Stacks](/docs/guides/organizing-projects-stacks/) for some best practices on organizing your Pulumi projects and stacks.
 
 {{% choosable language "javascript,typescript" %}}
+
 After some dependency installations from `npm`, the project and stack will be ready.
```
cnunciato, line 116: Added this whitespace because the Markdown `npm` isn't rendering properly without it:

image

lukehoban:
This is a pretty significant regression in the quality of this guide. Do we have a plan to recover all the content in modifying the program?  That content is not just for the website use case, it also helps uses understand updates, and other Pulumi concepts.

susanev:
@cnunciato @sean1588 lets sync up about this tomorrow, will grab some time

cnunciato:
@lukehoban Yeah, I know -- this is purely to fix something that's currently broken. The alternative would be to add the code to support the new restrictions, which is a good bit more complex (uses `dependsOn`, etc.); I wanted to get this up first, to head off anyone having a bad/broken experience now, then circle back and address whether and how to bring the experience back up to where we'd like it.

susanev:
@cnunciato do i understand correctly that the negatives of folks not getting a chance to understand pulumi concepts are outweighed by aws errors they will hit at the end of the tutorial?

what is your timeline on returning to this?

lukehoban (approved):
Let's merge this to fix the current failures, but come back and bring back in a working version with the original structure within the next couple days.

cnunciato:
@lukehoban Sounds great, thanks -- definitely on board with that.

cnunciato:
Filed and will follow up on this tomorrow: https://github.com/pulumi/pulumi-hugo/issues/2720

PR #2728, opened 2023-04-20
add more deets to resource version docs

susanev:
## Description
add more deets to resource version docs

## Checklist:

### DIFF
--- a/themes/default/content/docs/intro/concepts/resources/options/version.md
+++ b/themes/default/content/docs/intro/concepts/resources/options/version.md
@@ -9,7 +9,7 @@ menu:
     weight: 15
 ---
 
-The `version` resource option specifies a provider version to use when operating on a resource. This version overrides the version information inferred from the current package. This option was built to be used directly by the Pulumi SDK. This option should rarely be used by people directly.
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/intro/concepts/resources/options/version.md
```diff
@@ -9,7 +9,7 @@ menu:
     weight: 15
 ---
 
-The `version` resource option specifies a provider version to use when operating on a resource. This version overrides the version information inferred from the current package. This option should be used rarely.
+The `version` resource option specifies a provider version to use when operating on a resource. This version overrides the version information inferred from the current package. This option was built to be used directly by the Pulumi SDK. This option should rarely be used by people directly.
```
iwahbe, line 12: Optional Nit: "rarely" is confusing. We don't mean "don't do it often". We mean "You should know what your doing when you use this, and want to override default resource provider mapping in an untested way."
```suggestion
The `version` resource option specifies a provider version to use when operating on a resource. This version overrides the version information inferred from the current package. This option was built to be used directly by the Pulumi SDK. `version` should not be used directly during normal operations.
```

iwahbe (approved):

sean1588 (approved):

Issue #1707, opened 2022-07-06
Subscribe to newsletter button on blog pages


meagancojocar:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

The subscribe to newsletter call to action (CTA) has no placeholder text or label on what input the user is supposed to enter. I would recommend whoever enhances this looks at any front end framework as reference for how users expect a form to behave (i.e. [Materialize](https://materializecss.com/text-inputs.html))

### Affected area/feature

Bottom of the blog page is where I noticed this, [see here](https://www.pulumi.com/blog/amazon-eks-anywhere-bare-metal/).
![Screen Shot 2022-07-06 at 9 25 29 AM](https://user-images.githubusercontent.com/50200557/177598418-756ceb3e-8c18-4a7f-a359-64fa648204e2.png)

PR #2765, opened 2023-04-28
make the newsletter input look more like an input

susanev:
## Description
the newsletter input looks like a button without a label making a small change to make this look a bit more like an input
i don't think its possible to add a label cause this is hubspot
but this should make it a bit better
fixes https://github.com/pulumi/pulumi-hugo/issues/1707

### before
Screenshot 2023-04-28 at 3 40 38 PM

### after
Screenshot 2023-04-28 at 3 40 29 PM

## Checklist:

### DIFF
--- a/themes/default/theme/src/scss/main.scss
+++ b/themes/default/theme/src/scss/main.scss
@@ -351,4 +351,8 @@ pulumi-tree-item {
     }
 }
 
+section.newsletter-input pulumi-hubspot-form input.hs-input {
+    border-radius: 4px;
+}
+
 @tailwind utilities;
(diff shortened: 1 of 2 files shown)


themes/default/theme/src/scss/main.scss
```diff
@@ -351,4 +351,8 @@ pulumi-tree-item {
     }
 }
 
+section.newsletter-input pulumi-hubspot-form input.hs-input {
+    border-radius: 4px;
```
cnunciato, line 355: Forgive me 😇 :

```suggestion
    @apply rounded;
```

cnunciato (approved):

themes/default/theme/src/scss/main.scss
susanev: haha, its all good

PR #2782, opened 2023-05-04
Pulumi AI shareable links 

zchase:
## Description

This PR adds support for sharing conversations via a share button next to the copy button. This allows users to share generated results with others, as well as have others build other their original conversation.

sharing-menu-pulumiai

## Checklist:


themes/default/theme/stencil/src/components/ai/ai.tsx
```diff
@@ -264,10 +292,47 @@ export class PulumiAI {
...
+        this.versions = Array.from(versions);
+        this.currentVersion = null;
+
+        this.running = false;
+        this.scroll();
+        this.clear();
+        this.focus();
+
+        // Wait a bit for the versions to render, then colorize them.
+        setTimeout(() => this.highlight(this.outputVersions), 250);
```
cnunciato, line 335: Might be nice to eliminate some of this duplication, but we could also come back to it.

themes/default/theme/stencil/src/components/ai/ai.tsx
```diff
@@ -316,6 +382,14 @@ export class PulumiAI {
                         <span slot="content">${ this.copyButtonTooltip }</span>
                     </pulumi-tooltip>
                 </li>
+                ${ shareable ? `<li class="ml-2">
+                    <pulumi-tooltip>
```
cnunciato, line 386: So the way this shakes out is that the UI makes it look like what you're sharing is a code block, but what you're really sharing is effectively the whole page/conversation. In that sense, it seems like the share button might belong more at the "top level" somehow  -- although admittedly there doesn't seem to be a great place in the current layout that quite makes room for that. (At first I thought it might go better at the prompt level -- but that doesn't quite work either, as conversations can have multiple prompts.)

I guess it's worth asking what one would expect to be sharing here. If you'd expect to share the whole conversation, then we might want to come up with more of a top-level position for the button, if only to make it clearer to the user they'll be sharing the whole page. (For this, we might even consider updating the URL dynamically with the conversation ID and just letting users copy and share that.) For my part, I sort of imagined this working by scoping the share to the "prompt + result combo" level, in part because conversations can be iterative (and I'd probably be most interested in sharing the end result) ... but that's just me; others might want to use this differently. What do you think?

cnunciato (approved):
This all looks good and nothing need block, though it'd be good to consider the connection and button placement questions, in case there are improvements we might want to make there before merging.

scottslowe (approved):
Looks good to me, although I defer to others for the mechanics of how it all works and I don't really have a way of checking the functionality. :-)

cnunciato:
>and I don't really have a way of checking the functionality

@scottslowe Ah, but you do! If you follow the latest preview URL and browse to Pulumi AI (either at `/ai` or by clicking the link in the upper right corner), you can see the changes Zack's added in this PR and see how they feel. Copying a share link and pasting, say, into an in-cognito window should simulate what the experience will be like for users.

lukehoban:
Just tried this out - a couple notes:
1. The options for sharing were available even before the output finished producing, but didn't include a link to the in-progress response.  This wasn't intuitive - it feels like we should either (a) not allow sharing until the response is finished or (b) allow sharing the link to the result even before the full result is done rendering.
2. We seem to be using cmd+enter as the newline, and shift-enter submits the output.  This is the opposite of what ChatGPT does - and what I independently expected.
3. Not technically a change here - but it might be nice to not start a prompt if it's an empty string (or only whitespace), so that hitting <enter> or the wrong <special>+<enter> doesn't submit anything, and just leaves the input unmodified.  Again similar to ChatGPT.
4. Quite likely not for this PR, but it does feel like it would be amazing to do a short summarization of the session, and use that in the Twitter post template - so that each tweet is auto-personalized

themes/default/theme/stencil/src/components/ai/ai.tsx
```diff
@@ -228,11 +254,38 @@ export class PulumiAI {
...
+
+        if (this.submissionCount === 0) {
+            return `${this.siteUrl}/ai/?convid=${this.existingConversationId}`;
+        }
+
+        return `${this.siteUrl}/ai/?convid=${this.conversationId}`;
+    }
+
+    private get twitterLink() {
+        let postText = "Checkout what I've generated with Pulumi AI!";
```
cnunciato, line 270:
```suggestion
        let postText = "Check out what I generated with Pulumi AI!";
```

themes/default/theme/stencil/src/components/ai/ai.tsx
```diff
@@ -228,11 +254,38 @@ export class PulumiAI {
...
+        if (this.versions.length === 0) {
+            postText = "Pulumi AI can help you generate Infrastructure as Code specific to your use case. Give it a try!"
```
cnunciato, line 272:
```suggestion
            postText = "Pulumi AI can help you generate infrastructure as code specific to your use case. Give it a try!"
```

themes/default/theme/stencil/src/components/ai/ai.tsx
```diff
@@ -675,38 +782,60 @@ export class PulumiAI {
...
+                                            <span><i class="fas fa-link"></i>Copy Link</span>
+                                        </li>
+
+                                        <li>
+                                            <a href={this.twitterLink} target="_blank" rel="noopener noreferrer"><i class="fab fa-twitter"></i>Share on Twitter</a>
+                                        </li>
+                                    </ul> : undefined}
+
+                                    <div class="share-link" onClick={() => this.toggleShareMenu()}>
+                                        <span><i class="fas fa-share-square"></i> Share Conversation</span>
```
cnunciato, line 836: Can we also make this a smidge smaller to match the font size of what's to the left, kind of like this?

image


```suggestion
                                        <span><i class="fas fa-share-square"></i> Share conversation</span>
```

cnunciato (approved):
+1 to Luke's comments, but otherwise just a couple of little things for you to consider.

PR #2800, opened 2023-05-07
Update my (Friel's) photo

AaronFriel:
A slightly higher resolution is used at a slightly lower quality level to keep the size down.


susanev (approved):
fancy fancy 🧔🏻

PR #2807, opened 2023-05-09
adding swifttype to parent pages in docs

susanev:
## Description
today i noticed we have a list layout and a single layout for docs. and the list layout is for all parent pages.
the parent pages have content on them, but i think they were previously not being indexed in swifttype
this adds it there

i might be totally wrong, pls tell me if i am

here's the single layout
https://github.com/pulumi/pulumi-hugo/blob/master/themes/default/layouts/docs/single.html#L23-L25

## Checklist:

### DIFF
--- a/themes/default/layouts/docs/list.html
+++ b/themes/default/layouts/docs/list.html
@@ -23,7 +23,10 @@
                         <h1>{{ .Title }}</h1>
                     {{ end }}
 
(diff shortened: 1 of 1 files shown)


cnunciato (approved):
Seems fine, can't think of a reason not to do this. Thanks for spotting it!

PR #2817, opened 2023-05-11
Add feedback mechanism for pulumi ai

zchase:
## Description

This PR adds a feedback block at the end of response for Pulumi AI. The response are then stored in Metabase.

pulumi-ai-feeback-ui

## Checklist:


themes/default/theme/stencil/src/components/ai/ai.tsx
```diff
@@ -334,10 +346,13 @@ export class PulumiAI {
         this.highlight(this.outputStream);
     }
 
-    private onComplete() {
+    private onComplete(response: GenerateNewOutputResponse) {
+        console.log("output complete", response);
+
```
cnunciato, line 351:
```suggestion
```

cnunciato (commented):
LGTM for the most part, but a couple of questions:
* It looks like we only send the feedback if they click the submit button. Many won't choose to do that though -- and the UI doesn't make it clear you can/should click the Submit button before your up/down vote will be sent. Should we maybe capture the up/down click regardless of whether they click the submit button, and then send a comment if they choose to submit one? (I think this is how we do it on the website, but I might be wrong on that.) We should probably provide a cancel/no button as well to it's clear how to say no to that prompt. What do you think?
* How does this interact with the convo-sharing PR? I'm assuming this'll only show up for conversations initiated by the user?

zchase:
@cnunciato There is some logic to send all the outstanding feedback requests (ones that are "unsubmitted") when the component disconnects. I agree that the UI could use some fine-tuning.

I'm not exactly sure how it will interact with the sharing stuff to be honest. I think it would be best to get the sharing PR merged then rebase this on top of it to see where we are at.

shughes26 (approved):

PR #2856, opened 2023-05-17
fix links

susanev:
## Checklist:


themes/default/content/docs/concepts/glossary.md
```diff
@@ -6,6 +6,8 @@ h1: Pulumi glossay
 menu:
   concepts:
     weight: 12
+aliases:
+- h/docs/reference/glossary/
```
susanev, line 10:
```suggestion
- /docs/reference/glossary/
```

PR #2868, opened 2023-05-17
Throwaway: Search POC

cnunciato:


susanev:
sry for the conflicts @cnunciato, let me know if you want any help rebasing 💜

anita-trimbur:
- [ ] did we lose the "docs" and "packages" chips in the all results section at some point? i could've sworn they were in an older preview
- [ ] we should match the "packages" label in the docs type nav instead of "registry"
- [ ] it looks like the icon for "all results" is bigger than the others - is that because it's a fontawesome icon? i can get you an svg of the material one so everything matches
- [ ] do we have control of that underlining style for matching your query? it's very faint right now, am wondering if we can make that a little more prominent
- [ ] i found this weird bug when trying to resize a small window with search open
https://github.com/pulumi/pulumi-hugo/assets/97129308/8d0af463-d39b-43d8-88e5-450dbaa95efe
- [ ] we have some overlapping with the tabs right before the mobile breakpoint, we'll need to shrink the search field for those smaller screens
Screen Shot 2023-06-08 at 4 23 19 PM

- [ ] would we also be able to shrink the height of the search field? i think it's adding about 4px extra of height to that type nav
- [ ] i imagine this has to do with the changes you had to make for ranking that you were telling me about, but this for example was one i encountered that confused me for a second
Screen Shot 2023-06-08 at 4 15 12 PM

- [ ] some odd layering on packages browse page
Screen Shot 2023-06-08 at 4 25 51 PM

- [ ] i had robbie run through a couple of searches with me and this is the one that came back most surprising - he would be looking for the latest api definition, which appears fourth in this list
Screen Shot 2023-06-08 at 4 30 26 PM

themes/default/layouts/partials/blog/search.html
```diff
@@ -1,58 +1,9 @@
...
-        </form>
-    </div>
-    <div id="search-results" class="relative"></div>
+<div>
+    <div
+        id="autocomplete"
+        data-app-id="{{ getenv "ALGOLIA_APP_ID" }}"
+        data-search-key="{{ getenv "ALGOLIA_APP_SEARCH_KEY" }}"
+        data-facets="Docs,Registry"
+        data-index="preview"
```
cnunciato, line 7: Note to self: Change this to production before merging.

themes/default/layouts/partials/docs/search.html
```diff
@@ -21,14 +21,13 @@
...
-            </form>
-            <div id="search-results" class="relative"></div>
-        </div>
+    <div class="docs-search">
+        <div
+            id="autocomplete"
+            data-app-id="{{ getenv "ALGOLIA_APP_ID" }}"
+            data-search-key="{{ getenv "ALGOLIA_APP_SEARCH_KEY" }}"
+            data-facets="Docs,Registry"
+            data-index="preview"
```
cnunciato, line 30: Note to self: Change this to production before merging.

cnunciato:
Ok! I still have some comments I'd like to add, but this should be good to go. 🚀 

@anita-trimbur let me know what you think. Two issues I'm aware of at this point:

* When you arrow down/up to the last or first visible item, it's still slightly obscured, which is a little annoying, and is happening because the header (tabs) and footer (navigation instructions) are fixed in their positioning, so are sitting above the items (which would otherwise be fully visible). I wouldn't consider this a ship blocker personally, but wanted to let you know it's something I do plan to address as soon as I can.

* The Docs/Packages chips. They're there, but there's something about their color (which stands out quite a bit compared to everything else) that I find fairly distracting, as it pulls my attention away from the item metadata, which is far more important in terms of quickly deciding whether that item is the thing I'm looking for. I wonder if it might be better to use something less visually prominent -- like an icon. What do you think? Algolia does something similar, for example:

    image

I _think_ that's it -- lemme know what you think!

themes/default/theme/src/ts/algolia.ts
```diff
@@ -0,0 +1,335 @@
...
+                        <img src="/images/search/no-results.svg"/>
+                        <p>We couldn't find any results for <mark>${ state.query }</mark>.</p>
+                    `;
+                },
+            },
+        };
+    }
+
+    function maybeScrollTo(itemID) {
+        console.log({ wasKeyEvent });
```
cnunciato, line 291: Will remove.

themes/default/theme/src/ts/algolia.ts
```diff
@@ -0,0 +1,335 @@
+import algoliasearch from "algoliasearch/lite";
+import { autocomplete, getAlgoliaResults, getAlgoliaFacets } from "@algolia/autocomplete-js";
+import { createTagsPlugin } from "@algolia/autocomplete-plugin-tags";
+
+const autocompleteContainer = "#search";
```
cnunciato, line 5: I'll add plenty of comments to this file and the CSS file on Monday.

cnunciato:
>this should be good to go

Actually, strike that -- something weird is going on; the Docs tab is suddenly not working at all! I'll take a look. 😄

susanev:
@cnunciato great work on this so far!

im surprised to see that we are still using titles for the returned bold values. will you be able to switch this to h1s? i put a lot of effort into overhauling those h1s to inform this search, i hope we will be able to use them!

susanev:
@anita-trimbur what would you think of putting the badges either directly to the left or right of the breadcrumbs? now that we are in a dialog they seem quite disconnected from the content, and possibly easy to miss, thoughts?

themes/default/theme/src/scss/_search.scss
```diff
@@ -1,184 +1,55 @@
 .docs-type-nav-search {
-    display: flex;
-    flex-direction: column-reverse;
-    justify-content: space-between;
-    background-color: $white;
+    @apply flex flex-col-reverse justify-between bg-white;
```
susanev, line 2: can you leave the css as it was instead of changing to tailwind unless we have to?

anita-trimbur:
ayyy this is looking sick!! rad that you figured out how to do the dialog and everything! 

to address your callout about the chip color being too strong, we can drop the background on those and just use the colored borders. i think that should feel less distracting for folks. 

in addition to susan's comments, the other thing i'm wondering is if we can use a fixed height on the dialog so the changing heights isn't so distracting. i like the max-height you have on that container now. that would make the empty state before the user types in a query very empty, but to susan's point, i think we can include a small message as a low lift alternative to the suggested/recent search which we can come back to some other time.

themes/default/theme/src/ts/algolia.ts
```diff
@@ -0,0 +1,335 @@
...
+            }
+
+            wasKeyEvent = false;
+        }
+    }
+
+    const ac = autocomplete({
+        container,
+        openOnFocus: false,
+        placeholder: "Search (⌘/ctrl-c)",
```
sean1588, line 308: should this be ctrl-k?

sean1588 (commented):
This seems to work well. I think we do need to do a bit of fine-tuning when it come to the search results. Just poking around at some commonly used resources seems to come up with results I wouldn't expect sometimes.

for example, i would expect that a search for s3 bucket would return me the aws s3 bucket resource first, but it is quite far down in the list.

Screenshot from 2023-06-12 09-09-03


iam role for akamai seems to far out rank aws 
Screenshot from 2023-06-12 09-10-34

i wonder if there is any way we can leverage existing resource consumption to help inform the ranking of these results when it comes to ranking the packages, (i.e. rank the most commonly used resources first). I have this data in a structured json format from the related resources work I did if you want to try and leverage that at some point.

cnunciato:
@sean1588 Yeah, I'm very sure we're going to unearth lots of issues like with certain results ranking as we'd like, particularly it's put to more use. This PR has no ability to adjust that, though, as that all happens in the indexing. So if you could review this PR for the UI specifically, that'd be awesome. Thank you!

cnunciato:
@sean1588 Come to think of it, I've already created an issue for this! 

Would you mind adding what you noted above to that issue as a comment? Thanks!

https://github.com/pulumi/devrel-team/issues/570

cnunciato:
@susanev @anita-trimbur Thank you both! These are good lists to work by.

Unfortunately, we don't have direct control over most of the items in these lists because we're using Algolia's autocomplete UI, which only gives us so much in the way of customization. As rule, if it's a CSS-only tweak, three's a pretty good chance we can do at least _something_ about it -- but if it's JS-based, we may not be able to. (For example, the behavior with the button also showing the query is baked in; about all we could do there is hide the text, I think.) I'd of course happy to look into each one of these to make sure -- though it'd take a bit of time to do that of course. (And FWIW, I have time set aside this week for exactly this sort of thing.)

Of the issues cited here, I'd like to break them into two lists -- blocking and non-blocking. Here's my attempt at that (in no particular order), but feel free to adjust if you see anything else that should truly be blocking but isn't:

### Blockers

- [ ] Fix the ctrl-c typo
- [ ] Make the search button (the static thing in the nav) the right pixel width
- [ ] Remove the chip background color

### Non-blockers

- [ ] Handle `/` also for search (I actually took this out because it was interfering with my ability to search for things with legit slashes in them, like "@pulumi/awsx", which we definitely do need. Should be doable, but will require some code)
- [ ] Look into using H1s also (I'm taking a look at this now, but FWIW, the changes would need to be made at indexing time (so in pulumi/docs), so we needn't block this PR for this one --- unless of course the results we show now are results we really don't want users to see)
- [ ] Jumpiness when the dialog opens and closes (I'm not sure I see this, so if you could share a video, that would help!)  
- [ ] See if we can keep the x icon from closing the dialog
- [ ] Consider repositioning the chips/badges (inline didn't quite work because of wrapping issues -- I'll post here with an example so you can see what that looks like)
- [ ] Consider alternative treatment for the input field when the dialog is open (distracting, not super valuable until you close) 
- [ ] Consider alternate empty-state mobile experience
- [ ] Add a CTA of some sort to the empty-state experience (e.g., "Try something else?", a clear button, etc.)
- [ ] Consider a softer animation when the dialog opens and closes
- [ ] Look into using a fixed hight in the dialog
- [ ] Comments! I'd still like to add lots of comments to the code

Let me know what y'all think and if I've missed anything. Thanks again for all the help in making this happen!

themes/default/theme/src/scss/_search.scss
cnunciato: I'd prefer we kept to the standard Tailwind here, as it eliminates a bunch of code that Tailwind covers -- but I don't at all mind you writing CSS if that's what makes you happy!

susanev (approved):

anita-trimbur (approved):

themes/default/layouts/partials/blog/search.html
```diff
@@ -1,58 +1,3 @@
...
-                id="search-query"
-                type="text"
-                class="st-default-search-input py-2 px-4 pl-8 text-sm border border-gray-400 focus:border-blue-500 focus:outline-none rounded transition"
-                placeholder="Search the blog"
-            />
-        </form>
-    </div>
-    <div id="search-results" class="relative"></div>
+<div>
+    <div id="search" data-app-id="{{ getenv "ALGOLIA_APP_ID" }}" data-search-key="{{ getenv "ALGOLIA_APP_SEARCH_KEY" }}" data-facets="Docs,Registry" data-index="production"></div>
```
cnunciato, line 2: This should be `Blog`, not `Docs,Registry`:

```suggestion
    <div id="search" data-app-id="{{ getenv "ALGOLIA_APP_ID" }}" data-search-key="{{ getenv "ALGOLIA_APP_SEARCH_KEY" }}" data-facets="Blog" data-index="production"></div>
```

themes/default/theme/src/scss/_search.scss
susanev: @cnunciato we talked in the last frontend working group that docs is moving away from tailwind, its all mostly been removed. id prefer if you didn't slowly add it back.

PR #2875, opened 2023-05-18
fix blog search results

susanev:
## Description
i think we overlooked blog search when we updated the docs search
this should fix it up quickly
but we may want to revisit this

it seems for blog you have to type 2 letters to get results

### before
Screenshot 2023-05-18 at 6 43 48 AM

### after
Screenshot 2023-05-18 at 6 43 25 AM

## Checklist:


sean1588 (approved):
ahhhhh i totally forgot there was a blog search. thank you so much for fixing this susan.

PR #2895, opened 2023-05-22
update typography

susanev:
## Description
@anita-trimbur this is ready for your first pass, let me know what you want adjusted
also i need top and bottom margins for all headings and paragraphs, right now its using what was already set

fixes https://github.com/pulumi/docs/issues/8972

to do

- [x] get gilroy medium on website
- [x] adjust headings to gilroy medium
- [x] update link styles
- [x] add animation to link styles
- [x] fix special case links
- [x] add h1 styles
- [ ] code styles in text?
- [ ] update accordion styles
        > i actually like the size of the accordions right now, it's really just the underline on hover that's kinda lame. if we adjust the heading and icon color to gray700 inactive and then gray900 on hover (no underline), i think that'd be an improvement.
- [x] update table styles
- [x] cloud pages

## Checklist:


susanev:
what styles do we want for accordions? do you want to design a new accordion?
docs-accordions

susanev:
can we include in-paragraph link styles in this work? what are those styles?

susanev:
did the cloud headings have special heading sizes for cards?

susanev:
any changes to table headings?
Screen Shot 2023-05-21 at 6 24 18 PM

susanev:
@anita-trimbur 
>i know this is a case-by-case thing, but i was surprised by several instances where there isn't a normal heading progression. for instance, i expected that docs home would use h2s, but it only has h3s. is this something we'll have to revisit, or something that we might be able to adjust in more prominent places now?

we have to fix this on the pages themselves, if you keep a list of which ones are wrong i can easily update them

>wanted to confirm whether the h1 gradient hasn't been included yet or if you were wanting to omit

only haven't done it yet, and wanted to get this up for you to look around

anita-trimbur:
> what styles do we want for accordions? do you want to design a new accordion?

i actually like the size of the accordions right now, it's really just the underline on hover that's kinda lame. if we adjust the heading and icon color to gray700 inactive and then gray900 on hover (no underline), i think that'd be an improvement.

susanev:
another question for you @anita-trimbur if one heading is directly followed by another heading do you want any changes to the margins? the project file under concepts > projects is a good example of this.

PR #2896, opened 2023-05-22
Update "Understanding Stack Outputs" Python code

2tunnels:
## Description

I think this is a typo, there is no `format` function in `pulumi` module.

## Checklist:

### DIFF
--- a/themes/default/content/learn/building-with-pulumi/stack-outputs/index.md
+++ b/themes/default/content/learn/building-with-pulumi/stack-outputs/index.md
@@ -42,7 +42,7 @@ export const url = pulumi.interpolate`http://localhost:${frontendPort}`;
 {{% choosable language python %}}
 
 ```python
-pulumi.export("url", pulumi.format("http://localhost:{0}", frontend_port))
+pulumi.export("url", f"http://localhost:{frontend_port}")
 ```
 
 {{% /choosable %}}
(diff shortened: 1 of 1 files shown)


themes/default/content/learn/building-with-pulumi/stack-outputs/index.md
```diff
@@ -42,7 +42,7 @@ export const url = pulumi.interpolate`http://localhost:${frontendPort}`;
 {{% choosable language python %}}
 
 ```python
-pulumi.export("url", pulumi.format("http://localhost:{0}", frontend_port))
+pulumi.export("url", f"http://localhost:{frontend_port}")
```
Frassle, line 45:
```suggestion
pulumi.export("url", Output.format("http://localhost:{0}", frontend_port))
```

frontend_port would generally be an Output value and so can't be used with f-strings. There is a `format` method for this but it's on the `Output` class not the root pulumi module.

2tunnels: Hm, I thought `frontend_port` is a simple integer in this case:

```python
frontend_port = config.require_int("frontend_port")
```

So, should I update the PR to include full method path, like `pulumi.Output.format`?
Because `Output` class is not imported in that context.

Frassle: Yeh in this case it would be fine to f-string it, but keeping the style of the the Typescript example on this page I think these examples are intended to be more general. 

And yes if Output isn't imported directly anywhere than `pulumi.Output.format` would be more correct.

Frassle (approved):

PR #2920, opened 2023-05-24
auto scroll to active element

sean1588:
Auto scroll to active element in docs left nav. Not sure if there is a better way to do this, but gets the job done. Note, this will not scroll fully to the top for the elements nearer to the bottom of the list, because there is just no more to scroll since the scroll bar is already at the bottom and fully scrolled.

Peek 2023-05-24 13-52


susanev (approved):
thank you Sean!!!

PR #2927, opened 2023-05-25
pulumi up banner

susanev:
## Description
display a register for pulumi up banner only on the home page

todo
- [x] close x dimisses and saves in local storage
- [x] banner is sticky on scroll and sits above the white nav banner
- [x] banner should go to pulumi up page, right now it goes to localhost?
- [x] move banner to bottom
- [x] center content
- [x] underline register on hover
- [x] remove other banner
- [x] always underline register now

Screenshot 2023-05-25 at 1 25 53 PM

## Checklist:


kimberleyamackenzie (approved):
Code generally looks good to me, 2 things I found:
- [ ] I have to select the X pretty on the mark to not be navigated to the pulumi up page, I think the link target should be narrower to prevent this
- [ ] On mobile, the heart gets cut off by the X
Screenshot 2023-05-25 at 1 51 49 PM

On more of an opinion note, my attention really doesn't go to the banner when its on the bottom - I think my brain is trained to ignore stuff down there since its usually where all the cookie notifications live, so if it's not blocking content, I don't look.

susanev:
tots agree on the banner position, but folks asked specifically for this location so 🤷🏻

SaraDPH (approved):

PR #2935, opened 2023-05-26
remove future mention

susanev:
## Description
sometimes the future never comes
we should document what is, not what might happen in the future as it often disappoints users or confuses them

## Checklist:


guineveresaenger (approved):
thank you!

komalali (approved):

PR #2944, opened 2023-05-27
layoutz

susanev:
## Description
this is only in docs right now not registry

also, @anita-trimbur how would you feel about saving the hamburger menu for another pr? and leaving the little left-nav tab on mobile until then?

- [x] fix toc location on smaller screens
- [x] fix content responsiveness
- [x] i seem to. have broken responsiveness on code blocks
- [x] comments from Anita in figma
- [x] left nav scrolling does not yet account for search movement on mobile screens
- [x] left nav scrolling on smaller screens its not possible to get to the bottom entries because of magic numbers, i need to fix this
- [x] i broke the tweets on the home page
- [x] still having left nav scrolling issues
- [x] left nav scrolling on larger screens needs more space at bottom of nav
- [x] hide toc when open left nav, if table of contents is on screen
- [x] extra space added below h1 for pages without tocs
- [x] empty tocs are still showing up
- [x] add missing breakpoints
- [x] i need variables for my magic numbers
- [x] clean up css
- [x] clean up ts
- [x] confirm no impact to marketing pages

id like to save these items for other prs..

- registry work
- hamburger menu on mobile
- save left-nav open state in local storage? maybe?
- the left-nav html really really needs to be re-done

this makes it seem like layout issues but its really content issues, well and @anita-trimbur we need designs for a responsive table in docs

fixes some of https://github.com/pulumi/docs/issues/8988

what this pr does...

- fixes docs type nav +search to pin to sides
- hides left nav, and introduces left nav toggle on smaller screens
- moves table of contents below h1 on smaller screens
- redesigns docs page layouts
- adds a file for icons
- updates colors css file
- removes unused docs templates
- updates names of things to be more descriptive
- updates some things to use css instead of tailwind
- build out accordion functionality of table of contents on smaller screens
- collapses some css into the main file
- reduces left-nav font-size to 15px

## Checklist:


anita-trimbur:
v v excited about this pr! i feel like this really makes all the docs changes come together ✨

some minor style things i noticed:

* some spacing discrepancies between the main doc content and the toc - i think the gap needs reduced from 36px to 24px, and the left-right padding on the toc from 18px to 16px. it also looks like the doc content pay have extra padding on the sides that the designs don't account for
* the left nav itself takes up a little more width, which i think is stealing some of the width allocated for the toc at some screen sizes. left nav width can decrease to 240px from 265px 
* resolving the above should allow the toc the space so as not to become any narrower than 280px
* not sure if this is expected, but i can no longer click on any of the toc links

susanev:
@anita-trimbur all your comments have been addressed, pls let me know if there's anything else you find that needs fixing!
https://github.com/pulumi/pulumi-hugo/pull/2944#issuecomment-1568751119

anita-trimbur:
few more thingz:

- it looks like between 768 - 1024 we still get a really narrow toc at some widths. i think the reason is that the content area under the 1024 breakpoint should become narrower, decreasing from 700 to about 634px in the designs. that should give the toc room at those screen widths.
- i know you just fixed this according to my figma comment, but i totally forgot to test that hover style when the nav is already open and that border color change looks wack. we can probably drop the border color change.
- do you think that the left nav (when it's overlaying the content at smaller screen widths) should close when clicking outside the nav container?
- lastly, is it possible for the toc to scroll independently of the main content? i ask because there's some instances of really long tocs that go below the page fold and it's not possible to scroll to see those links without scrolling all the way to the bottom of the page itself. example: http://pulumi-hugo-origin-pr-2944-4af4e2ec.s3-website.us-west-2.amazonaws.com/docs/concepts/state/

susanev:
>  lastly, is it possible for the toc to scroll independently of the main content? i ask because there's some instances of really long tocs that go below the page fold and it's not possible to scroll to see those links without scrolling all the way to the bottom of the page itself. example: http://pulumi-hugo-origin-pr-2944-4af4e2ec.s3-website.us-west-2.amazonaws.com/docs/concepts/state/

@anita-trimbur im assuming this is desktop only? is that correct? what is the max-height you want? i set it to 400px but can change it

susanev:
> it looks like between 768 - 1024 we still get a really narrow toc at some widths. i think the reason is that the content area under the 1024 breakpoint should become narrower, decreasing from 700 to about 634px in the designs. that should give the toc room at those screen widths.

@anita-trimbur this would be easier if you gave me percentages, what percentage of the available width do you want for the content area vs the right table of contents? once we get to a small enough size that is

anita-trimbur:
@susanev i may not be following in which case we should zoom tomorrow, but i don't think a percentage is what we want because that would be the element would be of variable sizes depending on the screen width. the pixel values in the designs are all fixed values that are intended to work without scaling within the range of each breakpoint.

themes/default/theme/src/ts/docs-main.ts
```diff
@@ -0,0 +1,83 @@
...
+        docsMainNavToggleWrapper.toggleClass("docs-nav-show");
+        docsMainNavToggleWrapper.toggleClass("docs-nav-hide");
+        docsNavToggleIcon.toggleClass("close-docs-main-nav");
+        docsNavToggleIcon.toggleClass("open-docs-main-nav");
+        setTableOfContentsVisibility()
+    });
+})(document, jQuery);
+
+function setDocsMainNavPosition() {
+    if ($(this).width() < 1280) {
```
sean1588, line 33: do these screen size related changes need to be done using javascript? We typically don't do this using javascript and this can make it difficult to reason about when some of the responsive related styling is handled in css and some in javascript.  Also, don't need to block on this, we can update this after. Though, if there is a reason this needs to be done this way ignore me.

themes/default/theme/src/ts/docs-main.ts
```diff
@@ -0,0 +1,83 @@
+var docsMainNavToggleWrapper = $(".docs-main-nav-toggle-wrapper");
```
sean1588, line 1: we typically always use `const` or `let` to declare variables. Using `var` is generally considered bad practice in javascript and typescript at this point and is just something that has been left around in the language since its early days. It is mainly because `var` variables do not respect block scoping and easily can create unwanted behavior if not used carefully. You can change all the vars here to let and it should continue to function as is.

sean1588 (commented):
overall this looks great!  just a couple of comments on some of the JS, but great job with all of this!!

themes/default/theme/src/ts/docs-main.ts
susanev: i think the answer is yes, it needs to be done this way. but half way through building this i realized i could do the open and close nav in css only. but i think that will require more work so id rather do this in another pr though. but id love to zoom and talk it through.

anita-trimbur (approved):
🚀 🚀 🚀 🚀

sean1588 (approved):

Issue #2960, opened 2023-05-31
Unexpected inconsistent font style on environmental variables page


mnlumi:
### What happened?

In Chrome or Firefox browser, the variables shown on [this environmental variables page](https://www.pulumi.com/docs/cli/environment-variables/) are oddly bold.
![env-var1](https://github.com/pulumi/pulumi-hugo/assets/118926824/f97d1516-68ea-4c42-b2a1-3e1361fd4b34)
![env-var2](https://github.com/pulumi/pulumi-hugo/assets/118926824/ff1f73d5-c1b8-4885-8b67-d6bc876b52f5)

### Expected Behavior

I expect the font style to be consistent with other pages.

### Steps to reproduce

Go to [this page](https://www.pulumi.com/docs/cli/environment-variables/) in Chrome or Firefox and look at the variables.

### Output of `pulumi about`

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #2976, opened 2023-06-03
new code font!

susanev:
## Description
@anita-trimbur take a look at the preview and let me know what you think
also happy to try other versions of this font, or any of the other fonts

fixes https://github.com/pulumi/pulumi-hugo/issues/2960

## Checklist:


anita-trimbur:
i really like it! what do you think? do you like the slab, or would you want to see it without?

susanev:
@anita-trimbur cool, lets ship it!

lukehoban:
The code font chosen here uses ligatures. I’m fairly strongly against that by default. For reasons similar to those articulated here: https://practicaltypography.com/ligatures-in-programming-fonts-hell-no.html.

anita-trimbur:
we can change from the term variant to the fixed variant to omit ligatures

susanev:
lol down with ligatures, yah ill remove them later today.

susanev:
@anita-trimbur this is updated, take another look, let me know what you think

anita-trimbur (approved):
still really like this! but am totally willing to defer if sean thinks otherwise. i've actually been mistakenly using iosevka sans the slab in figma, and i think that's a fine alternative if anyone is more inclined.

sean1588:
I am fairly neutral on this font change as this is quite a subjective area when it comes to the aesthetics of things. 

I think aesthetically it looks nice, however I do think something feels a bit off about it though if I am being completely honest, but not sure what it is. I think it may be because I am not used to reading code in this font since it is not a font I am used to seeing in either my terminal or my code editor so that may be why it has a bit of an awkward feel me. I could be an anomaly though and maybe I just need to get used to seeing it so I'm willing to defer to what you all think.

susanev:
@sean1588 @anita-trimbur kk let me try dropping the slab

PR #2979, opened 2023-06-04
Feature/OIDC aws docs

Yuhta28:
## Description
I read the doc of [Configuring OpenID Connect for AWS](https://www.pulumi.com/docs/pulumi-cloud/deployments/oidc/aws/).  
This document introduced the procedure for creating OIDC providers and IAM roles from the console. I have changed the procedure to use CloudFormation template files because I think it is more efficient.
## Checklist:


susanev:
@Yuhta28 thank you for the contribution, we will take a look as soon as we can!

@pgavlin can you take a look?

pgavlin (changes requested):
Thanks for the contribution!

I have two high-level pieces of feedback:
1. I do not think that we should remove the existing documentation. Even though it is a manual process, I believe that it is important that the docs exist.
2. I love the idea of automating this. Could we use a Pulumi YAML program rather than a CloudFormation template? Is there a benefit to the CF template over a Pulumi YAML program in this case?

Yuhta28:
@pgavlin 
Thanks for feedbacks.
### First feedback
I agree with your point. I will revise the manual processto remain as well. 

### Second feedback
I have the reason why I shoud use the CloudFormation in this case.  First, We don't need to install AWS CLI or AWS IAM AccessKey and SecretKey on local. We believe that CloudFormation can be used more easily because it can be deployed simply by uploading a template file to AWS.  
Second, if you manage multiple AWS accounts, you need to switch IAM AccessKey and SecretKey or AssumeRole and deploy Pulumi from local. I found it troublesome, so I recommended CloudFormation.

## Next Action
I'm going to revise the doc with your first feedback.  Please consider whether you agree with my opinion on your second feedback.

PR #2997, opened 2023-06-07
Add note about docker architecture on Pulumi Deployment

sicarul:
## Description

Adding note to specify how to build cross-platform docker images in Pulumi Deployments

## Checklist:

- [x] ~I have manually confirmed that all new links work.~
- [x] ~I added aliases (i.e., redirects) for all filename changes.~
- [x] ~If making css changes, I rebuilt the bundle.~

### DIFF
--- a/themes/default/content/docs/pulumi-cloud/deployments/reference.md
+++ b/themes/default/content/docs/pulumi-cloud/deployments/reference.md
@@ -226,3 +226,11 @@ By default, the deployment executor will attempt to install dependencies for you
 This is enabled by skipping the default dependency installation step (under Advanced Settings in the UI), and setting a few pre-run commands and environment variables.
 
 ![Pulumi UI - Node Version](../ui-node-version.png)
+
+### Building Docker images for another architecture
+
+By default, the deployment executor will only build Docker images for the x86_64 architecture, if you need to build images for another architecture (e.g. linux/arm64), you can add the architecture as a pre-run command such as:
+
+```
+docker run --privileged --rm tonistiigi/binfmt --install arm64
+```


themes/default/content/docs/pulumi-cloud/deployments/reference.md
```diff
@@ -226,3 +226,11 @@ By default, the deployment executor will attempt to install dependencies for you
 This is enabled by skipping the default dependency installation step (under Advanced Settings in the UI), and setting a few pre-run commands and environment variables.
 
 ![Pulumi UI - Node Version](../ui-node-version.png)
+
+### Building Docker images for another architecture
+
+By default, the deployment executor will only build Docker images for the x86_64 architecture, if you need to build images for another architecture (e.g. linux/arm64), you can add the architecture as a pre-run command such as:
```
AaronFriel, line 232: This isn't quite right, to run _steps_ in a Dockerfile for another architecture requires an emulator to be installed. There are many tools (Rust, Go, C with the appropriate tools installed, etc.) that support cross-compilation and can build images for `linux/arm64` without any emulator by using `FROM --platform=$BUILDPLATFORM ...` and creating a statically linked binary that runs on the target platform.

Suggest two changes:
1. Saying that the deployment engine uses x86-64 (aka `linux/amd64`) hosts.
2. Suggest that if users are building binaries for another platform, whether via the `Command` provider, `Docker`, they reference the documentation for building cross platform tools.
3. Suggest the command below to install emulators for Docker, linking to Docker documentation.

sicarul:
Closing in favor of a better alternative by someone who better understands this topic, sorry!

themes/default/content/docs/pulumi-cloud/deployments/reference.md
desteves: @sicarul were you able to update this section of the PR? I see this has been re-opened.

sicarul: @EvanBoyle told me to send the PR through Slack, i sent the updated changes merging the latest version

cnunciato:
I think we need someone to provide a bit more details so we bring this to completion. It sounds like we're close, but I want to be sure I understand what led up to this so we don't overcomplicate.

@sicarul Was the problem that your Pulumi program was producing (when running in Deployments) an x86_64 Docker image, but you needed the image to be built for some other platform? If so, I think Friel's #1 and #2 suggestions above are probably enough (i.e., to essentially point users to the Docker docs). Would you agree?

sicarul:
@cnunciato yes, i solved being able to build in arm64 by using the command proposed in the PR, which allows to host ECS Fargate services using the cheaper graviton processors. I don't see how linking to Docker docs will solve this question for future users that have this problem, is there a link that plainly explains the same solution i used, or any other that also works?

I don't understand why we can't simply put the command that we know solves this problem in the docs.

cnunciato:
> I don't understand why we can't simply put the command that we know solves this problem in the docs.

Sorry, I didn't mean to imply that -- totally open to adding this, I'm just looking to understand what we're actually addressing. Your note made it sound like you weren't sure yourself, so combined with Friel's comment, your follow-up comment, and then closing and reopening the PR without more explanation, it seemed reasonable to ask for a little more info.

What you've presented here essentially says, "If you want to build for another platform, run this command." That isn't quite enough detail, IMO. What does this command actually do? What is this third-party tool we're implicitly recommending users add to their pre-run scripts? We can certainly add this, provided we understand and can explain what it is, what it does, and why you'd use it (vs. some other approach). Hence Friel's recommendations (which again seem reasonable to me, minus the emulator bit, which I'm still not totally clear on).

Are you planning on applying any of these recommendations? If not, would you like us to pick it up from here?

sicarul:
@cnunciato as far as i know, this package that gets installed is a third party package, i've not seen it documented officially in Docker (but maybe i just didn't find it), I don't know much more than "add the emulator with this command and compiling to arm64 works on Pulumi Deploy", if someone else has a more detailed explanation i'm happy to have them edit this PR or open a new one, IMHO even a short note like the PR i sent is better than no guidance at all.

sicarul:
@cnunciato should we merge or close this PR?

PR #3003, opened 2023-06-07
Update inputs/outputs doc

susanev:
Reverts pulumi/pulumi-hugo#3002


jkodroff (approved):
This is a huge improvement! Thanks so much for taking the time to do this!

themes/default/content/docs/concepts/inputs-outputs/_index.md
```diff
@@ -13,16 +13,272 @@ aliases:
...
+var myId = new RandomId("mine", RandomIdArgs.builder()
+    .byteLength(8) // byteLength is an input
+    .build());
+
+```
+
+{{% /choosable %}}
+{{< /chooser >}}
+
+_Inputs_ are generally representations of the parameters to the underlying API call of any resource that Pulumi is managing.
```
AaronFriel, line 84:
```suggestion
_Inputs_ are the parameters to the resource that Pulumi is managing.
```

Simplify language here, we have providers like Command, Docker, etc., that make this unclear, and component providers muddy the water.

jaxxstorm: This was the main reason for the use of the term "generally", I do think it's important to keep something like this in

jkodroff: FWIW, I like that Lee's wording here provides that extra context.

themes/default/content/docs/concepts/inputs-outputs/_index.md
```diff
@@ -1056,3 +1579,256 @@ public static Output<List<String>> split(Output<String> str) {
...
+```java
+// TODO
+```
+
+{{% /choosable %}}
+{{< /chooser >}}
+
+However, this pattern isn't possible because resource names need to be available and known at runtime in order to construct the URN of the resource.
+
+If you really wish to pass an output from one resource as the resource name of another resource, you will need to use a {{< pulumi-apply >}}.
```
AaronFriel, line 1830: I think using `{{< pulumi-apply >}}` is not working outside of choosers, I get a broken link. Should we use `[apply](#apply)` in this document?

themes/default/content/docs/concepts/inputs-outputs/_index.md
```diff
@@ -13,16 +13,272 @@ aliases:
...
+
+```java
+// TODO
+```
+
+{{% /choosable %}}
+{{< /chooser >}}
+
+{{% notes %}}
+_Raw value_ in this document is used to describe a standard string, boolean, integer or other typed value in your language of choice. _Raw value_ is a way of differentiating these language specific values from Pulumi's asynchronous values.
```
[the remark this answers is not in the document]
jaxxstorm: I think "or other typed value" covers this, no?

themes/default/content/docs/concepts/inputs-outputs/_index.md
```diff
@@ -1056,3 +1579,256 @@ public static Output<List<String>> split(Output<String> str) {
...
+
+### String Interpolation
+
+When working with outputs and apply, you may see an error message like so:
+
+```
+Calling __str__ on an Output[T] is not supported. To get the value of an Output[T] as an Output[str] consider: 1. o.apply(lambda v: f"prefix{v}suffix") See https://pulumi.io/help/outputs for more details.
+```
+
+The reason this is happening is because the _Output_ value is being used before it has been resolved by from the API.
```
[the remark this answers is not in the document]
jaxxstorm: I don't understand this suggestion, is the original sentence not semantically correct?

themes/default/content/docs/concepts/inputs-outputs/_index.md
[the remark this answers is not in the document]
jaxxstorm: I have added "data structures" here to try make this clearer

themes/default/content/docs/concepts/inputs-outputs/_index.md
```diff
@@ -34,9 +290,69 @@ Because outputs are asynchronous, their actual raw values are not immediately av
 
 ## Apply
 
-To access the raw value of an output and transform that value into a new value, use {{< pulumi-apply >}}. This method accepts a callback that will be invoked with the raw value, once that value is available.
+To access the raw (or _resolved_) value of an output, use {{< pulumi-apply >}}. This method accepts a callback that will be invoked with the raw value, once that value is available.
```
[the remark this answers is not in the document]
jaxxstorm: I think this is covered in the next section, where we talk about assigning the value of an apply to a new var. This section is intentionally here to allow readers to be able to understand how to use values inside an apply, which is the biggest stumbling block.

themes/default/content/docs/concepts/inputs-outputs/_index.md
```diff
@@ -509,7 +825,7 @@ let certValidation = new aws.route53.Record("cert_validation", {
 
 ## Working with Outputs and Strings {#outputs-and-strings}
 
-Outputs that contain strings cannot be used directly in operations such as string concatenation. String interpolation lets you more easily build a string out of various output values, without needing {{< pulumi-apply >}} or [Output.all](/docs/reference/pkg/python/pulumi#pulumi.Output.all). You can use string interpolation to export a stack output, provide a dynamically computed string as a new resource argument, or even for diagnostic purposes.
+Outputs that resolve to strings cannot be used directly in operations such as string concatenation until the output has been resolved. In these scenarios, you'll need to resolve the values using {{< pulumi-apply >}}.
```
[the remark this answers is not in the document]
jaxxstorm: As below, not sure why we're straying away from the "resolved" language. Is that not correct? I think this suggested change removes the understanding gathered by how outputs work. If it's not semantically correct I can reword it, but I don't want to lose the readers ability to differentiate between plain and Outputty values, and resolved is a good way to do that

themes/default/content/docs/concepts/inputs-outputs/_index.md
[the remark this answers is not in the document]
Frassle: Technically we return this error even if the output _has_ resolved, better to be consistent in the error than to have it working sometimes based on async races. Having said that I think I'd prefer to keep some form of your original wording rather than bringing "apply" into this.

themes/default/content/docs/concepts/inputs-outputs/_index.md
AaronFriel: Makes sense to me

themes/default/content/docs/concepts/inputs-outputs/_index.md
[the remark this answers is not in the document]
AaronFriel: It's incorrect to say the user can call __str__ after it has been resolved. There's no way for them to do that in plain user code, they can use an "apply" to map over it, but they can't `await` the Output.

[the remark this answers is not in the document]
AaronFriel: To be really clear, I am nitpicking this:

> is being used before it has been resolved

So I can wait until after it's been resolved to call `__str__`? No!

themes/default/content/docs/concepts/inputs-outputs/_index.md
AaronFriel: That makes sense to me 👍

themes/default/content/docs/concepts/inputs-outputs/_index.md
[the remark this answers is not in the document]
jaxxstorm: Is it specifically the use of the word "resolve" we consider to be the problem here?

It's difficult to find a word that accurately describes the asynchronous nature of outputs (in fact, I use the word `async` elsewhere in the document, and by this logic I shouldn't have?) 

We need a way of expressing that outputs are async, and resolved is the word I use to express that. The proposed changes don't help that understanding. If we're not happy with use of the word resolved, I am at a loss on how explain it.

themes/default/content/docs/concepts/inputs-outputs/_index.md
```diff
@@ -13,16 +13,272 @@ aliases:
...
+{{% /choosable %}}
+{{< /chooser >}}
+
+In this case, Pulumi is taking the _output_ from one resource and using it as the _input_ to another resources
+
+## Outputs
+
+All resources created by Pulumi will have properties which are returned from the cloud provider API. These values are *outputs*.
+
+Outputs are values of type {{< pulumi-output >}}, which behave very much like [promises](https://en.wikipedia.org/wiki/Futures_and_promises) or [monads](https://en.wikipedia.org/wiki/Monad_(functional_programming)). This is necessary because outputs are not fully known until the infrastructure resource has actually completed provisioning, which happens *asynchronously*. Outputs are also how Pulumi tracks dependencies between resources. When an output from one resource has been returned from the cloud provider API, Pulumi can link the two resources together and pass it as the input to another resource.
```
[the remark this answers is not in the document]
susanev: gonna leave this as is for now, feel free to follow up with another pr

themes/default/content/docs/concepts/inputs-outputs/_index.md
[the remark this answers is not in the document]
susanev: gonna leave this as is for now, feel free to follow up with another pr

themes/default/content/docs/concepts/inputs-outputs/_index.md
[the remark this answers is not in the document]
susanev: gonna leave this as is for now, feel free to follow up with another pr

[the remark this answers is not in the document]
AaronFriel: Happy to take up in a follow up PR.

themes/default/content/docs/concepts/inputs-outputs/_index.md
```diff
@@ -13,30 +13,346 @@ aliases:
...
+```javascript
+// TODO
+```
+
+{{% /choosable %}}
+{{% choosable language typescript %}}
+
+```typescript
+const myId = new random.RandomId("mine", {
+    byteLength: 8, // byte_length is an input
```
lukehoban, line 33:
```suggestion
    byteLength: 8, // byteLength is an input
```

themes/default/content/docs/concepts/inputs-outputs/_index.md
```diff
@@ -13,30 +13,346 @@ aliases:
...
+{{% choosable language java %}}
+
+```java
+var myId = new RandomId("mine", RandomIdArgs.builder()
+    .byteLength(8) // byteLength is an input
+    .build());
+
+```
+
+{{% /choosable %}}
```
lukehoban, line 81: I think we also need YAML.

themes/default/content/docs/concepts/inputs-outputs/_index.md
```diff
@@ -13,30 +13,346 @@ aliases:
...
+
+```java
+// TODO
+```
+
+{{% /choosable %}}
+{{< /chooser >}}
+
+{{% notes %}}
+_Plain value_ in this document is used to describe a standard string, boolean, integer or other typed value or data structure in your language of choice. _Plain_ value_ is a way of differentiating these language specific values from Pulumi's asynchronous values.
```
lukehoban, line 154: I don’t think we should introduce the term “asynchronous values” here. Can’t we just call them “outputs”, which is the name they are referred as everywhere else?

themes/default/content/docs/concepts/inputs-outputs/_index.md
```diff
@@ -13,30 +13,346 @@ aliases:
...
+
+An {{< pulumi-output >}} value can resolve to a _plain value_. For more information, see [apply](#apply).
```
lukehoban, line 156: This feels confusing to say this way - and as though it’s something that suggests you should try to wait for these values to resolve and pass that “by far the easiest” kind of resulting value. I think we want to intentionally avoid encouraging that mental model. 

Instead - we should encourage that you can pass outputs as inputs. And if you want to transform the value of an output, you can do so within an apply. But I think it is somewhat actively dangerous to encourage the promise like model of awaiting or resolving these values - as that’s exactly what gets folks in trouble when working with Pulumi outputs.

lukehoban: Reading through the rest of this - I think we can just remove this paragraph. This concept is explained well in context later, after introducing outputs where it makes more sense.

themes/default/content/docs/concepts/inputs-outputs/_index.md
```diff
@@ -13,30 +13,346 @@ aliases:
...
 
-- [Apply](#apply): a callback that receives the raw value, and computes a new output
+- [Apply](#apply): a callback that receives the plain value, and computes a new output
 - [Lifting](#lifting): directly read properties off an output value
 - [Interpolation](#outputs-and-strings): concatenate string outputs with other strings directly
 
 ## Apply
 
-To access the raw value of an output and transform that value into a new value, use {{< pulumi-apply >}}. This method accepts a callback that will be invoked with the raw value, once that value is available.
+To access the _plain_ (or _resolved_) value of an output, use {{< pulumi-apply >}}. This method accepts a callback that will be invoked with the plain value, once that value is available.
```
lukehoban, line 293: I’m not sure it helps (or is necessary) to introduce “resolved” as well here?

themes/default/content/docs/concepts/inputs-outputs/_index.md
```diff
@@ -105,7 +421,7 @@ During some program executions, `apply` doesn’t run. For example, it won’t r
...
+If you have multiple outputs and need to use them together, the `all` function acts like an `apply` over many resources, allowing you to use multiple outputs when creating a new output. `all` waits for all output values to become available and then provides them as _plain values_ to the supplied callback. This function can be used to compute an entirely new output value, such as by adding or concatenating outputs from two different resources together, or by creating a new data structure that uses them. Just like with `apply`, the result of [Output.all](/docs/reference/pkg/python/pulumi#pulumi.Output.all) is itself an Output<T>.
```
lukehoban, line 424: Independent of this PR, but rereading this, there are multiple statements that are not true here. This implies an api that might be nice, but doesn’t exist (something that combines both all and apply into a single call).

themes/default/content/docs/concepts/inputs-outputs/_index.md
lukehoban, line 828: I think some of the removed text was actually very helpful for suggesting what you should use for common string manipulation options over outputs. The new text suggests you should “reaolve” to a plain string, which I think a lot of users will want to interpret as awaiting the output, then using the resulting value. This is exactly what you *cant* do in Pulumi, but which I worry we are suggesting with the “resolve” terminology.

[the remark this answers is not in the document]
Frassle: Without getting into theory words that most programmers won't really get I think "resolved" is probably the best we can do. I get Friels concern that outputs don't resolve like promises, but they _do_ internally resolve (sometimes to `unknown`) and it feels like the best common word to describe them.

If we could use accurate words here the whole page would pretty much just be "`Output<T>` is a monad" but even in FP land people frequently ask "how do I extract the value from a monad" and well you can't (in general) and for `Output<T>` doing so would be unsafe and also not give you back just a `T`.

DSoko2:
This is a great improvement! I think the "Outputs" section could be even a bit more accessible for some programmers by rearranging it a bit:

I would consider moving accessing plain values from outputs (the last par and bullet point list) up and the technical details about dependency tracking (staring from "Outputs are als how Pulumi tracks...") down. From my insight, often the first time people think about what Output is and looking it up is when they notice they cannot just access the plain value of properties on resources.

You may consider removing the "monad" reference. I feel like the word "monad" makes many (not only novice) developers feel uncomfortable as they do not have a firm understanding of what it means. It also does not add much here because the hint towards promises tells people that know about monads that Output is most likely monad-like, too. Some programmers may not read on in the section after bumping into the monad reference.

"Promises" is well known in JS/TS/Web stetting, developers in other environments may not know it well or would have an easier time with "Future". Perhaps putting the async fact first and then appending "like Futures or Promises" may be a solution.

themes/default/content/docs/concepts/inputs-outputs/_index.md
```diff
@@ -13,30 +13,346 @@ aliases:
...
+
+```java
+// TODO
+```
+
+{{% /choosable %}}
+{{< /chooser >}}
+
+{{% notes %}}
+_Plain value_ in this document is used to describe a standard string, boolean, integer or other typed value or data structure in your language of choice. _Plain_ value_ is a way of differentiating these language specific values from Pulumi's asynchronous values.
```
mootari, line 154:
```suggestion
_Plain value_ in this document is used to describe a standard string, boolean, integer or other typed value or data structure in your language of choice. _Plain value_ is a way of differentiating these language specific values from Pulumi's asynchronous values.
```

themes/default/content/docs/concepts/inputs-outputs/_index.md
jkodroff: How about something like:

```
If you need the plain values of multiple outputs at the same time (for example, to create an access policy or a connection string), use the `all` function. `all` takes a list of outputs and returns them as a single output. You can then call `apply` on the result of `all` to access all of the raw values you need in the same context:

// examples here in each language
```

themes/default/content/docs/concepts/inputs-outputs/_index.md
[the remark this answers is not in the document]
AaronFriel: I don't think it helps users though to tell them to "resolve" anything, instead, they need to use string operations _with apply_, or the other methods described in the paragraph below (string interpolation methods).

themes/default/content/docs/concepts/inputs-outputs/_index.md
jaxxstorm: I've removed references to asynchronous

themes/default/content/docs/concepts/inputs-outputs/_index.md
jaxxstorm: Remove all references to "resolved"

themes/default/content/docs/concepts/inputs-outputs/_index.md
jaxxstorm: I'm going to consider we address this in a further PR

themes/default/content/docs/concepts/inputs-outputs/_index.md
jaxxstorm: Added

PR #3024, opened 2023-06-09
scope css properly

shughes26:
## Checklist:

- [ ] I have reviewed the [style guide](https://github.com/pulumi/pulumi-hugo/blob/master/STYLE-GUIDE.md).
- [ ] If blogging, I have reviewed the [blogging guide](https://github.com/pulumi/pulumi-hugo/blob/master/BLOGGING.md).
- [ ] I have manually confirmed that all new links work.
- [ ] I added aliases (i.e., redirects) for all filename changes.
- [ ] If making css changes, I rebuilt the bundle.

PR #3025, opened 2023-06-09
Revert "Sarah/pulumiup schedule"

susanev:
Reverts pulumi/pulumi-hugo#3021

@shughes26 this touches global styles and breaks things in docs, we need to target only things in marketing for CSS changes like this; do you mind opening a new pr?


shughes26:
https://github.com/pulumi/pulumi-hugo/pull/3024 already up

shughes26:
Alright let's go ahead with this one, my local is broken for different reasons that I have to sort out before I can get a better fix up

shughes26 (approved):

PR #3044, opened 2023-06-12
fixing gcp template code

eydward:
## Description
The code snippets under “review project” don’t actually match the created templates; for now I just made sure everything is consistent, but it might be better to somehow set them to auto-sync together? In particular, the code snippets under “review project” for typescript/javascript/python all fail because they don’t set location, so I added the equivalent of “location=US” everywhere (as it is in the current template code).
Also currently the tutorial tells you to configure Pulumi to access Google Cloud before making a project, which violates causality

## Checklist:


susanev (approved):
i wonder if this also needs a change to the quickstart, @anita-trimbur thoughts?

PR #3054, opened 2023-06-13
Use H1s in results where they exist

cnunciato:
This change adds a bit of logic to show `h1` values as result titles when we have them. Intended to accompany https://github.com/pulumi/docs/pull/9306 (but can merge independently of it).


sean1588 (approved):
i think you have a conflict to resolve now, but looks good!

PR #3067, opened 2023-06-14
Fix occasional jitter, tighten up CSS and add comments

cnunciato:
This change includes a few more minor improvements, all follow-ups from https://github.com/pulumi/pulumi-hugo/pull/2868#issuecomment-1587760415:

* Fixes the jitter we sometimes see while a query is running
* Reduces the contrast on the in-page form element when the dialog is open
* Adds a suggestion for no-results searches
* Adds comments to the CSS to better explain what it's doing and why


themes/default/theme/src/ts/algolia/autocomplete.ts
```diff
@@ -226,14 +226,15 @@ function initAutocomplete(el: HTMLElement) {
                         </div>
                     `;
                 },
-                noResults: ({ html, state }) => {
+                noResults: ({ html, state, setCollections }) => {
```
cnunciato, line 229: Ah, `setCollections` is unused -- I'll remove.

themes/default/theme/src/ts/algolia/autocomplete.ts
```diff
@@ -226,14 +226,15 @@ function initAutocomplete(el: HTMLElement) {
...
                     if (!state.query) {
                         return;
                     }
 
                     return html`
-                        <img src="/images/search/no-results.svg"/>
+                        <p><img src="/images/search/no-results.svg"/></p>
                         <p>We couldn't find any results for <mark>${ state.query }</mark>.</p>
+                        <p>Try again using fewer or more general keywords.</p>
```
anita-trimbur, line 237:
```suggestion
                        <p>Try a more general term or use fewer keywords.</p>
```

anita-trimbur:
lgtm! i'll leave it sean for the actual code review

sean1588 (approved):

PR #3079, opened 2023-06-15
Update eks.md

omidraha:
## Description

Fixed `MalformedPolicyDocument` error for Python code:

With current python snippet code
```python
assume_role_policy = json.dumps({
    "Version": "2012-10-17",
    "Statement": [{
        "Action": "sts:AssumeRole",
        "Effect": "Allow",
        "Sid": None,
        "Principal": {
            "Service": "ec2.amazonaws.com",
        },
    }],
})
```

The below `MalformedPolicyDocument` error raises:

```
  aws:iam:Role (role1):
    error: 1 error occurred:
    	* creating IAM Role (role1-451c16c): MalformedPolicyDocument: Syntax error at position (1,113)
    	status code: 400, request id: ***
```
It will be fixed with change `Sid` argument value from `None` to the `""` an empty string.

## Checklist:

### DIFF
--- a/themes/default/content/docs/clouds/aws/guides/eks.md
+++ b/themes/default/content/docs/clouds/aws/guides/eks.md
@@ -1311,7 +1311,7 @@ assume_role_policy = json.dumps({
     "Statement": [{
         "Action": "sts:AssumeRole",
         "Effect": "Allow",
-        "Sid": None,
+        "Sid": "",
         "Principal": {
             "Service": "ec2.amazonaws.com",
         },


cnunciato:
Nice catch, thanks @omidraha!

themes/default/content/docs/clouds/aws/guides/eks.md
```diff
@@ -1311,7 +1311,7 @@ assume_role_policy = json.dumps({
     "Statement": [{
         "Action": "sts:AssumeRole",
         "Effect": "Allow",
-        "Sid": None,
+        "Sid": "",
```
cnunciato, line 1314: @omidraha Since this property is apparently [optional](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_sid.html), what do you think about omitting it altogether? When I do that, it looks like the policy gets created successfully (and without a `Sid`) -- and when I then change the `Sid` to `""`, Pulumi detects no changes.

```suggestion
```

omidraha: In the tests related to the library, an empty string has been used as a test case.

You can find an example of this usage at the following link:

https://github.com/pulumi/pulumi-aws/blob/c0dcccb3d7aef5bd661e2f74ec4c5361273d7c12/sdk/python/pulumi_aws/iam/role_policy.py#L191

May be it have been preferred to explicitly include an empty string for clarity in the policy structure.

And also, the record created in the Amazon console will be different.

- 
-

PR #3085, opened 2023-06-18
config.md: pulumi:disable-default-providers is not a list element

marco-m:


interurban (approved):
Thank you @marco-m for the contribution!

PR #3093, opened 2023-06-21
Remove programming-model docs alias

cnunciato:
This change removes an alias to a placeholder page that we still need in order to properly redirect client-side requests based on HTML anchors. We have a number of places still linking to this page, at this particular URL (`/docs/intro/concepts/programming-model`), so we need this page to continue to exist in order to handle those requests.

### DIFF
--- a/themes/default/content/docs/concepts/how-pulumi-works.md
+++ b/themes/default/content/docs/concepts/how-pulumi-works.md
@@ -15,7 +15,6 @@ aliases:
 - /docs/tour/programs-resources/
 - /docs/tour/programs-stacks/
 - /docs/intro/concepts/how-pulumi-works/
-- /docs/intro/concepts/programming-model/
 ---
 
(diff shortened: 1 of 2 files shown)


themes/default/content/docs/concepts/how-pulumi-works.md
```diff
@@ -15,7 +15,6 @@ aliases:
 - /docs/tour/programs-resources/
 - /docs/tour/programs-stacks/
 - /docs/intro/concepts/how-pulumi-works/
-- /docs/intro/concepts/programming-model/
```
cnunciato, line 18: Removing this line prevents the How Pulumi Works page from "squatting" the `/docs/intro/concepts/programming-model` URL.

themes/default/content/docs/concepts/programming-model.md
```diff
@@ -1,7 +1,6 @@
 ---
 title: "Programming Model"
-aliases:
-- /docs/reference/programming-model/
+url: /docs/intro/concepts/programming-model
```
cnunciato, line 3: We have to add an explicit `url` param here to tell Hugo to render this page at this particular path, since it no longer exists at this location in the `content` tree.

themes/default/content/docs/concepts/programming-model.md
```diff
@@ -58,4 +57,4 @@ block_external_search_index: true
 -->
 <meta http-equiv="refresh" content="4; url=/docs/concepts">
 
-This content has moved. Redirecting to [Architecture &amp; Concepts](/docs/concepts)...
+This content has moved. Redirecting to [Pulumi Concepts](/docs/concepts)....
```
cnunciato, line 60: Page is now called something else, so just updating that also.

scottslowe (approved):
LGTM---appreciate the extra context for each of the changes, makes the PR easier to grok.

Issue #3094, opened 2023-06-21
GCP getting-started docs contain a section that doesn't work


cnunciato:
On this page:

https://www.pulumi.com/docs/clouds/gcp/get-started/begin/

... there's a shortcode being used (`configure-gcp`) that happens to be shared with the GCP install/config page in the Registry. Unfortunately, this shortcode emits some content that doesn't work in the getting-started context -- specifically this:

```
$ pulumi config set gcp:project pulumi-development
error: no Pulumi.yaml project file found. If you have not created a project yet, use `pulumi new` to do so: no project file found
``` 

This fails because at this point in the tutorial, the user hasn't yet created a Pulumi project.

We should fix this, probably by removing this shortcode from the page and moving the shortcode itself (which has no other use in this repo) to the Registry.

PR #3096, opened 2023-06-21
Pull shared content out of the GCP begin page

cnunciato:
This change removes the use of a shortcode from the GCP getting-started guide because the shortcode's content doesn't quite work in this context. The change restores the content that belonged to the shortcode before this additional content (which is otherwise used only [in](https://github.com/pulumi/registry/blob/30b15aaa0aee99e8f734f252be1da57b5831b358/themes/default/content/registry/packages/gcp/installation-configuration.md?plain=1#L23) [the](https://github.com/pulumi/registry/blob/30b15aaa0aee99e8f734f252be1da57b5831b358/themes/default/content/registry/packages/gcp-global-cloudrun/installation-configuration.md?plain=1#L18) [Registry](https://github.com/pulumi/registry/blob/30b15aaa0aee99e8f734f252be1da57b5831b358/themes/default/content/registry/packages/google-native/installation-configuration.md?plain=1#L20)) was added.

Fixes https://github.com/pulumi/pulumi-hugo/issues/3094.

As a follow-up, we should move this shortcode to the Registry. This'll need to be done in two steps (first added there, then removed here) so as not to break the build.


scottslowe (approved):
LGTM

PR #3098, opened 2023-06-22
separate docs builds deploy to test environment

sean1588:
fixes: https://github.com/pulumi/home/issues/2875

This PR configures hugo to deploy directly to the test environment when a PR gets merged to master. The current workflows will continue to function as they are now. This adds an additional workflow that gets runs when a PR is merged to master to do the deploy. This new workflow will only affect the test environment right now. All current behavior with getting content deployed to prod from this repo will continue to functioning as it is now. There is also a [PR](https://github.com/pulumi/docs/pull/9365) in docs that this is dependent on that must merge in before this one does.

Changes made:
- testing github environment
- add cypress tests
- add await script to await any deployments currently in progress both in this repo and the docs repo
- add pulumi login script and script to run pulumi commands

example run of a working deployment: https://github.com/pulumi/pulumi-hugo/actions/runs/5349849358
cdn endpoint of test env: https://d298m5mqqr8wsc.cloudfront.net/

## Checklist


scripts/ci/common.sh
```diff
@@ -15,6 +15,21 @@ aws_region() {
...
 
+clone_docs_infrastructure() {
+    git clone "https://github.com/pulumi/docs.git" --no-checkout docsrepo --depth 1
+    pushd docsrepo
+    git sparse-checkout init --cone
+    git sparse-checkout set infrastructure
+    git checkout
+    yarn install
+    # temporarily here until merging docs updates into master
+    curl -o infrastructure/Pulumi.www-testing.yaml -L "https://raw.githubusercontent.com/pulumi/docs/b83121c58ba66f5b53a1a8d36945a9d47c91f674/infrastructure/Pulumi.www-testing.yaml"
```
sean1588, line 26: i will remove these once the docs PR is merged

cypress/videos/site_spec.js.mp4
cnunciato, line 1: This folder should not be checked in.

sean1588: ohh good catch. thank you! will remove!

PR #3099, opened 2023-06-22
Fix a few papercuts with the Pulumi AI UI

cnunciato:
Addresses a handful of minor input-related issues, including:

* Extending the input maxlength to ~100K~ 10K
* Capping the number of textarea rows to 6
* Only submitting multi-line prompts input on command/control+enter
* Resetting the row-count to 1 on completion (since we also clear out the input field)

Fixes https://github.com/pulumi/pulumi.ai/issues/42.


zchase:
@cnunciato Looks great!! Thanks for the help!

themes/default/theme/stencil/src/components/ai/client.ts
```diff
@@ -19,7 +19,7 @@ import { parseCookie } from "../../util/util";
 export class PulumiAIClient {
     private socket: WebSocket;
 
-    public MAX_PROMPT_LENGTH = 1024;
+    public MAX_PROMPT_LENGTH = 102400;
```
zchase, line 22: I think this is too high of setting as a certain point the prompt will be bigger than the token limit of ChatGPT. `gpt-4` has a token limit of 8k and `gpt-3.5` has a token limit of 4k (a token is about 4 characters). So at most this could probably be ~24K, but we also still have to account for the fact that this input is injected into a larger prompt so we should constrain it a bit more.

zchase (approved):
LGTM outside the comment about the max length being too large.

themes/default/theme/stencil/src/components/ai/client.ts
cnunciato: Ah, good to know, thanks. What do you think of, say, 10K then?

cnunciato: I'll go ahead and do that -- we can come back to it if we notice any issues with it. Thanks again for the catch.

PR #3105, opened 2023-06-26
Update API Docs for Resource Search cursor pagination

djgrove:
## Description
Update the API docs to correct resource search cursor query param syntax.

## Checklist:

- [ ] ~~If blogging, I have reviewed the [blogging guide](https://github.com/pulumi/pulumi-hugo/blob/master/BLOGGING.md).~~
- [ ] ~~I have manually confirmed that all new links work.~~
- [ ] ~~I added aliases (i.e., redirects) for all filename changes.~~
- [ ] ~~If making css changes, I rebuilt the bundle.~~

### DIFF
--- a/themes/default/content/docs/pulumi-cloud/cloud-rest-api/_index.md
+++ b/themes/default/content/docs/pulumi-cloud/cloud-rest-api/_index.md
@@ -2768,8 +2768,7 @@ Allowed values: created, custom, delete, id, modified, module, name, package, pa
 
 **size**: How many results to return at a time.
 
-**paginate**: If requesting multiple pages of results, the pagination method to use. Should be "cursor" or "page"
-This tells the API that you would like to receive a continuation token for paging additional results.
+**paginate**: (Optional) The pagination method for retrieving multiple pages of results. One of `page` (default) for an ordered list of the first 10k results, or `cursor` for iterating through the entire unordered list of results.
 
 **page**: The page of results to return.
 The page parameter can only be used to fetch up 10,000 resources. If a query matches more than 10,000 resources, the "cursor" parameter should be used instead.
--- a/themes/default/content/docs/pulumi-cloud/cloud-rest-api/_index.md
+++ b/themes/default/content/docs/pulumi-cloud/cloud-rest-api/_index.md
@@ -2740,7 +2740,7 @@ Search for resources belonging to the given organization.
 | sort       | query | array[string]    | false    | Results are returned sorted by this field value.                                                                                         |
 | asc        | query | boolean          | false    | Whether to return results in ascending or descending sort order.                                                                         |
 | size       | query | integer          | false    | How many results to return at a time.                                                                                                    |
-| paginate   | query | "cursor", "page" | false    | If requesting multiple pages of results, the pagination method to use.
+| paginate   | query | string | false    | If requesting multiple pages of results, the pagination method to use. Valid values are "cursor" and "page", defaults to "page".
 | page       | query | number           | false    | The page of results to return.                                                                                                           |
 | cursor     | query | string           | false    | A continuation token for pagination.                                                                                                     |
 | facet      | query | array[string]    | false    | If provided, an aggregation will be returned with the top-5 values for the given facet, along with how many resources have those values. |
--- a/themes/default/content/docs/pulumi-cloud/cloud-rest-api/_index.md
+++ b/themes/default/content/docs/pulumi-cloud/cloud-rest-api/_index.md
@@ -2735,15 +2735,16 @@ Search for resources belonging to the given organization.
 
 | Name       | In    | Type          | Required | Description                                                                                                                              |
 |------------|-------|---------------|----------|------------------------------------------------------------------------------------------------------------------------------------------|
-| org        | path  | string        | true     | Name of the organization to search.                                                                                                      |
-| query      | query | string        | false    | The search query to execute. If omitted all resources are returned (subject to any pagination limits).                                   |
-| sort       | query | array[string] | false    | Results are returned sorted by this field value.                                                                                         |
-| asc        | query | boolean       | false    | Whether to return results in ascending or descending sort order.                                                                         |
-| size       | query | integer       | false    | How many results to return at a time.                                                                                                    |
-| page       | query | number        | false    | The page of results to return.                                                                                                           |
-| continue   | query | string        | false    | A continuation token for pagination.                                                                                                     |
-| facet      | query | array[string] | false    | If provided, an aggregation will be returned with the top-5 values for the given facet, along with how many resources have those values. |
-| properties | query | boolean       | false    | Whether to include resource properties in results. Not supported for all subscriptions.                                                  |
+| org        | path  | string           | true     | Name of the organization to search.                                                                                                      |
+| query      | query | string           | false    | The search query to execute. If omitted all resources are returned (subject to any pagination limits).                                   |
+| sort       | query | array[string]    | false    | Results are returned sorted by this field value.                                                                                         |
+| asc        | query | boolean          | false    | Whether to return results in ascending or descending sort order.                                                                         |
+| size       | query | integer          | false    | How many results to return at a time.                                                                                                    |
+| paginate   | query | "cursor", "page" | false    | If requesting multiple pages of results, the pagination method to use.
+| page       | query | number           | false    | The page of results to return.                                                                                                           |
+| cursor     | query | string           | false    | A continuation token for pagination.                                                                                                     |
+| facet      | query | array[string]    | false    | If provided, an aggregation will be returned with the top-5 values for the given facet, along with how many resources have those values. |
+| properties | query | boolean          | false    | Whether to include resource properties in results. Not supported for all subscriptions.                                                  |
 
 #### Detailed descriptions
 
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/pulumi-cloud/cloud-rest-api/_index.md
```diff
@@ -2735,15 +2735,16 @@ Search for resources belonging to the given organization.
...
+| org        | path  | string           | true     | Name of the organization to search.                                                                                                      |
+| query      | query | string           | false    | The search query to execute. If omitted all resources are returned (subject to any pagination limits).                                   |
+| sort       | query | array[string]    | false    | Results are returned sorted by this field value.                                                                                         |
+| asc        | query | boolean          | false    | Whether to return results in ascending or descending sort order.                                                                         |
+| size       | query | integer          | false    | How many results to return at a time.                                                                                                    |
+| paginate   | query | "cursor", "page" | false    | If requesting multiple pages of results, the pagination method to use.
```
EvanBoyle, line 2743: We should indicate that the default value is `page`

themes/default/content/docs/pulumi-cloud/cloud-rest-api/_index.md
```diff
@@ -2759,18 +2760,27 @@ If omitted, results are sorted according to their search relevance. If there is
...
 
+Special consideration needs to be given if using this in tandem with the **cursor** parameter, see below.
+
 Allowed values: created, custom, delete, id, modified, module, name, package, parent.urn, pending, project, protected, provider.urn, stack, type, urn.
 
 **asc**: Whether to return results in ascending or descending sort order.
 
 **size**: How many results to return at a time.
 
+**paginate**: If requesting multiple pages of results, the pagination method to use. Should be "cursor" or "page"
```
EvanBoyle, line 2771:
```
(Optional) the pagination method for retrieving multiple pages of results. One of `page` (default) for an ordered list of the first 10k results, or `cursor` for iterating through the entire unordered list of results.
```

themes/default/content/docs/pulumi-cloud/cloud-rest-api/_index.md
```diff
@@ -2759,18 +2760,27 @@ If omitted, results are sorted according to their search relevance. If there is
...
-The page parameter can only be used to fetch up 10,000 resources. If a query matches more than 10,000 resources, the "continue" parameter should be used instead.
+The page parameter can only be used to fetch up 10,000 resources. If a query matches more than 10,000 resources, the "cursor" parameter should be used instead.
 
-**continue**: A continuation token for pagination.
+**cursor**: A continuation token for pagination.
 Only available on Enterprise plans.
 
+To receive a cursor from the API, you need to issue a `paginate="cursor"` query parameter with your initial request. This informs the API that it should return a cursor as part of the JSON response.
+
+Undefined behavior can occur if the query results are impacted during pagination, for example as the result of a completed stack update. This also means that sort order must be deterministic to yield expected results, as cursor pagination respects sorting.
```
EvanBoyle, line 2782: > sort order must be deterministic to yield expected results, as cursor pagination respects sorting

I don't think I understand what this means. Cursor pagination does not respect sorting, right? I thought part of the change we made was stripping sorting from cursor based pagination to make results more stable. 

Probably we should say something simpler: 

```
Resource Search uses an eventually consistent data store and code that calls this API must account for that. For instance, stack updates trigger resources to be reindexed which may cause results to change while you are following the cursor (i.e. you may see duplicate results while following the cursor if updates are occurring frequently).
```

themes/default/content/docs/pulumi-cloud/cloud-rest-api/_index.md
```diff
@@ -2759,18 +2760,27 @@ If omitted, results are sorted according to their search relevance. If there is
...
+
 **page**: The page of results to return.
-The page parameter can only be used to fetch up 10,000 resources. If a query matches more than 10,000 resources, the "continue" parameter should be used instead.
+The page parameter can only be used to fetch up 10,000 resources. If a query matches more than 10,000 resources, the "cursor" parameter should be used instead.
 
-**continue**: A continuation token for pagination.
+**cursor**: A continuation token for pagination.
 Only available on Enterprise plans.
 
+To receive a cursor from the API, you need to issue a `paginate="cursor"` query parameter with your initial request. This informs the API that it should return a cursor as part of the JSON response.
```
EvanBoyle, line 2780: How about something like:

```
`paginate=cursor` returns a continuation token that can be used to page through all results of a query, not just the first 10,000. Cursor pagination will ignores any sort parameters in the query and instead returns results sorted by a stable resource ID.
```

themes/default/content/docs/pulumi-cloud/cloud-rest-api/_index.md
```diff
@@ -2735,15 +2735,16 @@ Search for resources belonging to the given organization.
...
-| continue   | query | string        | false    | A continuation token for pagination.                                                                                                     |
-| facet      | query | array[string] | false    | If provided, an aggregation will be returned with the top-5 values for the given facet, along with how many resources have those values. |
-| properties | query | boolean       | false    | Whether to include resource properties in results. Not supported for all subscriptions.                                                  |
+| org        | path  | string           | true     | Name of the organization to search.                                                                                                      |
+| query      | query | string           | false    | The search query to execute. If omitted all resources are returned (subject to any pagination limits).                                   |
+| sort       | query | array[string]    | false    | Results are returned sorted by this field value.                                                                                         |
```
cnunciato, line 2740:
```suggestion
| sort       | query | array[string]    | false    | The field(s) by which to sort results.                                                                                          |
```

themes/default/content/docs/pulumi-cloud/cloud-rest-api/_index.md
```diff
@@ -2735,15 +2735,16 @@ Search for resources belonging to the given organization.
...
+| asc        | query | boolean          | false    | Whether to return results in ascending or descending sort order.                                                                         |
```
cnunciato, line 2741: If this param is optional, does that mean results are sorted _descendingly_ by default? (For example, if `sort` field is provided but this param isn't?)

themes/default/content/docs/pulumi-cloud/cloud-rest-api/_index.md
```diff
@@ -2759,18 +2760,27 @@ If omitted, results are sorted according to their search relevance. If there is
 
 If specified more than once, the first parameter is the primary sort order and subsequent parameters control additional sorting criteria.
 
+Special consideration needs to be given if using this in tandem with the **cursor** parameter, see below.
```
cnunciato, line 2763:
```suggestion
Special consideration needs to be given if using this in tandem with the `cursor` parameter, see below.
```

themes/default/content/docs/pulumi-cloud/cloud-rest-api/_index.md
cnunciato: +1 to Evan's comment, but I'd capitalize `The pagination`.

themes/default/content/docs/pulumi-cloud/cloud-rest-api/_index.md
```diff
@@ -2759,18 +2760,27 @@ If omitted, results are sorted according to their search relevance. If there is
...
 **asc**: Whether to return results in ascending or descending sort order.
 
 **size**: How many results to return at a time.
 
+**paginate**: If requesting multiple pages of results, the pagination method to use. Should be "cursor" or "page"
+This tells the API that you would like to receive a continuation token for paging additional results.
+
 **page**: The page of results to return.
-The page parameter can only be used to fetch up 10,000 resources. If a query matches more than 10,000 resources, the "continue" parameter should be used instead.
+The page parameter can only be used to fetch up 10,000 resources. If a query matches more than 10,000 resources, the "cursor" parameter should be used instead.
```
cnunciato, line 2775:
```suggestion
The `page` parameter can only be used to fetch up 10,000 resources. If a query matches more than 10,000 resources, the `cursor` parameter should be used instead.
```

themes/default/content/docs/pulumi-cloud/cloud-rest-api/_index.md
cnunciato: +1, but with `Cursor pagination ignores`.

themes/default/content/docs/pulumi-cloud/cloud-rest-api/_index.md
cnunciato: We should probably make this `string` and indicate that valid values are `cursor` and `page`.

themes/default/content/docs/pulumi-cloud/cloud-rest-api/_index.md
djgrove: This is not entirely correct. We do respect any custom sorts applied by the customer--it is the relevance sorting that they see for a query issued in the UI that we drop if paginating.

This decision was made because sorting is _fine_, provided it is deterministic/stable. It was the relevance sort, which can yield different values on every request regardless of underlying data changes, that caused the bug and that we dropped.

All of our other fields are going to be stable unless a stack update finalizes as the pagination requests are in-flight, which I've called out. Happy to reword this.

themes/default/content/docs/pulumi-cloud/cloud-rest-api/_index.md
djgrove: Relates to [my comment](https://github.com/pulumi/pulumi-hugo/pull/3105/files#r1244195779) above

themes/default/content/docs/pulumi-cloud/cloud-rest-api/_index.md
djgrove: [7757cfa](https://github.com/pulumi/pulumi-hugo/pull/3105/commits/7757cfa27983f0980b2ca5f483973854ce3f8f67)

themes/default/content/docs/pulumi-cloud/cloud-rest-api/_index.md
djgrove: [152f60b](https://github.com/pulumi/pulumi-hugo/pull/3105/commits/152f60b25d5c2cfd937b87f3a16207f9444ae711)

themes/default/content/docs/pulumi-cloud/cloud-rest-api/_index.md
EvanBoyle: IMO this is too complicated and unnecessary to explain in detail. I would just say something accurate but vague instead:

```
Cursor pagination does not guarantee that results are returned in sorted order.
``

themes/default/content/docs/pulumi-cloud/cloud-rest-api/_index.md
blampe: Yes, descending by default. Added clarification.

meagancojocar (approved):
Love it!

djgrove (commented):
Approved @blampe's changes. Lets roll!

PR #3108, opened 2023-06-26
Document how to create content for search

cnunciato:


scottslowe (approved):
This is much needed, thank you! One minor suggestion to consider, but no need to block on it: would it be possible, or even desirable, to issue _stronger_ recommendations on the use of `search.keywords` in the front matter? Right now it seems more like a suggestion, and it feels like it should be a definite requirement.

PR #3115, opened 2023-06-27
add content width controls

sean1588:
## Checklist:


anita-trimbur:
ahhh this works great!! ✨ here are the few things i noticed:

- [ ] the toc isn't hugging the main content area any more, both when the width is expanded and the default width. on really wide screens, it gets super far away.
- [ ] can we add in a hover state? gray800 default and gray900 on hover
- [ ] can we also add a tooltip on hover? `Expand content width` and `Shrink content width`
- [ ] registry browse page resizes funky when the content width is expanded and then shrunk
- [ ] in registry, the expanded width is a little too wide, it pushes the package card all the way to right edge of the screen
- [ ] this might not be possible, but just wondering - sometimes on really long docs when you expand the content width, you lose the section you were on since the content is taking up less vertical space. is there any way we could make sure that the the active section is still visible when toggling the content width?

sean1588:
@anita-trimbur - I addressed most of the comments. Let me know what you think. Concerning the tool-tip though, I can follow up on that if that sounds good, since the current tool tip we have is rendering in strange spots falling off the screen, since the button is near the edge and I'll need to build in some extra customizations to our tooltip component to get it to render where we want.

For the last one about maintaining the same spot in the docs, it will be a bit hard to get this right in a lot of cases, since it is hard to determine where the spot should be once expanded/collapsed. If all the docs were only text, then this would be easier to determine, but since some contain code blocks, images, etc it can be hard to predict how much to scroll up/down.

anita-trimbur:
@sean1588 yeah totally cool with leaving the tooltip for now! def want to come back to it when we can since this isn't the most familiar pattern, but for sure don't need to block on it. and totally understood on the spot saving, i figured that might not be super doable. 

just a couple last things as of the most recent preview:
- [ ] max content width in the registry seems to be 1000px vs 1200px in docs
- [ ] something is still strange on registry browse - selecting the expand icon slightly shrinks the width and selecting the shrink icon slightly expands the width

otherwise everything looks great! will approve now just so youre not stuck waiting on me

anita-trimbur (approved):

sean1588:
@anita-trimbur - i fixed those registry issues. Only thing I'll call out with registry is that the content width will be 1200px assuming the screen width is wide enough to accomodate it, if not it will just get as wide as it can given the space. This happens at some screen sizes that are just slightly wider than 1526px,since  there is not enough width to fully accomodate the 1200px content width area along with the right nav as well. So in this case it will just get as wide as the user's screen size will allow.

sean1588:
also I opened an issue for the tooltip
https://github.com/pulumi/pulumi-hugo/issues/3134

zchase (approved):

PR #3118, opened 2023-06-27
Don't send Sentry events for browsers we don't support

cnunciato:
We've seen an uptick recently in Sentry errors originating from browsers that look like scrapers. This change adds a quick check to look at the user-agent string and bypass sending events to Sentry for browsers we know are problematic.


scottslowe (approved):
LGTM

PR #3123, opened 2023-06-29
Autogenerated Youtube video pages

aaronkao:
## Description

Auto generate pages for Youtube videos on PulumiTV and allow it to be indexable by Algolia search


desteves:
I noticed this open PR is labeled with `do-not-merge`. Could y’all let me know if we still need this open or are we good to close this? @aaronkao @cnunciato

PR #3131, opened 2023-07-06
Fix C# code example

niklasravnsborg:
Seems like some Go syntax accidentally made it into the C# example snippet.


cnunciato (approved):
Nice spot, @niklasravnsborg! Thanks -- really appreciate the fix. ⭐

PR #3135, opened 2023-07-07
Add workshop meta images

SaraDPH:
## Checklist:


themes/default/content/resources/azure-github-workshop/index.md
```diff
@@ -3,6 +3,7 @@
 title: "Getting Started with Azure IaC and Review Stacks"
 title_tag: "Getting Started with Azure IaC and Review Stacks (July 2023 Workshop)"
 meta_desc: "In this workshop, we’ll explore the new Azure Native v2 provider and Pulumi Review Stacks on GitHub"
+meta_image: "/images/resources/Azure-IaC-Review-Stacks.png"
```
zchase, line 6: I'd personally prefer that these images be stored in the directory as the content itself (`index.md` file). That we don't end up with long list of images in the `/resources` directory and it is clear which image maps to which workshop.

zchase (changes requested):

themes/default/content/resources/azure-github-workshop/index.md
SaraDPH: > I'd personally prefer that these images be stored in the directory as the content itself (`index.md` file). That we don't end up with long list of images in the `/resources` directory and it is clear which image maps to which workshop.

I think @cnunciato said it couldn't... I'll do whatever y'all think is best, my goal is just to get a meta image, where it should go it's up to you. :)

SaraDPH: Why is the Kubernetes workshop image not showing?!

sean1588 (approved):

themes/default/content/resources/azure-github-workshop/index.md
cnunciato: @zchase Yep, agreed -- we tried this first, but saw that the templates weren't set up for it yet, so didn't want to block on it. But we'll come back and make that work.

PR #3137, opened 2023-07-08
Update access-tokens.md

AaronFriel:
Fixes a typo and changes the table to increase in scope from left to right -> personal, team, organization.


ericrudder:
i think there are still some typos in the table?

search for "yag" ... i think these should be "tag" ... i think the singular/plural stuff also needs to be made consistent across the rows (e.g. tag or tags)

AaronFriel:
@ericrudder It looks like the singular/plural is bubbling up REST API semantics (item vs list) to this table. I don't think I'm the right person to make a call on representation here. @EvanBoyle or @caseyyh do you want to edit this branch or follow up on that?

meagancojocar (approved):
We can address the issue Eric identified in a separate PR.

Issue #183, opened 2021-05-03
social imgs are cut off in bad ways


susanev:
the social img for linkedin is cut off, we should prolly just use the regular wordmark logo for that space

<img width="535" alt="linkedin" src="https://user-images.githubusercontent.com/5489125/116920922-18c1c500-ac08-11eb-8fe3-2ca23eb73849.png">

cnunciato:
We should do this if we can, but it's not clear to me we can specify a different `og` (or other) image for LinkedIn previews. https://www.linkedin.com/help/linkedin/answer/46687/making-your-website-shareable-on-linkedin?lang=en

cnunciato:
We'll probably want to update the image we use for Greenhouse listings also:

![image](https://user-images.githubusercontent.com/274700/122318386-ceed1f80-ced3-11eb-9a35-ffaa98649c53.png)

susanev:
as an additional note, this is how our greenhouse links appear on twitter

<img width="533" alt="greenhouse-twitter" src="https://user-images.githubusercontent.com/5489125/122627679-f542c480-d065-11eb-89bb-84c2a8d38cfe.png">

kimberleyamackenzie:
Pushing this to the next milestone, as we're pulling in https://github.com/pulumi/pulumi-service/issues/6596 as higher priority.  

A few findings for when this gets picked back up:
- We can't set multiple og:image tags for different sites, but in theory we should be able to get one image with the right ratios to work everywhere and/or set something specific up for twitter if that's the one outlier because they have their own tags.
- Testing this locally and with the site previews doesn't work :(.  Probably have to set up ngrok to test this, but haven't verified that works.
- For linkedin, image must be a 1.91:1 ratio (1200x627 px) and min 200px width
- To test + verify metadata without actually posting https://www.linkedin.com/post-inspector/

SaraDPH:
Please add **<meta property="og:image:alt" content="Pulumi logo" />** and **<meta name="author" content="Pulumi">**

For the image can we do a small test to see if LinkedIn picks up the correct array? 

The goal is to keep the same image but add the dimensions, so it knows it is square, 
_<meta property="og:image" content="https://www.pulumi.com/logos/brand/twitter-card.png">_
**<meta property="og:image:width" content="300" />
<meta property="og:image:height" content="300" />**

And add a second image for LinkedIn with its dimensions:

_<meta property="og:image" content="[image below]" />_
**<meta property="og:image:width" content="600" />
[...]

PR #3140, opened 2023-07-10
Use a rectangular default og:image

cnunciato:
Adds a 1200x627 image to use as our default `og:image`, replacing the square image we've been using up to now. (Leaves that image in place, however, in case there's anything 
else out there that might be using it.)

Fixes https://github.com/pulumi/pulumi-hugo/issues/183.


zchase (approved):

SaraDPH:
Odd, it still appears as a square but the logo gets cutoff. That may be why the vertical logo was used.

cnunciato:
@SaraDPH Yeah, I've been trying different configurations here. I wanted to be able to continue using the smaller, square Twitter image by default (for pages that don't have an explicit `meta_image` set), but have LinkedIn use the larger rectangular one as its default, to get around the cropping issue. I think I have that now:
 
image

image

image

PR #3155, opened 2023-07-13
Ease docs navigation experience by allowing ToC to be scrolled separately

### LINKED ISSUE #3132: On-this-page nav no longer scrolls, making items inaccessible
The right-hand column of the docs navigation seems to have lost its ability to scroll independently. This is most problematic in the Registry, where, because the package card pushes this navigation farther down the page, the lowermost items drift offscreen, making its items inaccessible until the user scrolls to the bottom of the page: 

https://github.com/pulumi/pulumi-hugo/assets/274700/75e0cca2-9187-4a7f-9ae2-570785eaf63b

We should allow this column to scroll independently as the left-hand one does.

skooch:
## Description

Small change to set height of ToC wrapper class to allow scrolling in _docs-main.scss

Currently navigating docs is a bit frustrating because your ToC can overflow offscreen as you scroll down long pages.

Ideally we'd make the ToC individually scrollable, remove/style the scrollbar for visual consistency, and also have it scroll to follow your position relative to the content. For now, this at least makes the ToC navigable again.

#### Before

image

#### After

image

## Checklist:

Fixes #3132.

### DIFF
--- a/themes/default/theme/src/scss/docs/_docs-main.scss
+++ b/themes/default/theme/src/scss/docs/_docs-main.scss
@@ -369,7 +369,6 @@ body.section-registry {
         .docs-table-of-contents {
             max-width: 700px;
             min-width: 0;
-            height: 90vh;
             display: flex;
 
             .table-of-contents {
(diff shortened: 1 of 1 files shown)


cnunciato:
Thanks for the fix! We'll take a look this morning.

PR #3162, opened 2023-07-17
Add image aws golang engin

SaraDPH:
## Checklist:


cnunciato:
> The code in this workshop page looked different, hopefully it still works because I just followed the video instructions.

Yep, this is fine. There are a few variations of these Resources pages, so the fields at the top of these files vary a bit, but this particular field (`meta_image`) is common to all of them, so should work across the board.

cnunciato (approved):
Looks good! 

image

PR #3166, opened 2023-07-18
Restore Stacks blog post

meagancojocar:
## Checklist:


MitchellGerdisch (commented):
I'll add some ramblings here, but I think building up to the punchline of customers being able to undelete state files would flow better. Something like:

As Pulumi users know, `pulumi destroy` is used to delete a stack's resources. And although this action results in an empty state stack file, there is still a state file.
And as Pulumi users also know, if they want to completely delete the state file they can use `pulumi stack rm`. 
One can also run `pulumi stack rm --force` which will remove the state file even though it currently has resources in it. 
But, doing so means you've lost the ability to manage those resources going forward.
And there are cases where this is a valid choice.
But we here at Pulumi sometimes receive support tickets when someone did that `pulumi stack rm --force` mistakenly.
And one of the benefits for our customers using Pulumi Cloud is that we keep back ups of the state files and so when asked, we would resurrect the missing state file and our customers can continue as if nothing happened.

Well, the good news is that our customers can now resurrect deleted state files themselves and no longer have to open a support ticket to have this done for them.

Watch it in action!  .....

themes/default/content/blog/restore-stacks/index.md
```diff
@@ -0,0 +1,53 @@
...
+---
+
+Starting today, you can restore previously deleted stacks in the Pulumi Cloud console. We've had a number of requests from customers to recover stacks that have been deleted. The common use cases to need to recover a deleted stack are either because the stack was accidentally deleted or the stack was intentionally deleted but, later on, they want to restore and preserve the activity history on the stack and just remove its resources.
+
+<!--more-->
+
+Customers typically use `pulumi destory` to delete the resources in a stack but leave the stack itself. But in some cases customers might use `pulumi stack rm --force` which forces the deletion of the stack but leaves behind resources which are no longer managed by the stack. In the case of a force deleted stack, restoring it allows customers to be able to either maintain the stack or properly destroy its contents. Using the new Pulumi Cloud Restore Stacks feature, the last 10 deleted stacks in an organization can be restored by an organization admin.
+
+Watch it in action!
+![Walking through the Restore Stacks experience](restore_stacks.gif)
```
cnunciato, line 49:
```suggestion
![Walking through the Restore Stacks experience](https://www.pulumi.com/uploads/restore-stacks.gif)
```

scottslowe (approved):
Approved (assuming you all will get the animated GIF added)

themes/default/content/blog/restore-stacks/index.md
```diff
@@ -0,0 +1,56 @@
...
+Starting today, you can restore previously deleted stacks in the Pulumi Cloud console. We've had a number of requests from customers to recover stacks, either because the stack was accidentally deleted or the stack was intentionally deleted but, later on, they want to restore and preserve the activity history on the stack and just remove its resources.
+
+<!--more-->
+
+Customers often use `pulumi destory` to delete the resources in a stack but leave the state file. If they want to delete the state file they use `pulumi stack rm`. Customers might also use `pulumi stack rm --force`, which forces the deletion of the state file but leaves behind its resources and in doing so you lose the ability to manage those resources going forward. There are cases where a force delete is a valid choice, but at Pulumi we sometimes receive customer support tickets when someone runs `pulumi stack rm --force` accidentally. One of the benefits of using Pulumi Cloud is that it maintains state file versions enabling Pulumi to be able to restore deleted stacks. Starting today we have built this recovery process into the product, allowing customers to self-serve restore their stacks from a deleted state.
```
MitchellGerdisch, line 46: destory should be destroy

MitchellGerdisch, line 46: Feel free to ignore these:

- should `allowing customers to self-serve restore their stacks ...` be `... to self-serve restoring their stacks ...` or `... to self-serve restoration of their stacks ...`

- and maybe say `... enabling Pulumi to be able to restore deleted state files ...` instead of `... deleted stacks ...` 
Similarly with the last line around `... restore their stacks from a deleted state ...` should be `... restore deleted state files ...` The reason being I worry that someone might think we offer some magic where we'll redeploy resources from a restored state file if we say we are restoring the stack.

Issue #3208, opened 2023-07-31
EKS docs show incorrect type for IAM role


dev-yyh:
This page: [https://www.pulumi.com/docs/clouds/aws/guides/eks/https://www.pulumi.com/docs/clouds/aws/guides/eks/#configuring-your-eks-clusters-worker-nodes-and-node-groups](https://www.pulumi.com/docs/clouds/aws/guides/eks/#configuring-your-eks-clusters-worker-nodes-and-node-groups)

Using Python examples in a document results in errors.
In the Python example, the value of Sid should be “” instead of None. Using the example in the docs leads to an error.

PR #3209, opened 2023-07-31
Fix EKS docs

dev-yyh:
## Description
Fix: #3208
## Checklist:


cnunciato (commented):
Thanks for the fix @dev-yyh! We took a quick look at this this morning and our suggestion would be to remove this property, as it's optional and the value `""` isn't significant. If you're okay with that, feel free to apply and we'll get this merged for ya.

dev-yyh:
Thank you for your suggestion, @cnunciato.
I changed it as you suggested.

cnunciato (approved):

PR #3214, opened 2023-08-01
Remove bad link

cnunciato:
No longer works, so unlinking.


sean1588 (approved):
yeah, i couldn't find a link to replace it either

:rocket:

ericrudder:
i think this is fine as is.

i also *think *this is the account you're looking for, in case the other
contribution pops up:  https://github.com/aohoyd

PR #3221, opened 2023-08-03
Fix GCP getting-started tutorial

cnunciato:
This change fixes a number of issues with the GCP getting-started tutorial. There are a lot of little changes here, so I apologize for the noise. The primary changes are these:

* Adds required `location` properties to bucket declarations. These exist in the `gcp-*` templates, but were missing from the code samples. Without these, initial deployments would fail.

* Fixes a Go variable name in the modify-program step. Without this, the Go program would not compile.

* Fixes a warning about using a deprecated syntax for the `FileAsset` in YAML.

* Adds the `BucketIAMBinding` declaration earlier on in the program, removing the need for an additional `BucketIAMMember`. This was the main issue, and it broke in a few ways: 

    * First, in Go, the `bucketMember` variable was unused, leading to a compile error.
    * Second, for all languages, the update in the deploy-changes step would fail on creating the `BucketIAMMember` and on updating the `BucketObject`: 
      ```
      gcp:storage:BucketIAMMember (bucketIAMMember):
        error: expected non-nil error with nil state during Create of urn:pulumi:dev::quickstart::gcp:storage/bucketIAMMember:BucketIAMMember::bucketIAMMember

      gcp:storage:BucketObject (index.html):
        error: 1 error occurred:
            * Error when reading or editing Storage Bucket Object "index.html-0514f1d": googleapi: Error 403: christian@pulumi.com does not have storage.objects.get access to the Google Cloud Storage object. Permission 'storage.objects.get' denied on resource (or it may not exist)., forbidden
      ```

    * Third, the destroy step would also fail (also for all languages) on the bucket being non-empty, leaving the program wedged and requiring that the file be deleted out of band before the destroy could be completed:
        ```
        gcp:storage:Bucket (my-bucket):
          error: deleting urn:pulumi:dev::quickstart::gcp:storage/bucket:Bucket::my-bucket: 1 error occurred:
              * Error trying to delete bucket my-bucket-544ee90 containing objects without `force_destroy` set to true
        ```

I'll admit I don't fully understand why the provider wasn't able to reconcile all of these changes in one step, but I suspect it has something to do with the combination of changes (to overall bucket permissions, a bucket object property that  triggers a replacement, the addition of a `BucketIAMMember` resource, etc.) and some sort of competition between them -- probably involving the replacement. I'll investigate a bit more for my own understanding (and file a bug on the provider if necessary), but regardless, removing the `BucketIAMMember` resource (which is not actually required here) and applying the bucket permissions earlier on in the workflow simplifies things overall and works reliably now across all languages.

As part of this, I've also [added some automation for this tutorial to the getting-started repo](https://github.com/pulumi/getting-started/pull/3) so we can test this code daily along with the code for the AWS version. The code changes here were all sourced from that repo.

Lastly I went ahead and updated all of the output snippets as well (as they'd drifted a bit) and tightened up some of the language as I walked through these changes myself.

Fixes https://github.com/pulumi/docs/issues/9595.

### DIFF
--- a/themes/default/content/docs/clouds/gcp/get-started/deploy-changes.md
+++ b/themes/default/content/docs/clouds/gcp/get-started/deploy-changes.md
@@ -20,18 +20,18 @@ Now let's deploy your changes.
 $ pulumi up
 ```
 
-Pulumi will run the `preview` step of the update, which computes the minimally disruptive change to achieve the desired state described by the program.
+Pulumi will run the `preview` step of the update, which computes the minimally disruptive change to achieve the desired state described by the program:
 
 ```
-Previewing update (dev):
+Previewing update (dev)
 
-     Type                         Name                   Plan
-     pulumi:pulumi:Stack          quickstart-dev
- +   └─ gcp:storage:BucketObject  index.html             create
-
+     Type                             Name               Plan
+     pulumi:pulumi:Stack              quickstart-dev
+ +   ├─ gcp:storage:BucketIAMBinding  my-bucket-binding  create
+ +   └─ gcp:storage:BucketObject      index.html         create
 
 Resources:
-    + 1 to create
+    + 2 to create
     2 unchanged
 
 Do you want to perform this update?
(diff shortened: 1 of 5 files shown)


scottslowe (approved):
LGTM

PR #3235, opened 2023-08-07
Updated YAML shortcode value

toriancrane:
## Description

Updated Pulumi choosable YAML reference from Node.js to Pulumi YAML.


cnunciato (approved):
🎉

PR #3245, opened 2023-08-09
Updated resource name change info

toriancrane:
## Description

To resolve [this issue](https://github.com/pulumi/docs/issues/9591), I have added additional text to the Resource concept page to indicate/highlight that changing a resource's name will create a new resource and delete the old one.

### DIFF
--- a/themes/default/content/docs/concepts/resources/names.md
+++ b/themes/default/content/docs/concepts/resources/names.md
@@ -183,6 +183,18 @@ resources:
 
 {{< /chooser >}}
 
+{{% notes "warning" %}}
+
+Be careful when you change a resource’s name because changing the name of a resource will create a new resource and delete the old/original resource.
+
+{{% /notes %}}
+
+{{% notes "warning" %}}
+
+If you’d like to rename a resource without destroying the old one, refer to the [aliases](/docs/concepts/options/aliases/) resource option.
+
+{{% /notes %}}
+
 If the `name` property is not available on a resource, consult the [Registry](/registry/) for the specific resource you are creating. Some resources use a different property to override auto-naming. For instance, the `aws.s3.Bucket` type uses the property `bucket` instead of name. Other resources, such as `aws.kms.Key`, do not have physical names and use other auto-generated IDs to uniquely identify them.
 
 Overriding auto-naming makes your project susceptible to naming collisions. As a result, for resources that may need to be replaced, you should specify `deleteBeforeReplace: true` in the resource’s options. This option ensures that old resources are deleted before new ones are created, which will prevent those collisions.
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/concepts/resources/names.md
```diff
@@ -323,7 +323,7 @@ error: Duplicate resource URN 'urn:pulumi:production::acmecorp-website::custom:r
 
 Any change to the URN of a resource causes the old and new resources to be treated as unrelated—the new one will be created (since it was not in the prior state) and the old one will be deleted (since it is not in the new desired state). This behavior happens when you change the `name` used to construct the resource or the structure of a resource’s parent hierarchy.
 
-Both of these operations will lead to a different URN, and thus require the `create` and `delete` operations instead of an `update` or `replace` operation that you would use for an existing resource. In other words, be careful when you change a resource’s name.
+Both of these operations will lead to a different URN, and thus require the `create` and `delete` operations instead of an `update` or `replace` operation that you would use for an existing resource. In other words, be careful when you change a resource’s name because **changing the name of a resource will create a brand new resource and delete the old/original resource**.
```
cnunciato, line 326: At least as of today, we rarely use bold for emphasis, so I'd probably recommend sticking to that policy for now, just for consistency:

```suggestion
Both of these operations will lead to a different URN, and thus require the `create` and `delete` operations instead of an `update` or `replace` operation that you would use for an existing resource. In other words, be careful when you change a resource’s name because changing the name of a resource will create a new resource and delete the old/original resource.
```

That said, this does seem like something worth calling out -- perhaps even as an alert, since I believe the user who filed it was surprised by this behavior. We have a `{{< notes >}}` shortcode that we use for callouts that takes an optonal `type` param (`info` or `warning`) that might be a good fit for this. What do you think? Here's an example:

https://www.pulumi.com/docs/concepts/stack/#generate-an-update-plan

image

It may also be worth mentioning that by default, we create new resources (and update existing resources' relationships to them) before deleting old ones. (Deleting resources is a bit less scary when you know this.)

cnunciato (commented):
Totally works, only one question for you to consider. 

Well, maybe two: Do you think, given the page is about resource names, and we're trying to explain what happens when you change them, that the best place for this info is under Resource URNs (which are a bit of an implementation detail)? What you have here is totally fine, mainly just wondering what you think, as a fairly new user yourself.

themes/default/content/docs/concepts/resources/names.md
scottslowe: I really like the idea of putting that extra warning in a "warning"-type note callout.

toriancrane:
> Totally works, only one question for you to consider.
> 
> Well, maybe two: Do you think, given the page is about resource names, and we're trying to explain what happens when you change them, that the best place for this info is under Resource URNs (which are a bit of an implementation detail)? What you have here is totally fine, mainly just wondering what you think, as a fairly new user yourself.

It's funny that you mention this because I had the thought as I was reading the page that the info was quite far down on it. I do think it makes sense to have it higher up as a callout (will update)

toriancrane:
@cnunciato @scottslowe updated based on your comments. I decided to move it in the high level Resource Names section rather than the Logical Names section because I wondered if readers would assume that the delete/create change would occur only if a logical name was changed. Unless my understanding is incorrect, this would also happen if a Physical Name is specified? (I wonder if that's also why this info was placed in the URN section to begin with.)

scottslowe (approved):
I like the new approach better.

cnunciato (approved):
Looks good! 🚀

PR #3252, opened 2023-08-10
Sarah/776

shughes26:
## Description
https://github.com/pulumi/marketing/issues/776

[figma](https://www.figma.com/file/0JzU3FNV4CAHSEVUaqiA9c/paid-ad-landing?type=design&node-id=32-283&mode=design&t=JpiDV7mZg9xO7sjk-0) for reference

new page created at `/signup`
"small" size option added to color icon partial (defaults to existing size, now named "large")

## Checklist:


anita-trimbur:
thanks for all your work on this sarah, it looks great!! just a few things from me:

1. text styles on the three bullets of copy can be in inter 20px instead of gilroy 24px - that'll give us enough room to repeat that "try pulumi" primary button so it's not only up in the header
2. let's use sentence case on the title so that "get started" doesn't read like a proper noun
3. some layering issues on small screen sizes with the footer
Screen Shot 2023-08-11 at 11 15 17 AM
4. not blocking, but if it's easy to replace the hamburger menu in the header on small screen sizes with the "try pulumi" primary button, that'd be sick

themes/default/content/signup/_index.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Get started with Pulumi
+meta_desc: Pulumi is an infrastructure as code tool that enables engineers to write infrastructure code in a language they already know and love. Get started for free.
+meta_image:
```
zchase, line 4: If we don't end getting a meta image before merging, I'd just remove this field so it uses the default image.

zchase (approved):

themes/default/content/signup/_index.md
anita-trimbur: do we just need the meta image to say "get started with pulumi"?

shughes26: We're going to go with the default image, I'll just delete the field

PR #3279, opened 2023-08-22
Fix a typo in the typescript dynamic provider example

MitchellGerdisch:
## Checklist:

- [x ] I have reviewed the [style guide](https://github.com/pulumi/pulumi-hugo/blob/master/STYLE-GUIDE.md).

### DIFF
--- a/themes/default/content/docs/concepts/resources/dynamic-providers.md
+++ b/themes/default/content/docs/concepts/resources/dynamic-providers.md
@@ -472,7 +472,7 @@ export class MyResource extends pulumi.dynamic.Resource {
     public readonly myNumberOutput!: pulumi.Output<number>;
 
     constructor(name: string, props: MyResourceInputs, opts?: pulumi.CustomResourceOptions) {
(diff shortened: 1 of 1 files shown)


SpainTrain:
I think it needs to be instantiated like in  https://github.com/pulumi/pulumi-hugo/pull/3278/files (this is consistent with the Python examples too)

cnunciato (approved):
Thanks @MitchellGerdisch!

cnunciato:
May have jumped the gun here, as I just noticed @SpainTrain's comment. I'll verify this locally and if necessary, revert.

cnunciato:
Yeah, looks like this does need to be an instance of `pulumi.dynamc.ResourceProvider`. We have an example of this in the examples repo: https://github.com/pulumi/examples/blob/955617680f6224019b86f14c76f5a1f72ba1a917/aws-ts-eks-distro/eksdistro/index.ts#L105 I'll revert.

cnunciato:
Oh no! I see you submitted the correction as well! I should've check for that first. 🤦 I'll see if I can get this fixed up.

PR #3290, opened 2023-08-23
Add meta image to Migrate Hub page

SaraDPH:
## Checklist:


SaraDPH:
@cnunciato why can't I see the image in the preview? :(

cnunciato:
@SaraDPH The image path (in `meta_image`) and the actual location of the file don't match. I believe we're currently putting these page-specific images in `static` -- right?

SaraDPH:
> @SaraDPH The image path (in `meta_image`) and the actual location of the file don't match. I believe we're currently putting these page-specific images in `static` -- right?
> 
> It looks like we have another image in this folder also -- we should remove that one if we're no longer using it. I'll pick this up for ya soon as I'm in.

There is another image?!

cnunciato:
Looks good I think! 

image

cnunciato (approved):

PR #3295, opened 2023-08-23
WIP: Move Learn into Docs (as Tutorials)

cnunciato:


themes/default/layouts/learn/module.html
toriancrane, line 1: Not at all a time-sensitive request, but when clicking the Creating Pulumi Programs link on the left menu:

image

is there a way to have the Nav title reflect the name of the topic rather than "Tutorials"? For example, in the below image, it shows "Tutorials" in the left hand menu when I think it would make more sense to say "Creating Pulumi Programs".

image

cnunciato: Absolutely! We can do pretty much anything we want with this.

cnunciato:
Closing this for now just to clear out my open PRs, but will leave the branch for reference for when we decide to come back to this.

PR #3299, opened 2023-08-24
Remove tweets on b version of homepage

zchase:
## Description

This PR removes the tweets from the "B" version of the home page as they were overlapping the get started CTA. They were overlapping because some styles were being stripped out during the production build, therefore I could repro locally.

Screenshot 2023-08-23 at 6 17 31 PM

## Checklist:

### DIFF
--- a/themes/default/layouts/page/home-b.html
+++ b/themes/default/layouts/page/home-b.html
@@ -351,78 +351,15 @@
     </section>
 
     <!-- Get Started -->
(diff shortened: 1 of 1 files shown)


sean1588 (approved):

sean1588:
hmm i wonder if it is purgecss that is yanking out styles it doesn't think it needs

zchase:
@sean1588 IIRC I've hit this before around using size qualifiers like `lg:` which makes sense purgecss is pulling it out.

PR #3337, opened 2023-09-05
Azure campaign images (to be used on external ads)

SaraDPH:
## Description
These images won't be live on the website. Since they don't have storage, I need them to exist in pulumi.com, so they can be pulled to appear on the ads.

## Checklist:


shughes26:
I think it would be good if we could put these (and future images for the same use) into a new subdirectory under the `static/images` directory just to help with organization

jkodroff (approved):

PR #3343, opened 2023-09-07
Improve the first YAML example

lukehoban:
The example has a few issues:
1. It doesn't work - since AWS made breaking changes to public buckets earlier this year
2. It doesn't show off `config` or `options`
3. It uses the now-deprecated capitalized `Fn::` syntax

These changes address all three.


EronWright (approved):
Much improved, thanks!

Issue #3091, opened 2023-06-20
Hide "On this Page" when there isn't anything to show below it


cnunciato:
On pages without H2 headings, we show an "On this Page" widget, but it's empty:

![image](https://github.com/pulumi/pulumi-hugo/assets/274700/2041f05a-dc0e-46ea-aa2a-7e41d2794f85)

We should either always show at least something in this section or treat it somehow differently when there aren't any links below it.

PR #3349, opened 2023-09-08
hide "on this page" by default

sean1588:
## Description

fixes: #3091 

We wrote the logic [to explicitly show the table of contents ("on this page") ](https://github.com/pulumi/pulumi-hugo/pull/3349/files#diff-072f0832bb3eebd378aa49a6c42cfbe4631fceb90a6752feef6013868f957115R115)when elements exist for it to display, so hide it by default so we do not have it displayed when no elements exist under it. I was trying to just hide this by default directly in the CSS, because looks like this is being controlled in multiple places in the CSS and I coudn't quite reason about the precedence/composition structure here in order for it to work right in all instances, so I resorted to just doing it in jquery, since we are also explicitily calling when to show it.

## Checklist:


cnunciato (approved):
:shipit:

Issue #3354, opened 2023-09-10
Getting started code out of sync with pulumi/templates


EvanBoyle:
The template repo was updated for AWS v6, kubernetes v4 and, azure native v2. However, the code snippets in the getting started guides did not get updated. 

This example is still using v3 kubernetes and the instructions that tell you to copy paste the code after you've provisioned a v4 template will fail with a compilation error as reported here 

https://www.pulumi.com/docs/clouds/azure/get-started/review-project/
https://twitter.com/timburks/status/1700598180084433388

Affected paths:

- [x] AWS Go
- [x] Azure Go
- [x] Kubernetes Go

PR #3359, opened 2023-09-11
Fix AWS Go module reference in the Getting Started guide

cnunciato:
Part of #3354.


scripts/lint/lint-markdown.js
```diff
@@ -1,3 +1,4 @@
+
```
cnunciato, line 1: Shouldn't be in the PR, I'll remove.

sean1588 (approved):

mnlumi (approved):

PR #3361, opened 2023-09-11
blog tweaks

usrbinkat:
## Description

last tweaks to blog

## Checklist:

### DIFF
--- a/themes/default/content/blog/mlops-huggingface-llm-aws-sagemaker-python/index.md
+++ b/themes/default/content/blog/mlops-huggingface-llm-aws-sagemaker-python/index.md
@@ -164,6 +164,6 @@ To recap, in a few commands, we created a new Pulumi Python project from a ready
 
 AI and ML is rapidly becoming a necessity with every new day. It may appear daunting or out of reach at first glance, but with the power of IaC written as Pulumi Python programs, getting started has never been easier.
 
-If you followed along then tell us how it worked out for you! We would love to know what you are looking forward to, or if you have ideas for future installments of the Pulumi Python #MLOpsChallenge series!
+If you followed along then tell us how it worked out for you! We would love to know what you are looking forward to, or if you have ideas for future installments of the Pulumi Python #PulumiMLOps series!
 
-Join us on [Twitter](https://twitter.com/pulumicorp), and on the [Pulumi Community Slack](https://slack.pulumi.com) to decide what #MLOpsChallenge we take on next!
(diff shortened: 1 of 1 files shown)


elistaykova:
@usrbinkat, I'll leave here comments for the blog (current version I am reading is: http://pulumi-hugo-origin-pr-3361-c13afd83.s3-website.us-west-2.amazonaws.com/blog/mlops-huggingface-llm-aws-sagemaker-python/ )

1) Intro: Good job on framing the problem more concretely. I do wish we are explaining the concept and speaking of the outcome more clearly in the intro. 
ex: you are saying - running LLM on sagemaker is a popular deployment option. Why? Why do people do that, what does it help them achieve? what's the outcome of that job?
ex: "(IaC) offers a reliable and easy to duplicate deployment practice." generic statement, that doesn't give much value. The next sentence hints at you can use Python to write app and iac code.
is this the only reason why engineers should switch to our way of building? Not sure I would based on how it's written. Does not make technically convincing argument.

2) second para "Pulumi makes AI/ML Easy?"
are we asking? it looks like a doubtful statement. Write it with conviction
maybe this is a silly comment, but why do we link to a generic Pulumi Templates blog, and don't link to the template in question?

3) Instructions -  I may be off mark here, do engineers find a list of code instructions helpful? i don't think this is the interesting part of the blog - this makes the blog read like a tutorial or a code recipe. The login, the project setup, the destroy command, and so on... - should this be in the readme? you can still link to all these code instructions, and in the blog focus on the teaser, the WOW factor to get people excited to build

4) part 3 where you show the sagemaker endpoints is good - i see what resources are provisioned and how, how long it took. Though you are warning in the copy it can take up to 20min? is that something else? file shows sub 6min.

5) Test endpoint - i think the section that's supposed to be the WOW, reads like a formality
OK, we are helping people create an endpont and test the endpoint. What happened? Was that easy, hard, magical, expected, no one else can do that? 

6) conclusion - i wouldn't anchor on Pulumi Python and MLOPS as The Thing. I don't think we should pigeonhole ourselves into a category. Why not keep it broad-ish? And more generic like, AI Infrastructure? 
FWIW:
- MLOps gets 12.6 million google searches
- AI Infrastructure gets 664 million searches
go bigger (even if MLOps is the pedantically correct term, it's for a more niche audience)

the goal of our python/AI series is to show folks that we are accelerant for engineers spinning up AI efforts (whether you are building AI apps, LLMs, using AI internally to ship, whatever), we will help you accelerate these efforts with our unique approach, that's the premise

scottslowe (approved):
Approved

PR #3376, opened 2023-09-14
Update Puluminaries page

scottslowe:
## Description

Add Jowanza Joseph.
Add Tyler Mulligan.

## Checklist:


cnunciato (changes requested):
Can we reduce the size of the images? These are generally 200x200 and around 100k. (We need to keep an eye on file sizes not only for site performance, but also because we've been bumping up against the 500MB Go module limit, which wedges the pipeline when we exceed it.)

scottslowe:
@cnunciato Done. I'll open a separate PR to size down images not added via this PR.

interurban (approved):
thx for fixing the image sizes !

aaronkao (approved):

cnunciato (approved):

PR #3382, opened 2023-09-18
Update install-go.html to align with Go terms

Frassle:
We were calling 1.20 to 1.21 a minor release. Go calls that a major release.

### DIFF
--- a/themes/default/layouts/shortcodes/install-go.html
+++ b/themes/default/layouts/shortcodes/install-go.html
@@ -10,7 +10,7 @@
     </div>
     <div class="content">
         <p>
-            Pulumi requires a <a href="https://go.dev/doc/devel/release#policy">supported version of Go</a>&mdash; this typically refers to the two most recent major releases. If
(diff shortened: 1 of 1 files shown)


themes/default/layouts/shortcodes/install-go.html
```diff
@@ -10,7 +10,7 @@
     </div>
     <div class="content">
         <p>
-            Pulumi requires a <a href="https://go.dev/doc/devel/release#policy">supported version of Go</a>&mdash; this typically refers to the two most recent minor releases. If
+            Pulumi requires a <a href="https://go.dev/doc/devel/release#policy">supported version of Go</a>&mdash; this typically refers to the two most recent major releases. If
```
thomas11, line 13: I'm afraid to readers not super familiar with Go’s non-standard terminology, this will sound like 1.x, 2.x. Thanks Go team… Can we clarify this somehow?

```suggestion
            Pulumi requires a <a href="https://go.dev/doc/devel/release#policy">supported version of Go</a>&mdash; this typically refers to the two most recent major releases. Note that Go calls 1.20, 1.21, etc. major releases, unlike semantic versioning. If
```

thomas11 (approved):

PR #3404, opened 2023-09-26
adding ai/answers to site nav

interurban:
## Description
Adding pulumi AI Answers to Nav

### DIFF
--- a/themes/default/layouts/partials/header.html
+++ b/themes/default/layouts/partials/header.html
@@ -316,10 +316,10 @@
                                         </li>
                                         <li>
                                             <div class="list-title">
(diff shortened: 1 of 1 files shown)


interurban:
Hey @cnunciato can you double check and approve when ready?

cnunciato (approved):
Looks good, but you'll want to make the same change to the mobile version as well I think. I can't seem to issue a suggestion for that on this review, but it's right here: 

https://github.com/pulumi/pulumi-hugo/blob/b21f99c1bb2a4e476226514ac477243d41b02868/themes/default/layouts/partials/header.html#L451

PR #3415, opened 2023-09-28
fix text wrapping issue in registry table

sean1588:
fixes: https://github.com/pulumi/docs/issues/9920

### DIFF
--- a/themes/default/theme/src/scss/docs/_docs-main.scss
+++ b/themes/default/theme/src/scss/docs/_docs-main.scss
@@ -90,6 +90,7 @@ body.section-registry {
             padding: 0 0 12px 0;
             margin: 0;
             line-height: 1.6;
+            overflow-wrap: break-word;
         }
 
         table {


themes/default/theme/src/scss/docs/_docs-main.scss
```diff
@@ -90,6 +90,7 @@ body.section-registry {
             padding: 0 0 12px 0;
             margin: 0;
             line-height: 1.6;
+            overflow-wrap: break-word;
```
cnunciato, line 93: Should we scope this to tables? Only concerned we might end up with broken words where we aren't currently expecting them. (Clicking around so far, though, this looks okay.)

sean1588: yeah probably not a bad idea

cnunciato (approved):

sean1588:
I moved this PR to registry, since it manages its own version of theme now.

PR #3423, opened 2023-09-29
fix go code for ApplyT examples

### LINKED ISSUE #2739: confirm accuracy of inputs and outputs go example in concepts docs
_this feedback was submitted by way of docs feedback_

https://www.pulumi.com/docs/intro/concepts/inputs-outputs/
the go example for pulumi.All() was reported as being incorrect
folks would also like to know how to get back errors

desteves:
fixes #2739

### DIFF
--- a/themes/default/content/docs/concepts/inputs-outputs/_index.md
+++ b/themes/default/content/docs/concepts/inputs-outputs/_index.md
@@ -532,8 +532,9 @@ connectionString := pulumi.All(sqlServer.Name, database.Name).ApplyT(
 )
 ```
 
+{{% notes %}}
 **A note on error handling** The function `ApplyT` spawns a Goroutine to await the availability of the implicated dependencies. This function accepts a `T` or `(T, error)` signature; the latter accomodates for error handling. Alternatively, one may use the `ApplyTWithContext` function in which the provided context can be used to reject the output as canceled. Error handling may also be achieved using an `error` `chan`.
-
+{{% /notes %}}
 {{% /choosable %}}
 {{% choosable language csharp %}}
 
--- a/themes/default/content/docs/concepts/inputs-outputs/_index.md
+++ b/themes/default/content/docs/concepts/inputs-outputs/_index.md
@@ -522,14 +522,17 @@ connection_string = Output.all(server=sql_server.name, db=database.name) \
 {{% choosable language go %}}
 
 ```go
+
 connectionString := pulumi.All(sqlServer.Name, database.Name).ApplyT(
-    func (args []interface{}) (string, error) {
-        server := args[0].(string)
-        db := args[1].(string)
-        return fmt.Sprintf("Server=tcp:%s.database.windows.net;initial catalog=%s...", server, db), nil
-    },
+    func (args []interface{}) pulumi.StringOutput  {
+        server := args[0]
+        db := args[1]
+        return pulumi.Sprintf("Server=tcp:%s.database.windows.net;initial catalog=%s...", server, db)
+    }
 )
 ```
+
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/concepts/inputs-outputs/_index.md
```diff
@@ -522,15 +522,18 @@ connection_string = Output.all(server=sql_server.name, db=database.name) \
...
+    func (args []interface{}) pulumi.StringOutput  {
+        server := args[0]
+        db := args[1]
+        return pulumi.Sprintf("Server=tcp:%s.database.windows.net;initial catalog=%s...", server, db)
+    }
 )
 ```
 
+**A note on error handling** The function `ApplyT` spawns a Goroutine to await the availability of the implicated dependencies. This function accepts a `T` or `(T, error)` signature; the latter accomodates for error handling. Alternatively, one may use the `ApplyTWithContext` function in which the provided context can be used to reject the output as canceled. Error handling may also be achieved using an `error` `chan`.
+
```
cnunciato, line 536: If you were thinking of this as a callout, we have these which you might want to use instead:

```suggestion
{{% notes %}}
**A note on error handling** The function `ApplyT` spawns a Goroutine to await the availability of the implicated dependencies. This function accepts a `T` or `(T, error)` signature; the latter accomodates for error handling. Alternatively, one may use the `ApplyTWithContext` function in which the provided context can be used to reject the output as canceled. Error handling may also be achieved using an `error` `chan`.
{{% /notes %}}
```

They look like this:

image

themes/default/content/docs/concepts/inputs-outputs/_index.md
```diff
@@ -522,15 +522,18 @@ connection_string = Output.all(server=sql_server.name, db=database.name) \
...
-    func (args []interface{}) (string, error) {
-        server := args[0].(string)
-        db := args[1].(string)
-        return fmt.Sprintf("Server=tcp:%s.database.windows.net;initial catalog=%s...", server, db), nil
-    },
+    func (args []interface{}) pulumi.StringOutput  {
+        server := args[0]
+        db := args[1]
+        return pulumi.Sprintf("Server=tcp:%s.database.windows.net;initial catalog=%s...", server, db)
+    }
```
cnunciato, line 531: My compiler complains about a missing comma here:

image

```
Diagnostics:
  pulumi:pulumi:Stack (inputs-outputs-go-dev):
    # inputs-outputs-go
    ./main.go:24:5: syntax error: unexpected newline in argument list; possibly missing comma or )
```

When I add a trailing comma, all is well:

```suggestion
    },
```

cnunciato (approved):
Approving to keep unblocked, but we should doublecheck that missing-comma issue and apply the suggestion if that repros for you.

themes/default/content/docs/concepts/inputs-outputs/_index.md
desteves: Ooops...I see the same actually but when the '}' immediately precedes the ')', it seems okay. This is how I have it in my test env:

```go

		one := pulumi.String("123").ToStringOutput()
		another := pulumi.String("abc").ToStringOutput()
		connectionString_CHECK := pulumi.All(one, another).ApplyT(
			func(args []interface{}) pulumi.StringOutput {
				server := args[0].(string)
				db := args[1].(string)
				return pulumi.Sprintf("Server=tcp:%s.database.windows.net;initial catalog=%s...", server, db)
			}) // <--- this works
		ctx.Export("connectionString_CHECK", connectionString_CHECK)
```

So we can represent it with the comma as you've shown or as above. Any preference?

PR #3427, opened 2023-09-30
Fix example code bug for configuring policies

### LINKED ISSUE #3426: Docs on configuring policies has a code bug
https://www.pulumi.com/docs/using-pulumi/crossguard/configuration/

The TypeScript examples refer to `validateTypedResource` but that should be `validateResourceOfType`.

justinvp:
The TypeScript examples for configuring policies is using `validateTypedResource`, but that should be `validateResourceOfType`.

Fixes #3426


ericrudder (approved):
Thanks for fixing

cnunciato (approved):
Thanks Justin!

PR #3447, opened 2023-10-07
Move some GIFs into the uploads bucket

cnunciato:
Wins back a little over 100MB, and drops the max file size from 4MB to 3MB.


scripts/lint/lint-markdown.js
```diff
@@ -284,6 +284,8 @@ const opts = {
         MD028: false,
         // Allow indentation in unordered lists.
         MD007: false,
+        // Allow bare URLs.
+        MD034: false,
```
cnunciato, line 288: This was necessary to make some `figure` shortcodes work. Seems all right, but let me know if this is a particularly bad thing to do. (And if so, how to escape it on a case-by-case basis, as I couldn't seem to figure this out.)

scottslowe (approved):
General question: should we be using relative URLs (`/uploads/blog...`) instead of absolute URLs (`https://www.pulumi.com/uploads/blog...`)? No need to block, just a suggestion.

cnunciato:
Good question! Technically we could, but because the `/uploads` path doesn't exist in this repo (it's mapped as another origin in the www.pulumi.com CloudFront CDN), the local development server and PR previews wouldn't see the content that's been moved to that folder. So we have to reference by its fully qualified URL.

PR #3450, opened 2023-10-09
Fix C# snippet

iamdmitrij:
## Description

Noticed that official C# documentation has a code which doesn't compile: https://www.pulumi.com/docs/languages-sdks/dotnet/#example
Snippet was simply missing a `class` modifier.

- [x] Fix C# documentation snippet

### DIFF
--- a/themes/default/content/docs/languages-sdks/dotnet/_index.md
+++ b/themes/default/content/docs/languages-sdks/dotnet/_index.md
@@ -52,7 +52,10 @@ public class MyStack : Stack
     public MyStack()
     {
         // Create an Azure Resource Group
-        var resourceGroup = new ResourceGroup("resourceGroup");
+        var resourceGroup = new ResourceGroup("resourceGroup", new ResourceGroupArgs
+        {
+            Location = "West Europe",
+        });
 
         // Create an Azure Storage Account
         var storageAccount = new Account("storage", new AccountArgs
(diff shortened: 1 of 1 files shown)


toriancrane (changes requested):
Thank you for your contribution!

After testing the proposed update, I have just one requested change. The deployment still fails with the updated code because the resource group definition is missing the required "location" argument.

```csharp
public class MyStack : Stack
{
    public MyStack()
    {
        // Create an Azure Resource Group
        var resourceGroup = new ResourceGroup("prTestResourceGroup", new()
        {
            Location = "West Europe",
        });
        ...
        ...
        // the rest of the code example
```

If you can include this with your update, I can approve it and get it merged. Thanks again!

iamdmitrij:
> Thank you for your contribution!
> 
> After testing the proposed update, I have just one requested change. The deployment still fails with the updated code because the resource group definition is missing the required "location" argument.
> 
> ```cs
> public class MyStack : Stack
> {
>     public MyStack()
>     {
>         // Create an Azure Resource Group
>         var resourceGroup = new ResourceGroup("resourceGroup", new()
>         {
>             Location = "West Europe",
>         });
>         ...
>         ...
>         // the rest of the code example
> ```
> 
> If you can include this with your update, I can approve it and get it merged. Thanks again!

Good catch, thanks! Fixed it now.

toriancrane (approved):
Looks good to me!

PR #3454, opened 2023-10-10
fixing nav for ESC

interurban:
Fixing the missing icon and description


cleverguy25 (approved):

themes/default/layouts/partials/header.html
```diff
@@ -108,7 +108,7 @@
                                                 <a href="/product/esc/">
                                                     <i class="fas fa-shield-alt fa-fw"></i>
```
cnunciato, line 109: Hm. Isn't this this icon (the shield) the one that's missing?

PR #3477, opened 2023-10-11
Nav: 10_11 fixing missing icon

interurban:
## Description
Fixing the missing icon from nav from tonights launch

### DIFF
--- a/themes/default/layouts/partials/header.html
+++ b/themes/default/layouts/partials/header.html
@@ -133,7 +133,7 @@
                                         <li>
                                             <div class="list-title">
                                                 <a href="/product/internal-developer-platforms/">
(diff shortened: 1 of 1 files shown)


themes/default/layouts/partials/header.html
```diff
@@ -133,7 +133,7 @@
                                         <li>
                                             <div class="list-title">
                                                 <a href="/product/internal-developer-platforms/">
-                                                    <i class="fas fa-box-open fa-fw"></i>
+                                                    <i class="ffas fa-users fa-fw"></i>
```
cnunciato, line 136: I think this needs to be this: 

```suggestion
                                                    <i class="fas fa-users fa-fw"></i>
```

cnunciato (approved):
:shipit:

PR #3497, opened 2023-10-12
Add 'esc login' steps

toriancrane:
## Description

Added content about logging into the esc cli and how users will be prompted. Also marked the `Access token` prereq as optional.

## Checklist:

### DIFF
--- a/themes/default/content/docs/pulumi-cloud/esc/get-started/_index.md
+++ b/themes/default/content/docs/pulumi-cloud/esc/get-started/_index.md
@@ -21,7 +21,8 @@ In this tutorial, we’ll demonstrate the power of Pulumi ESC in managing config
 
 You will need the following tools to complete this tutorial:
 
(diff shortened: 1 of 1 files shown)


scottslowe (approved):
Just one suggestion, but no need to block on it: should there be some text that indicates that a user using `pulumi env` doesn't need to go through the `esc login` steps? Perhaps it's already there and I didn't see it (in which case ignore me).

toriancrane:
> Just one suggestion, but no need to block on it: should there be some text that indicates that a user using `pulumi env` doesn't need to go through the `esc login` steps? Perhaps it's already there and I didn't see it (in which case ignore me).

No worries! There is a note at the top that says:

```
Pulumi ESC is a service of Pulumi Cloud that can be used with or without Pulumi IaC. This means that if you already have the Pulumi IaC CLI installed, you do not need to install the Pulumi ESC CLI, and you may substitute `pulumi env` anywhere you see the `esc env` command in this guide.
```

PR #3515, opened 2023-10-19
Fix Dynamic Resource Providers Docs

meagancojocar:
## Description
I found it really deceptive that these docs say its supported in all languages and show code examples in them all but then the chooser is blank and we dont actually support it in all languages. 

CleanShot 2023-10-18 at 17 11 20

## Checklist:


interurban:
Hey @meagancojocar  sorry for being behind; I'm looking into where our PR review links went for these and will approve shortly

themes/default/content/docs/concepts/resources/dynamic-providers.md
```diff
@@ -50,7 +50,6 @@ const myProvider: pulumi.dynamic.ResourceProvider = {
 }
 ```
 
-{{% /choosable %}}
```
cnunciato, line 53: This needs to be put back in I think. (Should fix your build error.)

PR #3520, opened 2023-10-20
block indexing on blog/tags page

sean1588:
## Description

fixes: https://github.com/pulumi/marketing/issues/818

Disable search indexing on /blog/tags page.

I verified in the page source:

Screenshot from 2023-10-20 14-09-19

## Checklist:

### DIFF
--- a/themes/default/content/blog/tag.md
+++ b/themes/default/content/blog/tag.md
@@ -2,6 +2,7 @@
 title: "Pulumi Blog: All tags"
 layout: tags
 meta_desc: A list of all Pulumi blog tags, sorted alphabetically.
+block_external_search_index: true
 ---
 
 All [Pulumi blog](/blog/) tags, sorted alphabetically.


zchase (approved):

cnunciato (commented):
Are we 100% sure this will not affect the crawling of these pages? We may not want them indexed, but we do need to make sure the bots continue crawling them.

sean1588:
@cnunciato  - good question, not sure how credible this is, but here is a blog post from hubspot I found that claims, `noindex` will only tell the bots not to index and/or revoke the page from the existing search index, but it will not disable bots from crawling the page. 

https://blog.hubspot.com/marketing/how-to-unindex-pages-from-search-engines

Also saw this on googles own documentation, which talks about the use of noindex, but doesn't specifically mention anything with regards disabling crawling. 
https://developers.google.com/search/docs/crawling-indexing/block-indexing

I'm inclined to think the bots will still crawl the page, it just wont index it.

cnunciato:
Hm, I dunno. On the one hand it looks like we might want to use `noindex,follow` for these: https://www.youtube.com/watch?v=ZjRGkc__FwQ&t=57s

On the other, I'm also seeing that having `noindex` at all may be interpreted as "don't even follow the links on this page": https://www.youtube.com/watch?v=9GNg8R-X8LQ&t=3291s

Given we're trying to keep the SEO funnel as wide as we can, I'm just being a bit cautious around things that could potentially narrow it (unless there's such good reason to do so that the benefits outweigh the risks).

sean1588:
@cnunciato, hypothetically speaking, if the bots were not able to crawl the tags page, would that necessarily affect us? Couldn't the bots follow links from elsewhere on the site to reach those pages? But maybe I am misunderstanding how all  this can affect our seo.

cnunciato:
@sean1588 Right now, via the page links, yes -- but I'm assuming that if we're suppressing the tag pages, we'd want to suppress the "page" pages as well. If we did that, then the only posts that'd be crawlable organically would be the ones we linked to from elsewhere (which would be a small subset of all of them).

PR #3534, opened 2023-10-25
Uploaded resized image

toriancrane:
## Description

My team image was too large so I resized it and am now re-uploading it.


pierskarsenbarg (approved):
LGTM

PR #3536, opened 2023-10-25
Add clarity to Pulumi import command

toriancrane:
## Description

Resolves #2068

PR #3538, opened 2023-10-25
Apply suggestion from merged PR

cnunciato:
Suggestion submitted here: https://github.com/pulumi/pulumi-hugo/pull/3536#discussion_r1372253927, but the PR was auto-merged before it could be considered/applied.


toriancrane (approved):
LGTM

PR #3550, opened 2023-10-27
Publishing part 1 of Joe's TNS article

aaronkao:
## Description

Publishing this article on our blog:  https://thenewstack.io/infrastructure-as-code-in-any-programming-language/


themes/default/content/blog/infrastructure-as-code-in-any-programming-language/index.md
```diff
@@ -0,0 +1,87 @@
+---
+title: "Infrastructure as Code in Any Programming Language"
+
```
cnunciato, line 3: We need to make sure to set a `canonical` reference here to point to the original.
```suggestion
canonical_url: https://thenewstack.io/infrastructure-as-code-in-any-programming-language/
```

aaronkao: That would give us no SEO juice right?

cnunciato: I think so, although I'm not sure it's totally deleterious; we may still get some benefit, but the majority will go to the original, yeah. If we're essentially republishing someone else's content, though, the proper thing to do is to indicate this by referring to the original as canonical. Not doing so (again to the best of my understanding) risks being seen as duplicating content can negatively impact our pagerank.

A middle ground might be to link back to the original site somehow -- like in a note at the top of the post. That might be seen as enough -- but chances are the result would be the same. I do think the proper thing to do here is to recognize the original by referencing it as canonical.

Unless of course we can get *them* to identify *us* as canonical. That'd be ideal -- but presumably unlikely.

scottslowe (approved):
Approved pending resolution of Christian's comments

themes/default/content/blog/infrastructure-as-code-in-any-programming-language/index.md
aaronkao: I guess the question is what if we made some minor changes. How many minor changes would we have to make before Google doesn't think its a duplicate.

themes/default/content/blog/infrastructure-as-code-in-any-programming-language/index.md
```diff
@@ -0,0 +1,87 @@
+---
+title: "Infrastructure as Code in Any Programming Language"
+canonical_url: https://thenewstack.io/infrastructure-as-code-in-any-programming-language/
+date: 2023-10-30T22:36:29Z
```
aaronkao, line 4:
```suggestion
date: 2023-10-31T22:36:29Z
```

PR #3576, opened 2023-11-02
Add oct launch to journey svg

anita-trimbur:
fixes pulumi/marketing#835

adds series c + esc + pulumi for platform teams to the journey svg on the about page


ericrudder:
as long as we are here, can we fix overall layout (less whitespace on left, less squeezing of text on the right) and can we just show two launches rather than odd break with ampersand?  thx

anita-trimbur:
@ericrudder let me know if this change addresses your feedback

ericrudder (approved):
thank you

PR #3585, opened 2023-11-03
Moved tutorials content to different path

toriancrane:
## Description

This PR moves the launched tutorials to live under `using-pulumi` to resolve the redirect problem caused by putting them under `docs/tutorials`


cnunciato (approved):
Thanks @toriancrane!

cnunciato:
For posterity, we're moving these because there's a regex rule at the CloudFront Lambda@Edge that matches `docs/tutorials` that was redirecting these to /registry.

PR #3588, opened 2023-11-03
Fixes the dup entries in the install page TOC 

desteves:
Moves and consolidates headers up a level so that these are not nested within the language choosable shortcode. This then prevents the right nav from rendering duplicates. Headers level 1 through level 3 are updated.

- Affected page(s): https://www.pulumi.com/docs/install/
- Fixes 3583

### DIFF
--- a/themes/default/content/docs/install/_index.md
+++ b/themes/default/content/docs/install/_index.md
@@ -1,8 +1,8 @@
 ---
 title_tag: "Download & Install Pulumi"
 meta_desc: This page contains detailed instructions for downloading and installing Pulumi.
-title: "Download & install"
-h1: Download & install Pulumi
+title: "Download & Install"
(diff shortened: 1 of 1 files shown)


sean1588:
@desteves - can you open this PR from a branch created in this repo? Right now it looks like you opened the PR from your fork of the repo and it looks like we have our pipeline configured to not run against forks, so it won't generate the PR preview.

desteves:
> @desteves - can you open this PR from a branch created in this repo? Right now it looks like you opened the PR from your fork of the repo and it looks like we have our pipeline configured to not run against forks, so it won't generate the PR preview.

Sure, let me update this!

desteves:
See https://github.com/pulumi/pulumi-hugo/pull/3625 instead.

PR #3616, opened 2023-11-10
Add initial "What is Serverless" SEO page

scottslowe:
## Description

This PR will add a "What is Serverless" page to the "What Is" section of the web site to garner some serverless SEO authority.

## Checklist:

### DIFF
--- a/themes/default/content/what-is/what-is-serverless.md
+++ b/themes/default/content/what-is/what-is-serverless.md
@@ -1,0 +1,133 @@
+---
+title: What is Serverless?
+meta_desc: |
+    Understand what is serverless, along with some of the benefits of using serverless architectures for modern application development.
+type: what-is
+page_title: What is Serverless?
+customer_logos:
+  title: Leading engineering organizations are building with Pulumi
+  logos:
+    - items:
+      - snowflake
+      - tableau
+      - atlassian
+      - fauna
+      - ware2go
+    - items:
+      - mindbody
+      - sourcegraph
+      - fenergo
+      - skai
+      - lemonade
+    - items:
+      - clearsale
+      - angellist
+      - webflow
+      - supabase
+      - ro
+---
+
+Serverless architectures are popular because they make it easier for software developers to focus on application development without additional operation overhead. With Pulumi, you can deploy and manage your serverless infrastructure, and this article shows you different architectural design considerations for your serverless application.
+
+Before we go into the specifics of serverless architecture design, let's go through the basics of serverless - this is important for the rest of the article.
+
+## What is Serverless Architecture
+
+Serverless architecture is a cloud computing model that simplifies the management of infrastructure. This allows software developers to focus on writing code without worrying about server provisioning, scaling or maintenance. The idea behind it is functions as a service (FaaS), in which the environment dynamically scales in response to demand and code is executed in response to events. In the serverless architecture, developers focus on writing and deploying code in the form of serverless functions, which are executed in response to specific events or triggers.
+
+Different cloud providers have their different serverless offerings. Azure has Azure Functions, AWS has Lambda, and Google Cloud has Google Cloud Functions. These are serverless platforms that allow developers to build event-driven functions that respond to various triggers. In addition to these major players, other cloud providers like Oracle and IBM also offer their own serverless solutions, creating a diverse ecosystem of options for developers to choose from. These serverless platforms enable software developers to build and deploy applications that respond to triggers, such as HTTP requests, database changes, and more, while abstracting away the complexities of infrastructure management.
+
+Using serverless eliminates the responsibilities of operational management, which streamlines the development process. This means you don't have to set up complex security mechanisms or firewalls. In addition, serverless computing abstracts away all operational complexities, saving you from the complexities of operating system patching and guaranteeing a simplified and trouble-free development process.
+
+Key characteristics of serverless architecture include:
+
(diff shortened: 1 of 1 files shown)


aaronkao:
For some reason this isn't rendering to me with the new layout format the new pages started to use

themes/default/content/what-is/what-is-serverless.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Conclusion
+
+Serverless architectures empower developers to build scalable, cost-effective applications without the burden of managing servers. By understanding its components, associated cloud services, and benefits, technical professionals can leverage serverless to streamline development processes and create more agile, responsive applications in the cloud.
+
+Pulumi's code-driven approach to infrastructure as code aligns well with serverless architectures, allowing developers to use the same tools for writing their serverless functions and their infrastructure definitions. To learn more about using Pulumi to deploy serverless applications, check out [this blog post](https://www.pulumi.com/blog/easy-serverless-apps-and-infrastructure-real-events-real-code/).
```
aaronkao, line 110: Seems like these templates are a better destination to start than that blog post which is getting pretty old

https://www.pulumi.com/templates/serverless-application/

themes/default/content/what-is/what-is-serverless.md
```diff
@@ -0,0 +1,112 @@
...
+
+Our [community on Slack](https://slack.pulumi.com/) is always open for discussions, questions, and sharing experiences. Join us there to share your experience with using Pulumi for serverless applications, or to ask questions about how Pulumi can streamline your serverless application deployments. Become part of our growing community of cloud professionals!
```
aaronkao, line 112: How do we update that big purple CTA at the bottom so its tied to the context of the page? Would love for it to say "Get Started with deploying serverless apps today" or soemthing like that

themes/default/content/what-is/what-is-serverless.md
```diff
@@ -0,0 +1,112 @@
...
+
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		w.Write(responseJSON)
+	})
+    ```
+
+1. **Embrace statelessness**: Serverless functions should be stateless. This encourages scalability and fault tolerance since every serverless function call should be able to function on its own, regardless of previous or subsequent calls. The recommended practise for managing data that has to be persistent beyond the span of one function execution is to store it externally in reliable data storage systems like object stores or databases:
+
+	![serverless-resources](./serverless-resources.png)
```
aaronkao, line 86: This file seems to be missing

themes/default/content/what-is/what-is-serverless.md
```diff
@@ -0,0 +1,112 @@
...
+
+* [AWS Lambda](https://aws.amazon.com/lambda/)
+* [Azure Functions](https://azure.microsoft.com/en-us/products/functions/)
+* [Google Cloud Functions](https://cloud.google.com/functions/)
+* [Netlify Functions](https://functions.netlify.com/)
+* [Cloudflare Workers](https://workers.cloudflare.com/)
+
+## What is event-driven architecture, and is it different than serverless?
+
+_Event-driven architecture_ describes the broader architectural pattern where functions respond to events. Serverless architectures are event-driven architectures. While often associated with serverless, event-driven architectures can exist independently, involving both serverless and traditional infrastructure.
```
aaronkao, line 29: Perhaps this should be organized into H2 "How is Serverless Different Than..."

And the H3 is "Functions as a service" and "Event-driven architectures"

aaronkao, line 29: should we maybe discussed its more pushed based rather than poll based. and then cite services like SNS and SQS and Kinesis and DDB streams that help with these architectures?

aaronkao (commented):
Some comments

scottslowe:
> For some reason this isn't rendering to me with the new layout format the new pages started to use

It's possible this branch was opened before the layout changes were merged.

themes/default/content/what-is/what-is-serverless.md
scottslowe: As far as I am aware this would have to be done by editing the layout. Having multiple CTAs would require multiple layouts, I think. I don't know this for certain; Zack or Sean or Christian are probably better able to answer this question.

themes/default/content/what-is/what-is-serverless.md
scottslowe: I'm no expert, but based on the reading I've been doing AWS Lambda actually _is_ a poll-based solution (I've seen numerous references to automatic polling being set up for Lambda when it's connected to an SQS queue, for example). That being said, I think I can incorporate some additional text here around message queues, streaming solutions, or event buses.

scottslowe: This seems to be adding hierarchy only for the sake of hierarchy. Does it significantly improve the readability of the page, or improve the SEO value of the page?

themes/default/content/what-is/what-is-serverless.md
scottslowe: I've updated the text to point to the serverless architecture templates instead of the blog post.

themes/default/content/what-is/what-is-serverless.md
scottslowe: I've added some text on message queues, event buses, and publish-subscribe services as being used to deliver events among the components of an event-driven architecture.

desteves (approved):

PR #3640, opened 2023-11-17
docs: fix issue in javascript example

nandunbandara:
Fix issue in javascript example on the "Pulumi" concepts page.

## Description
In the javascript code, the use strict mode was written as "user mode" and it was fixed.

## Checklist:

- [] I added aliases (i.e., redirects) for all filename changes.


desteves:
Good catch! TY

desteves (approved):
LGTM

desteves:
I did a quick search `find . -name '**.md' -exec grep -H "user strict" {} \;` and this seems to be the only place with the typo.

PR #3674, opened 2023-11-29
Really add AWSx docs fixes this time

cnunciato:
Reapplies the changes that [had to be reverted yesterday](https://github.com/pulumi/pulumi-hugo/pull/3673) due to an unanticipated quirk in the way Hugo loads remote modules. 

In addition to the documentation updates this PR aims to fix, it also works around the aforementioned quirk by renaming any `go.mod` files that exist under `themes/default/static/programs` to `go.mod.txt` automatically to prevent `go` from omitting their parent folders at `hugo mod get`-time. As a user and program author, you generally shouldn't notice this, but it's unfortunately necessary -- at least until we get this repo pulled back into pulumi/docs. 

I know there's a lot here, but the gist is that the code samples that had previously been baked into `themes/default/content/docs/clouds/aws/guides/_index.md`, `ecr.md`, and `ecs.md` have all been pulled into one of several new Pulumi programs under `themes/default/static/programs` and replaced with the new `{{< example-program >}}` shortcode. So if the programs themselves all look right to the eye, then the PR should be good, as the programs themselves are all being tested now (via `make test`, which we'll begin running nightly once this PR merges).


.github/workflows/scheduled-test.yml
```diff
@@ -0,0 +1,74 @@
...
+  test:
+    runs-on: ${{ matrix.platform }}
+    strategy:
+      matrix:
+        platform:
+          - ubuntu-latest
+        go-version:
+          - 1.18.x
+        node-version:
+          - 16.x
```
desteves, line 24: I believe we have node 18 for other workflows

.github/workflows/scheduled-test.yml
```diff
@@ -0,0 +1,74 @@
...
+
+jobs:
+  test:
+    runs-on: ${{ matrix.platform }}
+    strategy:
+      matrix:
+        platform:
+          - ubuntu-latest
+        go-version:
+          - 1.18.x
```
desteves, line 22: I noticed we have 1.19 in other workflows. Should we bump it to be consistent?

themes/default/static/programs/awsx-ecr-eks-deployment-service-typescript/package.json
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "awsx-ecr-eks-deployment-service-typescript",
+    "devDependencies": {
+        "@types/node": "^14"
```
desteves, line 4: ^18 ?

.github/workflows/scheduled-test.yml
cnunciato: We should keep to the versions Pulumi supports, each of which are linked from [our language pages](https://www.pulumi.com/docs/languages-sdks/). As of today, that is:

* Node 18
* Python 3.8
* Go 1.20
* .NET 6
* Java 11

I've updated all programs and workflows accordingly. Thanks for the callout!

scottslowe: Just a general note regarding Python: 3.8 will be going full EoL (no more security updates) sometime in 2024, so it may be a good idea to bump that to something newer.

PR #3678, opened 2023-11-29
Remove TypeScript only note

bertomoore:
## Description

Remove the note saying Pulumi Crosswalk is only available for TypeScript runtime

## Checklist:


cnunciato:
@bertomoore Thanks for filing this! The wording here could certainly be improved. It's a bit tricky in that while Crosswalk itself is generally available across languages, most of the functionality described on this page (`bucket.onObjectCreated`, `aws.lambda.Function`, etc. ) actually is only available in JavaScript/TypeScript. So rather than delete this note, I'd probably suggest that we update the text to clarify that instead.

Issue #2281, opened 2022-11-30
Update Crosswalk for AWS guides for 1.0


lukehoban:
- [x] ECS - upgrade and add languages
  - https://github.com/pulumi/pulumi-hugo/pull/2274
- [x] ECS - upgrade and add languges
  - https://github.com/pulumi/pulumi-hugo/pull/2284
- [x] API Gateway - add missing languages
- [x] EKS - add languages
  - https://github.com/pulumi/pulumi-hugo/pull/2286
- [x] ELB - upgrade to 1.0 and add all languages
  - fixes pulumi/pulumi-aws#2337
- [x] VPC - rework for 1.0 and add all languages (in progress @danielrbradley)
  - https://github.com/pulumi/pulumi-hugo/pull/2329
- [x] Index - upgrade for 1.0 and add all languages
  - https://github.com/pulumi/pulumi-hugo/pull/2328

Blog posts referencing AWSx:
- [x] https://www.pulumi.com/blog/crosswalk-for-aws-1-0/
- [x] https://www.pulumi.com/blog/crosswalk-for-aws-all-languages/
- [x] https://www.pulumi.com/blog/introducing-pulumi-crosswalk-for-aws-the-easiest-way-to-aws/
- [x] https://www.pulumi.com/blog/pulumi-universal-iac/#crosswalk-for-aws-in-all-pulumi-languages
- [ ] https://www.pulumi.com/registry/packages/aws/how-to-guides/ecs-fargate/
  - (wrongly?) filed under AWS Classic/How-to Guides

Not required:
- Autoscaling - only supported in AWSx Classic
- Cloudwatch - only supported in AWSx Classic
- IAM - says only supports TS for AWSx, but also includes other languages for plain resources, but missing Java and YAML
- Lambda - only supported in TS

lukehoban:
Changing to kind/bug since all of the remaining examples to be updated no longer work by default.

mnlumi:
Added AWSx blog posts that need to be reviewed for 1.0 updates as well.

PR #3693, opened 2023-12-01
Update the ELB and EKS pages in the AWSx docs

cnunciato:
Fixes a _bunch_ of examples and incorrect docs on this page: https://www.pulumi.com/docs/clouds/aws/guides/elb/

- [x] The Creating a Load Balancer example doesn't work at all (`alb.createListener()` is no longer a thing)
- [x] The `enableCrossZoneLoadBalancing` property no longer exists
- [x] The examples presented are only in TypeScript (though all languages are actually supported)
- [x] The Load Balancing EC2 Instance Targets example is very outdated, no longer compiles, and when refactored to fix APIs also doesn't work because the security group doesn't allow ingress, only egress
- [x] The API for creating a load balancer in a VPC went from `external` to `internal` 
- [x] There's no longer a `vpc` property on the ALB resource at all (you have to use `subnetIds` now)
- [x] The docs refer to `createListener` as well, so this needs to be removed/reworded
- [x] The `ApplicationTargetGroup` and `NetworkTargetGroup` resources no longer exist
- [x] There's no longer a `defaultAction` property, only `defaultActions`
- [x] There's no longer a `createTargetGroup` either

Also updates the https://www.pulumi.com/containers/ and https://www.pulumi.com/crosswalk/aws/ pages.

Fixes https://github.com/pulumi/docs/issues/10160.
Fixes https://github.com/pulumi/docs/issues/10230. 
Also covers some of https://github.com/pulumi/pulumi-hugo/issues/2281.


config/_default/config.yml
```diff
@@ -11,6 +11,13 @@ module:
           target: content
         - source: static
           target: static
+          excludeFiles:
```
cnunciato, line 14: This is an optimization to make `hugo server` not watch for changes to dependencies of example programs.

themes/default/content/docs/clouds/aws/guides/elb.md
```diff
@@ -108,73 +95,21 @@ To target an EC2 instance with your load balancer, you must do the following:
...
-};
-```
+{{< example-program path="awsx-load-balanced-ec2-instances" >}}
 
 After deploying this using `pulumi up`, we will have a fully functional endpoint:
 
 ```bash
-$ for i in {1..5}; do curl http://$(pulumi stack output endpoint); done
+$ for i in {1..5} ; do curl "http://$(pulumi stack output endpoint)" ; done
 Hello World, from Server 1!
```
interurban, line 110: is this suppose to 5, then 4 .. (and so on?)

interurban (approved):
just a ton of amazing work in here christian !!!

PR #3713, opened 2023-12-07
[project-file] add newly supported template fields

dschaller:
## Description
Template blocks now support optional `displayName` and `metadata` fields that are used by the console to aid in the discovery of any template (pulumi-hosted or self-hosted).

## Checklist:

### DIFF
--- a/themes/default/content/docs/concepts/projects/project-file.md
+++ b/themes/default/content/docs/concepts/projects/project-file.md
@@ -157,7 +157,7 @@ Schemas are only valid for project property keys. For setting the value of a pro
 
 | Name | Required | Description |
 | - | - | - |
-| `displayName` | optional | A user-friendly name for the template. |
+| `displayName` | optional | A user-friendly name for the template. This should follow `Title Case` format and be succinct. |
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/concepts/projects/project-file.md
```diff
@@ -154,8 +157,10 @@ Schemas are only valid for project property keys. For setting the value of a pro
 
 | Name | Required | Description |
 | - | - | - |
+| `displayName` | optional | A user-friendly name for the template. |
```
lukehoban, line 160: Minor - but should we make any suggestion on format?  I think we generally are encouraging Title Case and relatively short?

lukehoban (approved):
🙌

cnunciato (changes requested):
Ok to close this PR, assuming we aren't currently planning to do this?

dschaller:
I feel like we might still want to add this but should maybe denote that it requires version>=3.95

dschaller:
@cnunciato updated - let me know what you think

cnunciato (approved):

PR #3719, opened 2023-12-07
Use local mode for program tests

cnunciato:


cnunciato:
Ah, hold a moment -- I didn't mean to check in these `go.mod.txt` files. Will remove.

scottslowe (approved):

PR #3720, opened 2023-12-07
Add footer links to examples

### LINKED ISSUE #1766: repo.url not valid in 0.40.0 (latest) of awsx
[themes/default/content/docs/guides/crosswalk/aws/ecr.md](https://www.pulumi.com/docs/guides/crosswalk/aws/ecr/)

Frustrating issue when it's not clear which version the docs refer to - `repo.url` is not a valid way to get the repository URL in `v0.40.0` of awsx. Had to use `repo.repository.repositoryUrl` instead.

But I upgraded to `1.0.0-beta.7` instead of trying to make `0.40.0` work - should make it easier to tell that the docs like this are not compatible with the latest, default installed version of `awsx`.

cnunciato:
Adds a footer containing a GitHub icon that links to the example on GitHub:

image

The goal here is to give users a way to see the full example, with `package.json`/`requirements.txt`, along with any additional dependencies (like Dockerfiles referenced in the snippet, etc.).

Fixes #1766
Fixes https://github.com/pulumi/docs/issues/9665


toriancrane:
Ahh okay seeing it on the screenshot was helpful in locating it. I was looking at the literal footer of the page lol I think it is awesome to have this, but I think we should also have a more visible/intuitive presentation of at least the pulumi-related packages (which version of awsx, which version of python/typescript, which version of the azure-native provider etc).

desteves (approved):
I tried the preview with various inspect dimensions (iPad, mobile etc). The footer link works. All LGTM

scottslowe (approved):

PR #3723, opened 2023-12-08
Updates ESC Code Examples for GCP

toriancrane:
## Description

Updates the GCP code examples on the OIDC and esc gcp secrets provider pages to move the GPC access token to the `environmentVariables` section.


scottslowe (approved):

arunkumar611:
Thank you, Torian. As discussed in slack, we should have the GCP token specified in both pulumiConfig and environmentVariables. 

also for fixing the GCP secrets page, I already made that change! https://github.com/pulumi/pulumi-hugo/pull/3714

PR #3726, opened 2023-12-08
Add note wrt opening ESC secrets

desteves:
Adds an info note to clarify that esc open is the correct command when opening a secret.

### DIFF
--- a/themes/default/content/docs/pulumi-cloud/esc/environments.md
+++ b/themes/default/content/docs/pulumi-cloud/esc/environments.md
@@ -65,10 +65,6 @@ $ esc env set myorg/test foo bar
 
 To retrieve a single value and its definition, use `esc env get <environment-name> <key>`:
 
-{{% notes type="info" %}}
-Use [`open`](#opening-an-environment) to access secrets.
-{{% /notes %}}
-
 ```bash
 $ esc env get myorg/test foo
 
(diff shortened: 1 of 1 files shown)


desteves:
@arunkumar611 let me know if this'll work for https://github.com/pulumi/pulumi-hugo/issues/3727

arunkumar611:
We are working on a new doc update that might help with this issue. I'm not sure if this change alone will explain away the ambiguity on get vs. open. 

Instead of the new note - maybe we could modify the existing one 

from: 

Note that esc env get returns only statically defined plain-text values and definitions; it does not resolve provider-managed values or secrets. (For these, get returns the item definition with a value of [unknown].) To resolve dynamically retrieved values or secrets, you must instead [open the environment](http://localhost:1313/docs/pulumi-cloud/esc/environments/#opening-an-environment).

to 

Note that esc env returns only statically defined plain-text values and definitions; It does not return secrets defined in the environment or resolve values from specified provider configurations. (For these, get returns the item definition with a value of [unknown].) To resolve dynamically retrieved values or secrets, you must instead [open the environment](http://localhost:1313/docs/pulumi-cloud/esc/environments/#opening-an-environment).

desteves:
@arunkumar611 Please take a look at the new changes based on the feedback provided :)

desteves:
@arunkumar611, any thoughts on merging these changes?

arunkumar611 (approved):
LGTM

PR #3728, opened 2023-12-08
Add GitHub icon to the example-program partial also

cnunciato:
Forgot to add this in #3720! It adds the GitHub icon to any `example-program`s we show on marketing pages as well (via the `example-program` _partial_):

image


interurban (approved):
Awesome improvement !

PR #3735, opened 2023-12-11
Update the API Gateway docs

cnunciato:
This change updates the AWS API Gateway docs to bring all examples current, covered in all supported languages, and pulled into example programs so they can be tested going foreward.

There's a lot here, I know! 😅  But it all bubbles up into one page:

[/docs/clouds/aws/guides/api-gateway/](http://pulumi-hugo-origin-pr-3735-1f07fc2c.s3-website.us-west-2.amazonaws.com/)

If this page reads well, all of the examples look right, and there aren't any files in the PR that shouldn't be, you should feel comfortable approving the PR. All of the examples definitely work (save for a few that I've made sure we have issues for) and should all be consistent in terms of their behavior. To make this work, I had to consolidate several free-floating snippets into programs, so I've updated the prose accordingly, but tried not to make too many substantive changes to the text.

This PR also happens to add support for quoting line ranges across languages for a given example. I ended up not needing to use this in this PR, but I left it in anyway, as I'm sure we'll find a use for it soon enough. To use it, just pass the lines you want to include alongside each language:

```
{{< example-program path="awsx-apigateway-s3" languages="typescript:5-8,python:4-12,go:10-15" >}}
```

.. and so on. I also added a little enhancement that conditionally toggles the language tab based on whether the example exists for that language, just so you don't have to remember to do that (and to prevent showing tabs that don't have any content). All of this works for both versions, shortcode and the partial.

Fixes https://github.com/pulumi/pulumi-hugo/issues/2281
Fixes https://github.com/pulumi/docs/issues/10266


desteves (commented):
Browsed the preview, LGTM.
Testing all the go pulumi programs locally.

desteves (approved):
Testing all programs -- 1702910636
Testing awsx-apigateway-api-keys-go/
Testing awsx-apigateway-auth-cognito-go/
Testing awsx-apigateway-auth-lambda-go/
Testing awsx-apigateway-custom-domain-go/
Testing awsx-apigateway-http-proxy-go/
Testing awsx-apigateway-lambda-go/
Testing awsx-apigateway-openapi-full-go/
Testing awsx-apigateway-openapi-route-go/
Testing awsx-apigateway-s3-go/
Testing awsx-apigateway-validation-types-go/


All tests passed -- 1702910636

PR #3756, opened 2023-12-15
Update go mods for examples

desteves:
Fixes https://github.com/pulumi/pulumi-hugo-private/issues/186 
Updates go mod dependencies in the go examples to match.


cnunciato (approved):
🙌

PR #3758, opened 2023-12-15
Update links to Python SDK docs

justinvp:
This updates all links under https://www.pulumi.com/docs/reference/pkg/python/ based on the new standard Python docs site. In some cases, we had previously been linking to Python-specific docs, when instead the links should have been to more general conceptual docs.

Related: https://github.com/pulumi/docs/pull/10279


cnunciato (approved):
We'll merge this right after https://github.com/pulumi/docs/pull/10279 goes out.

PR #3826, opened 2024-01-10
Update Tutorial Code Injection

toriancrane:
## Description

This PR swaps out older code injection method with the updated testable code injection framework.

Resolves [this issue](https://github.com/pulumi/docs/issues/10532).


toriancrane:
> The only thing I noted is that the first TypeScript example doesn't show the comments for both step 1 (create the EC2 instance) and step 2 (create the security group), whereas the Python and YAML code snippet show both steps. Is this intentional?

@scottslowe good catch, thank you! fixed.

desteves (approved):

PR #3830, opened 2024-01-10
Note that Python 3.12 is not yet supported

justinvp:
## Description

There is an upstream issue that is currently preventing us from being able to support Python 3.12. Add a note about this in the docs until we're able to support Python 3.12.

Part of https://github.com/pulumi/home/issues/3259 (internal)

## Checklist:

### DIFF
--- a/themes/default/content/docs/languages-sdks/python/_index.md
+++ b/themes/default/content/docs/languages-sdks/python/_index.md
@@ -19,7 +19,7 @@ aliases:
 Pulumi supports writing your infrastructure as code in the Python language running on any [supported version](https://devguide.python.org/versions/#versions).
 
 {{% notes type="warning" %}}
-Python 3.12 is not yet supported due to an upstream issue in the `grpcio` package that the `pulumi` PyPi package depends on. See [pulumi/pulumi#14258](https://github.com/pulumi/pulumi/issues/14258).
+Python 3.12 is not yet supported. See [pulumi/pulumi#14258](https://github.com/pulumi/pulumi/issues/14258) for details.
 {{% /notes %}}
 
 {{< install-python >}}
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/languages-sdks/python/_index.md
```diff
@@ -18,6 +18,10 @@ aliases:
 
 Pulumi supports writing your infrastructure as code in the Python language running on any [supported version](https://devguide.python.org/versions/#versions).
 
+{{% notes type="warning" %}}
+Python 3.12 is not yet supported due to an upstream issue in the `grpcio` package that the `pulumi` PyPi package depends on. See [pulumi/pulumi#14258](https://github.com/pulumi/pulumi/issues/14258).
```
tgummerer, line 22:
```suggestion
Python 3.12 is not yet supported. See [pulumi/pulumi#14258](https://github.com/pulumi/pulumi/issues/14258).
```

Not sure if we need to mention the details of the upstream issue here?  It feels like we're pushing the blame upstream, but we're responsible for which dependencies we pull in, and if feels like we should probably own that.

justinvp: Agreed! It is our problem and our responsibility. And yet, I'd still like to provide some indication of "why" without pushing the blame. Perhaps the link to the issue is enough of a "why"?

tgummerer: I felt like the issue is enough of a why, or maybe just saying "due to an issue in a dependency of the `pulumi` PyPi package" if we want to be more verbose?  Having the package named feels a little bit like calling it out, which I think is what we should avoid.  I don't have a very strong opinion on whether we do something like that or just link to the issue.

tgummerer (approved):

PR #3860, opened 2024-01-20
Add a post-install note to the getting-started guide

### LINKED ISSUE #1671: some users dont remember to check their path or restart/refresh their terminal after installing pulumi
in a few usability studies, and also observed with users outside of studies some folks forget to check their path or restart/refresh their terminal after installing pulumi. this leads some folks to assume that pulumi did not install correctly.

https://www.pulumi.com/docs/get-started/install/ mentions to check ur path and ur version the get started guides do not. can we do better in this area to catch this problem before folks hit it?

relatedly even if the cli gives them this warning after install, many folks over look it.

cnunciato:
Adds a brief note to the guide about how to verify Pulumi installed successfully.

Fixes #1671.

### DIFF
--- a/themes/default/layouts/shortcodes/install-pulumi.html
+++ b/themes/default/layouts/shortcodes/install-pulumi.html
@@ -10,13 +10,35 @@
         </div>
     </pulumi-choosable>
     <pulumi-choosable type="os" value="windows">
+        {{- .Inner -}}
         <div class="highlight">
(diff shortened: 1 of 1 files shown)


scottslowe (commented):
Just wondering if we should update the `static/latest-version` file to something more recent than `v2.23.1`?

scottslowe (approved):

PR #3870, opened 2024-01-24
fix(names.md): Wrong python syntax in examples

SnoozeFreddo:
python dicts do not accept assignments {test='blub'}. There is also a parenthesis closing which was not needed.

### DIFF
--- a/themes/default/content/docs/concepts/resources/names.md
+++ b/themes/default/content/docs/concepts/resources/names.md
@@ -140,9 +140,7 @@ let role = new aws.iam.Role("my-role", {
 {{% choosable language python %}}
 
 ```python
-role = iam.Role('my-role', {
-    name='my-role-001'
-})
+role = iam.Role('my-role', name='my-role-001')
 ```
 
 {{% /choosable %}}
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/concepts/resources/names.md
```diff
@@ -141,7 +141,7 @@ let role = new aws.iam.Role("my-role", {
 
 ```python
 role = iam.Role('my-role', {
-    name='my-role-001'
+    'name': 'my-role-001'
```
toriancrane, line 144: Hi @SnoozeFreddo,

Thank you for your submission! I actually believe the issue here is the use of the curly braces. This should actually be:

```python
role = aws.iam.Role('my-role', 
    name='my-role-001'
)
```

themes/default/content/docs/concepts/resources/names.md
```diff
@@ -219,7 +219,7 @@ let role = new aws.iam.Role("my-role", {
 
 ```python
 role = iam.Role('my-role', {
```
toriancrane, line 221: The same thing here, I believe it is these wrapping curly braces that are the problem.

```python
role = iam.Role('my-role',
    name=f'my-role-{ pulumi.get_project() }-{ pulumi.get_stack() }'), 
    opts=ResourceOptions(delete_before_replace=True)
)
```

Attempting to deploy with a colon in Python actually gives me an error when running `pulumi up`.

I do believe that this is an indicator that the examples on this page can be improved, so I will be creating a separate issue for that.

toriancrane (changes requested):
Thank you for your feedback! I have left a few comments for recommended changes.

themes/default/content/docs/concepts/resources/names.md
SnoozeFreddo: I just assumed that the internal *kwargs option will somehow magically work. I also don't have aws, i therefore couldn't check if it's working. 

Anyway, i just pushed changed to put it in the format you mentioned and used 'black' on it.

SnoozeFreddo:
> Thank you for your feedback! I have left a few comments for recommended changes.

done

toriancrane (approved):
Looks good to me!

PR #3877, opened 2024-01-25
Remove note about Python 3.12 not being supported

justinvp:
Python 3.12 is now supported with the v3.103.0 release of the Pulumi Python SDK (https://pypi.org/project/pulumi/).

Ref: https://github.com/pulumi/pulumi/pull/15190


julienp (approved):

cnunciato (approved):
🎉

PR #3885, opened 2024-01-29
Add segment consent form

zchase:
## Description

Fixes: https://github.com/pulumi/marketing/issues/935

## Checklist:

### DIFF
--- a/themes/default/layouts/_default/baseof.html
+++ b/themes/default/layouts/_default/baseof.html
@@ -14,5 +14,7 @@
         {{ block "footer" . }}
             {{ partial "footer.html" . }}
         {{ end }}
+
+        <div id="segment-consent-manager"></div>
     </body>
 </html>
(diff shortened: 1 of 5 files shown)


SaraDPH:
@zchase, the consent form looks good. It's a bit easy to miss on a big screen - and EU people need to opt-in, by default lack of response means opt-out. Also, could you let me know if you tested the consent data flow to Google Tag Manager?

@lindsaymarolich, we need to put on our roadmap a v2.0. This version complies with CNIL, CCPA, LGPD, SB-220, and PDPA. Yet, it's in a grey area for GDPR and lacks some ICO best practices.

zchase:
@SaraDPH There isn't a way to prompt on a 3rd visit because we have no way of tracking them until they opt in. The banner will show on the screen until they either opt-in or opt-out.

We will need to defer to @anita-trimbur on how to make the banner more visible.

SaraDPH:
@zchase doesn't Segment collect that data? To some extent, our "owned" cookies can be considered Essential.

It shouldn't be a blocker for banner implementation, but ideally, it should have that as a next step behavior trigger, three visits if opt-out or not yet opt-in, show banner.

zchase:
@SaraDPH Segment won't load until a user opts-in so we won't have any data until the user takes an action. We do collect user preferences in local storage so maybe we _could_ key off that since we set defaults. I can look into it adding some date field we can key off of.

shughes26 (approved):

sean1588 (approved):

PR #3889, opened 2024-01-30
Update ESC syntax reference

lblackstone:
## Description
* Add note about ciphertext on fn::secret
* Add example for fn::fromBase64


toriancrane (approved):
LGTM!

PR #3890, opened 2024-01-31
Correct python logging call in logging.md

SPFZ:
## Description

The current python logging example does only work if you do `import pulumi as log`. 

This PR corrects that and make it compatible with the quickstart tutorial and other examples in this documentation by relying on `import pulumi`.

## Checklist:

- [X] If blogging, I have reviewed the [blogging guide](https://github.com/pulumi/pulumi-hugo/blob/master/BLOGGING.md).
- [X] I added aliases (i.e., redirects) for all filename changes.
- [X] If making css changes, I rebuilt the bundle.


toriancrane (approved):
Looks good to me!

PR #3915, opened 2024-02-07
Inputs/Outputs Content Enhancement

toriancrane:
## Description

This PR provides the following Phase 1 updates for the [Inputs/Outputs content enhancement project](https://github.com/pulumi/docs/issues/10553):

- Breaks down the single page Inputs/Outputs content into multiple pages
- Makes a small update to the outputs section of the landing page
- Adds enhanced content explanations and code examples on the Apply page
- Adds enhanced content explanations and code examples on the All page
- Moves the "Assets & Archives" and "Serialization" pages from under Inputs/Outputs to Concepts.

Content enhancements for the remaining pages (Lifting, Outputs and Strings, Outputs and JSON) will be rolled out in Phase 2+.

Relevant URL paths:
- Landing page: `/docs/concepts/inputs-outputs/`
- Apply page:  `/docs/concepts/inputs-outputs/apply/`
- All page: `/docs/concepts/inputs-outputs/all/`

Resolves the following issues:
- https://github.com/pulumi/docs/issues/10762
- https://github.com/pulumi/docs/issues/10554
- https://github.com/pulumi/docs/issues/10555
- https://github.com/pulumi/docs/issues/10556


scottslowe:
@toriancrane I took a quick look at this early version per your Slack message. Overall, I think the organization and way you're breaking up the content works well.

Two comments:
1. This is probably due to the early stage of the work, but the breakdown of the sub-pages feels a bit...tangled. I'm confident you already know this and will be working to improve it as your work progresses, but did want to point it out.
2. I don't think "Assets and Archives" and "Function Serialization" belong under "Inputs and Outputs", but it looks like they are in this PR. Might be something to check.

Keep up the great work!

toriancrane:
> @toriancrane I took a quick look at this early version per your Slack message. Overall, I think the organization and way you're breaking up the content works well.
> 
> Two comments:
> 
> 1. This is probably due to the early stage of the work, but the breakdown of the sub-pages feels a bit...tangled. I'm confident you already know this and will be working to improve it as your work progresses, but did want to point it out.
> 2. I don't think "Assets and Archives" and "Function Serialization" belong under "Inputs and Outputs", but it looks like they are in this PR. Might be something to check.
> 
> Keep up the great work!

Hey @scottslowe ,

Thank you for the feedback! Regarding your comments:

1. Can you provide a bit more clarification on what you mean by "tangled"?
2. The location of these two pages are currently under Inputs/Outputs in our docs. I just left them where they already are.

`/docs/concepts/inputs-outputs/assets-archives/`

image

If they actually do not belong here, please let me know where they are supposed to live and I can work on including that as part of this update.

themes/default/content/docs/concepts/inputs-outputs/apply.md
```diff
@@ -0,0 +1,818 @@
...
+meta_image: /images/docs/meta-images/docs-meta.png
+menu:
+  concepts:
+    weight: 2
+    parent: inputs-outputs
+---
+
+## Overview
+
+[Outputs](/docs/concepts/inputs-outputs/#outputs) are asynchronous, meaning their actual plain values are not immediately available. As such, there are limitations on the ways in which you can retrieve these values.
```
Frassle, line 15: I think it's worth making very clear that not only are outputs async but in some cases _there is no plain value_. E.g. during a preview the outputs for a new resource don't have any possible value.

toriancrane: @Frassle Right now there is this note is at the bottom of the page:

{{< notes type="info" >}}
During some program executions, {{< pulumi-apply >}} doesn’t run. For example, it won’t run during a preview, when resource output values may be unknown. Therefore, you should avoid side-effects within the callbacks. For this reason, you should not allocate new resources inside of your callbacks either, as it could lead to `pulumi preview` being wrong.
{{< /notes >}}

Is this referring to the same thing? If so, should it be moved higher up?

Frassle: Yup same thing, I think that should be more prominent. A lot of users think output is harder than it ought to be because they think it's just an async.

julienp (commented):
Great to see this getting some documentation love! I read through the preview of the new docs to improve my understanding of this topic and noticed a couple small things.

themes/default/content/docs/concepts/inputs-outputs/_index.md
```diff
@@ -320,1662 +318,18 @@ In this case, Pulumi is taking the _output_ from one resource and using it as th
...
-
-Pulumi automatically captures dependencies when you pass an output from one resource as an input to another resource. Capturing these dependencies ensures that the physical infrastructure resources are not created or updated until all their dependencies are available and up-to-date.
-
-Because outputs are asynchronous, their actual plain values are not immediately available. If you need to access an output’s plain value—for example, to compute a derived, new value, or because you want to log it—you have these options:
-
-- [Apply](#apply): a callback that receives the plain value, and computes a new output
-- [Lifting](#lifting): directly read properties off an output value
-- [Interpolation](#outputs-and-strings): concatenate string outputs with other strings directly
-
-## Apply
```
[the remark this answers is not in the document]
toriancrane: It does point to the new pages, this section is showing what was removed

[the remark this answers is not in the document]
julienp: I meant the note here http://pulumi-hugo-origin-pr-3915-c91ac70c.s3-website.us-west-2.amazonaws.com/docs/concepts/inputs-outputs/#apply but I can't add a comment on unchanged lines in GitHub.

Screenshot 2024-02-12 at 14 15 31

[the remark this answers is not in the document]
toriancrane: Ahhh i see it now, thank you!

themes/default/content/docs/concepts/inputs-outputs/_index.md
```diff
@@ -320,1662 +318,18 @@ In this case, Pulumi is taking the _output_ from one resource and using it as th
...
-    domainName: "example.com",
-    validationMethod: "DNS",
-});
-let certValidation = new aws.route53.Record("cert_validation", {
-    records: [
-        certCertificate.domainValidationOptions[0].resourceRecordValue
-    ],
-...
-```
+- [Apply](/docs/concepts/inputs-outputs/apply/): a callback that receives the plain value and computes a new output
```
scottslowe, line 327: Should we tweak the wording here to create a tighter alignment/integration with the sidebar headings? For example, "Apply: a callback that receives the plain value and computes a new output, generally used for accessing single values" (or similar). Not completely convinced this is necessary, but wanted to submit it for consideration.

themes/default/content/docs/concepts/inputs-outputs/apply.md
```diff
@@ -0,0 +1,822 @@
...
+This example is not applicable in YAML.
+```
+
+{{% /choosable %}}
+
+This is where {{< pulumi-apply >}} comes into play. As mentioned before, all properties of a resource are of type Output<T>, meaning the values only become known after the infrastructure has been provisioned. When a Pulumi program is executed with `pulumi up`, the {{< pulumi-apply >}} function will wait for the resource to be created and for its properties to be resolved before printing the desired value of the property. This is not something a standard `print | log` statement is capable of doing.
+
+### Accessing single output values
+
+The syntax of {{< pulumi-apply >}} is shown below:
```
scottslowe, line 490: HUGE +1 for adding language-specific syntax reference for `apply`. This was a big miss in our docs previously---great work!

themes/default/content/docs/concepts/inputs-outputs/lifting.md
```diff
@@ -0,0 +1,316 @@
...
+title: Accessing nested properties with Lifting
+h1: Accessing nested properties with Lifting
+meta_image: /images/docs/meta-images/docs-meta.png
+menu:
+  concepts:
+    weight: 4
+    parent: inputs-outputs
+---
+
+While often, Outputs return asynchronous values that wrap primitive types like strings or integers, sometimes an output has an object with deeply nested values. These properties need to be passed to other inputs as well.
```
scottslowe, line 13: This wording feels awkward. Maybe it's just the extra comma after "While often"?

themes/default/content/docs/concepts/inputs-outputs/lifting.md
```diff
@@ -0,0 +1,316 @@
...
+
+For example, to read a domain record from an ACM certificate, you need to access the domain validation options, which returns an array. Because that value is an output, we would normally need to use {{< pulumi-apply >}}:
+
+{{< chooser language "javascript,typescript,python,go,csharp,java" >}}
```
scottslowe, line 17: Love the addition of the explanations in Go regarding the use of `Index()` and `Elem()`, again a miss in the docs previously.

scottslowe (approved):
A few minor comments/suggestions, but nothing that is worth blocking on. HUGE improvement over the prior versions IMO!

themes/default/content/docs/concepts/inputs-outputs/lifting.md
toriancrane: The lifting page will be updated in a future phase of edits. This round is just the reorg, apply, and all pages for now

themes/default/content/docs/concepts/inputs-outputs/_index.md
```diff
@@ -187,7 +185,7 @@ resources:
 {{% notes %}}
 _Plain value_ in this document is used to describe a standard string, boolean, integer or other typed value or data structure in your language of choice. _Plain value_ is a way of differentiating these language specific values from Pulumi's asynchronous values.
 
-Once an {{< pulumi-output >}} value has returned (generally from an API) to Pulumi, it can be used as a _plain value_. For more information, see [apply](#apply).
+Once an {{< pulumi-output >}} value has returned (generally from an API) to Pulumi, it can be used as a _plain value_. For more information, see [apply](/docs/concepts/inputs-outputs/apply/).
```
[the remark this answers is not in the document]
toriancrane: I agree, removed it

cnunciato:
>  I wonder if it wouldn't make more sense to focus on _only_ "use case" subpages, like this:
>     * Accessing Single Outputs
>     * Accessing Output Properties
>     * Accessing Multiple Outputs
>     * Outputs and Strings
>     * Outputs and JSON
> 
> This seems to provide a clear link between method (`apply` == accessing single outputs, lifting == accessing output properties, `all` == accessing multiple outputs, etc.) and use case.

Still reviewing, but wanted to chime in on this one. The more I think about this, the more I like the idea of the choices being between "Apply", "All", and "Helpers", because that's really this all boils down to I think. Lifting and the string and JSON methods are ultimately just helpers -- wrappers around Apply and All -- but when you're new to Pulumi, that's incredibly non-obvious, and it's this non-obviousness that makes understanding which one to reach for and why incredibly confusing. If on the other hand we were to lean into Apply and All as "the way", and then presented these three categories of helpers as extra things that can make working with Apply and All a little nicer, users would be able to understand how they all relate to one another -- and hopefully grok that everything ultimately comes down to Apply and All.

Thoughts on this? Again I think it's both simplifying and clarifying, so it makes sense to me, but curious what you all think.

themes/default/content/docs/concepts/inputs-outputs/apply.md
```diff
@@ -0,0 +1,822 @@
...
+meta_image: /images/docs/meta-images/docs-meta.png
+menu:
+  concepts:
+    weight: 2
+    parent: inputs-outputs
+---
+
+## Overview
+
+[Outputs](/docs/concepts/inputs-outputs/#outputs) are asynchronous, meaning their actual plain values are not immediately available. Their values will only become available once the resource has finished provisioning. The asynchronous nature of Outputs is also why, when certain operations such as [`pulumi preview`](/docs/cli/commands/pulumi_preview/) runs, the outputs for a new resource do not yet have any possible values. As such, there are limitations on the ways in which you can retrieve and interact with these values.
```
interurban, line 15: I wonder if adding both the reason + benefits of Pulumi would be helpful? (to folks new to pulumi at least but we could save it for a tutorial)  

The reason outputs are asynchronous is because provisioning resources is an asynchronous operation. It takes time for a cloud provider and Pulumi optimizes the process by executing operations in parallel, rather than sequentially.

themes/default/content/docs/concepts/inputs-outputs/apply.md
```diff
@@ -0,0 +1,822 @@
...
+
+Resources:
+    34 unchanged
+
+Duration: 12s
+```
+
+We can now see the value of the VPC ID property that we couldn't see before when using a regular `print | log` statement.
+
+### Creating new output values
```
interurban, line 692: I really like this use case; I wonder would we consider putting it above the VPC example?

themes/default/content/docs/concepts/inputs-outputs/outputs-and-json.md
```diff
@@ -0,0 +1,564 @@
...
+
+var policyWithNoStatements = Pulumi.Output.JsonDeserialize<IAMPolicy>(jsonIAMPolicy).Apply(policy =>
+{
+    // delete the policy statements.
+    policy.Statement = Pulumi.Output.Create(new List<StatementEntry> { });
+    return policy;
+})
+```
+
+For more details [view the .NET documentation](/docs/reference/pkg/dotnet/Pulumi/Pulumi.Output.html)
```
cnunciato, line 560:
```suggestion
For more details [view the .NET documentation](/docs/reference/pkg/dotnet/Pulumi/Pulumi.Output.html).
```

cnunciato (approved):
Nice work on this!  🙌 ⭐🍦 

First, apologies for how long it took me to get through it! Appreciate your patience. 

I've left you a lot of feedback, but none of it need block -- although I'd fairly strongly prefer to have us leave the Inputs & Outputs section where it is in the nav, and I'd love to discuss collapsing the "helper pages" into a single page as I've proposed as well. Ultimately I'm comfortable deferring to you on both, so I'm approving to leave it to you, but it might be nice to discuss them first as a team if you're up for it.

The rest of my feedback is has mainly to do with our existing content, which still leaves a _log_ of room for improvement, IMO. (The stuff you've added is great; I had only a few comments for you on that.) I'd love for us to address at least some of these issues now, since we're here (and given how important this content is for new users), but it definitely needn't be attached to this PR (which is another reason why I'm approving for now). If you see anything here you agree is worth doing now, awesome! If not, we can discuss as a team and file follow-up whatever follow-up issues we feel are necessary. Mainly I just wanted to capture all this stuff it while it was all top of mind to give us something to talk about and work from.

Thank you again for the excellent work! This is great, and another big step forward.

toriancrane:
> I think I'm mainly proposing collapsing the lifting, strings, and JSON pages (which are all pretty small) into a single third page with a name like "using output helpers" or similar.

These three pages in particular have not received any kind of updates yet just FYI. I plan to add enhancements to them in the next Phase of edits so they may get larger or they may still stay small. I think it will make more sense to know which approach to take (i.e. collapsing into a single page vs keeping separate) once I have a better sense of what the updated content will look like. I want to be able to show meaningful examples, response objects, and deeper explanations where needed while also keeping the pages at reasonable sizes.

themes/default/content/docs/concepts/inputs-outputs/_index.md
```diff
@@ -320,1662 +318,18 @@ In this case, Pulumi is taking the _output_ from one resource and using it as th
...
-{{% /choosable %}}
-{{% choosable language python %}}
+## Tracking dependencies
 
-```python
-certificate = aws.acm.Certificate('cert',
-    domain_name='example.com',
-    validation_method='DNS'
-)
+Outputs are also how Pulumi tracks dependencies between resources. When an output from one resource has been returned from the cloud provider API, Pulumi can link the two resources together and pass it as the input to another resource.
```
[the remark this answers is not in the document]
toriancrane: This section is specifically meant to be talking about how dependencies work in Pulumi. This is another section that will be fleshed out more in a future phase.

themes/default/content/docs/concepts/inputs-outputs/outputs-and-strings.md
```diff
@@ -0,0 +1,168 @@
...
+// Would like to produce a string equivalent to: http://{hostname}:{port}/
+var url = Output.tuple(hostname, port)
+        .applyValue(t -> String.format("http://%s:%s/", t.t1, t.t2));
+```
+
+{{% /choosable %}}
+
+{{< /chooser >}}
+
+However, this approach is verbose and unwieldy. To make this common task easier, Pulumi exposes interpolation helpers that allow you to create strings that contain outputs. These interpolation methods wrap [all](/docs/concepts/inputs-outputs/all/) and [apply](/docs/concepts/inputs-outputs/apply/) with an interface that resembles your language's native string formatting functions.
```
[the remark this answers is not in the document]
toriancrane: The updates of this page is not included in this round of edits.

themes/default/content/docs/concepts/inputs-outputs/lifting.md
```diff
@@ -0,0 +1,316 @@
...
+      records:
+        # YAML handles inputs and outputs transparently.
+        - ${cert.domainValidationOptions[0].resourceRecordValue}
+```
+
+{{% /choosable %}}
+
+{{< /chooser >}}
+
+This approach is easier to read and write and does not lose any important dependency information that is needed to properly create and maintain the stack. This approach doesn’t work in all cases, but when it does, it can be a great help.
```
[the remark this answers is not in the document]
toriancrane: The updates of this page is not included in this round of edits, and I do plan to cover this because I actually know the answer now lol I've just been thinking about the best way of how to present that clearly.

themes/default/content/docs/concepts/inputs-outputs/all.md
```diff
@@ -0,0 +1,311 @@
...
+
+{{< chooser language "javascript,typescript,python,go,csharp,java,yaml" >}}
+
+{{% choosable language javascript %}}
+
+```javascript
+const pulumi = require("@pulumi/pulumi");
+// ...
+
+const connectionDetails1 = pulumi.all([sqlServer.ipAddress, database.port])
```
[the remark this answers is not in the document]
toriancrane: I thought about that as well. The names for these data structures are different across languages, so I was hoping the catch-all in the paragraph above would be enough, but maybe I will just add a comment above each to reflect it.

I like the idea of adding something that uses the new structure as an input. I would have to give it a bit more thought as to what would make sense here. Would you consider this a blocking thing or is it something I could reasonably include in the next round of edits? I'm aiming for a minimally viable product that I can publish before COB today.

themes/default/content/docs/concepts/inputs-outputs/_index.md
```diff
@@ -7,7 +7,7 @@ meta_image: /images/docs/meta-images/docs-meta.png
 menu:
   concepts:
     identifier: inputs-outputs
-    weight: 5
+    weight: 1
```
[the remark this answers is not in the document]
toriancrane: This was actually a typo, I will fix this.

themes/default/content/docs/concepts/inputs-outputs/_index.md
```diff
@@ -320,1662 +318,18 @@ In this case, Pulumi is taking the _output_ from one resource and using it as th
...
-            .applyValue(List::of))
-        .build());
-```
-
-{{% /choosable %}}
-
-{{< /chooser >}}
-
-Instead, to make it easier to access simple property and array elements, an {{< pulumi-output >}} lifts the properties of the underlying value, behaving very much like an instance of it. Lift allows you to access properties and elements directly from the {{< pulumi-output >}} itself without needing {{< pulumi-apply >}}. If we return to the above example, we can now simplify it:
+All resources created by Pulumi will have properties which are returned from the cloud provider API. These values are called *outputs*.
```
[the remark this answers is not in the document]
toriancrane: I'm not sure that I agree with this one. I almost feel like that would cause some confusion by stating that outputs are not plain values, and then right after that stating you can access their plain values.

Right after this sentence, we say:

"Outputs are values of type {{< pulumi-output >}}, which behave very much like [promises](https://en.wikipedia.org/wiki/Futures_and_promises) or [monads](https://en.wikipedia.org/wiki/Monad_(functional_programming)). Simply put, outputs are a way of representing values that are not initially known but will become available once the infrastructure resource has completed provisioning, which happens *asynchronously*.

Because outputs are asynchronous, their actual plain values are not immediately available. If you need to access an output’s plain value, you can do so using one of the following options:"

themes/default/content/docs/concepts/inputs-outputs/apply.md
toriancrane: @interurban Can you clarify what you mean here? I'm not sure I understand.

themes/default/content/docs/concepts/inputs-outputs/apply.md
toriancrane: I'll include this explanation on the landing page

themes/default/content/docs/concepts/inputs-outputs/apply.md
```diff
@@ -0,0 +1,822 @@
...
+```python
+<resource>.<property-name>.apply(lambda <property-name>: <function-to-apply>)
+```
+
+{{% /choosable %}}
+
+{{% choosable language go %}}
+
+```go
+<resource>.<property-name>.ApplyT(func(<property-name> string) error {
```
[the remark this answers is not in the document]
toriancrane: @cnunciato I'm really lost on this one. Keep in mind I don't know the first thing about Golang so I don't fully understand what the syntax is doing, only that it works lol

themes/default/content/docs/concepts/inputs-outputs/all.md
```diff
@@ -0,0 +1,311 @@
...
+If you need to access and use multiple outputs together, the `all` function acts like an [`apply`](/docs/concepts/inputs-outputs/apply/) across many resources, allowing you to retrieve and use multiple outputs at the same time. The `all` function waits for all output values to become available and then provides them as _plain values_ to the {{< pulumi-apply >}} function.
+
+This can be used to compute an entirely new output value, such as creating a new string by adding or concatenating outputs from two different resources together, or by creating a new data structure that uses their values. Just like with `apply`, the result of `all` is itself an Output<T>.
+
+### Creating a new string
+
+To demonstate, let’s say we have created a server resource and a database resource, and their Output values are as follows:
+
+```python
+# Example outputs for the server resource
```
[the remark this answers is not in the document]
cnunciato: No problem, here you go, in all languages:

```typescript

const sqlServer = pulumi.output({
    name: "myDbServer",
    ipAddress: "10.0.0.0/24",
});

const database = pulumi.output({
    name: "myExampleDatabase",
    engine: "sql-db",
});

export const sqlServerName = sqlServer.name;
export const databaseName = database.name;
```

```javascript
"use strict";
const pulumi = require("@pulumi/pulumi");

const sqlServer = pulumi.output({
    name: "myDbServer",
    ipAddress: "10.0.0.0/24",
});

const database = pulumi.output({
    name: "myExampleDatabase",
    engine: "sql-db",
});

exports.sqlServerName = sqlServer.name;
exports.databaseName = database.name;
```

```python
import pulumi

sql_server = pulumi.Output.from_input({
    "name": "myDbServer",
    "ipAddress": "10.0.0.0/24",
});

database = pulumi.Output.from_input({
    "name": "myExampleDatabase",
    "engine": "sql-db",
});
```

```go
package main

import (
	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
)

func main() {
	pulumi.Run(func(ctx *pulumi.Context) error {

		sqlServer := pulumi.ToOutput(&pulumi.StringMap{
			"name":      pulumi.String("myDbServer"),
			"ipAddress": pulumi.String("10.0.0.0/24"),
		})

		database := pulumi.ToOutput(&pulumi.StringMap{
			"name":   pulumi.String("myExampleDatabase"),
			"engine": pulumi.String("sql-db"),
		})

		ctx.Export("sqlServerName", sqlServer.ApplyT(func(values map[string]string) string {
			return values["name"]
		}))

		ctx.Export("databaseName", database.ApplyT(func(values map[string]string) string {
			return values["name"]
		}))

		return nil
	})
}
```

```csharp
using System.Collections.Generic;
using Pulumi;

return await Deployment.RunAsync(() =>
{
    var sqlServer = Output.Create(new
    {
        Name = "myDbServer",
        IpAddress = "10.0.0.0/24",
    });

    var database = Output.Create(new
    {
        Name = "myExampleDatabase",
        IpAddress = "sql-db",
    });

    return new Dictionary<string, object?>
    {
        ["sqlServerName"] = sqlServer.Apply(server => server.Name),
        ["databaseName"] = database.Apply(db => db.Name),
    };
});
```

```java
package myproject;

import com.pulumi.Context;
import com.pulumi.Pulumi;
import com.pulumi.core.Output;
import java.util.Map;

public class App {
    public static void main(String[] args) {
        Pulumi.run(App::stack);
    }

    public static void stack(Context ctx) {

        var sqlServer = Output.of(Map.of(
            "name", "myDbServer",
            "ipAddress", "10.0.0.0/24"
        ));

        var database = Output.of(Map.of(
            "name", "myExampleDatabase",
            "engine", "sql-db"
        ));

        ctx.export("sqlServerName", sqlServer.applyValue(values -> {
            return values.get("name");
        }));

        ctx.export("databaseName", database.applyValue(values -> {
            return values.get("name");
        }));
    }
}
```

```yaml
name: myproject
runtime: yaml

variables:
  sqlServer:
    name: "myDbServer"
    ipAddress: "10.0.0.0/24"
  database:
    name: myExampleDatabase
    engine: sql-db

outputs:
  sqlServerName: ${sqlServer.name}
  databaseName: ${database.name}
```

PR #3948, opened 2024-02-16
Add ESC context docs

glena:
## Description

Add docs of new ESC context interpolation and subject customization

https://github.com/pulumi/pulumi-service/issues/15758

## Checklist:

### DIFF
--- a/themes/default/content/docs/esc/environments.md
+++ b/themes/default/content/docs/esc/environments.md
@@ -205,6 +205,30 @@ $ esc env get myorg/test greeting
   • test:10:15
 ```
 
+### Pulumi Contextual information
+
+Contextual information is available and can be used to interpolate on any environment value.
+
+```yaml
+values:
+  personalNamespace: ${context.rootEnvironment.name}/${context.pulumi.user.login}
+```
+
+It can be accessed through the `context` attribute and contains the following information:
+
+```yaml
+context:
+  rootEnvironment:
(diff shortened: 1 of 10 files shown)


themes/default/content/docs/esc/environments.md
```diff
@@ -205,6 +205,30 @@ $ esc env get myorg/test greeting
...
+Contextual information is available and can be used to interpolate on any environment value.
+
+```yaml
+values:
+  personalNamespace: ${context.rootEnvironment.name}/${context.pulumi.user.login}
+```
+
+It can be accessed through the `context` attribute and contains the following information:
+
+```yaml
```
arunkumar611, line 219: I think this one is slightly confusing. Because the users still need to assign a path name to use the Context. copy-pasting this into a YAML won't work every if I assign a path name.

Maybe remove this yaml and we can explain it via text?

arunkumar611:
I think we should add the syntax to syntax reference page: docs/esc/reference/

themes/default/content/docs/esc/environments.md
glena: The idea was to show its structure and hierarchy. As we add more attributes to each object, will look really weird.

I just pushed the change, feel free to push changes

glena:
> I think we should add the syntax to syntax reference page: docs/esc/reference/

done

arunkumar611 (approved):
With the changes, it looks good.

PR #3981, opened 2024-02-28
Fix the location in the portal where to find default stack permissions.

ringods:
## Description

Fix the location in the portal where to find default stack permissions. It is under `Access Management` rather than `General`.

## Checklist:


cnunciato (approved):
Thanks @ringods!

PR #3992, opened 2024-03-01
Update the import doc, adding detail about where to find the type token

### LINKED ISSUE #1265: Document where to get the "type" token for imports
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

Raised in [community slack](https://pulumi-community.slack.com/archives/C84L4E3N1/p1648257992751949):

> VS: Hi guys,
I'm going to import some existing resources with pulumi import [type] [name] [id] command. How can I get type for my resources?
> AN: If you look at the documentation of your provider, on each resource you can see an example of an import command that gives you the type of the resource.
> FW: You can also check the SDK code. https://github.com/pulumi/pulumi-aws/blob/master/sdk/nodejs/s3/bucket.ts#L334 for example is the type for old aws providers s3 bucket resource.
> VS: Thanks, guys.
It's not obvious from the first look. Especially, if you're a newbie to Pulumi. It'd would be great to add this to pulumi import doc (https://www.pulumi.com/docs/reference/cli/pulumi_import/)
> PS: Hey VS - if we update the docs to say “look for the Import sections in the API Docs for the type” would that help - we can also link to an example

We should document this better. Possibly both here in the docs site and as part of the import command itself.

### LINKED ISSUE #3965: Add more context on when to use `ResourceOptions.import` vs `pulumi import`
[themes/default/content/docs/concepts/options/import.md](https://www.pulumi.com/docs/concepts/options/import/)

From a Slack question I asked, @lukehoban added this context, which we should incorporate into the docs IMO:

> Managing imports via code is much better for some workflows, where different stacks from the same project need to import resources in a well-defined upgrade process as changes are rolled out across environments instead of just an imperative pulumi import.

In order to make the docs more user-friendly, I would add something to the effect of "you probably want to use `pulumi import` unless you're doing a fancy multi-stack migration".

If we want to go even further, a user guide illustrating the code-based import process would be a great addition to the docs, although it seems like a relatively niche use case. (I wasn't able to find any prior art in the blog that details how to use `ResourceOptions.import` at all, much less in a step-by-step fashion.)

cnunciato:
Adds details about where to find the type token and lookup property when adopting a resource, and attempts to better clarify the difference between `pulumi import` and the `import` resource option.

Fixes #1265.
Fixes #3965.


themes/default/content/docs/using-pulumi/adopting-pulumi/import/_index.md
```diff
@@ -12,161 +12,184 @@ aliases:
 - /docs/guides/adopting/import/
 ---
 
-Most infrastructure projects require working with existing cloud resources, either by building on top of existing
-resources or adopting existing resources under management with a new and more robust infrastructure provisioning solution.
+Most infrastructure as code projects require working with existing cloud resources, whether those resources were originally created with another IaC tool or manually provisioned with a cloud provider console or CLI. Interacting with a previously created cloud resource with Pulumi typically happens in one of two ways:
 
-No matter how you've provisioned these resources &mdash; manually in your cloud provider's console or CLI, using an
-infrastructure as code tool like Terraform or AWS CloudFormation &mdash; Pulumi enables you to adopt and manage your resources.
+* Referencing the properties of the existing cloud resource in order to use those properties to configure a Pulumi-managed resource.
```
jkodroff, line 17: Debatable, and nit-picky, but because it says "one of two ways", I would use an ordered list. Alternatively, could word the preview paragraph as "one of the following ways".

```suggestion
1. Referencing the properties of the existing cloud resource in order to use those properties to configure a Pulumi-managed resource.
```

themes/default/content/docs/using-pulumi/adopting-pulumi/import/_index.md
```diff
@@ -12,161 +12,184 @@ aliases:
...
+* Adopting the existing resource to bring it under management by Pulumi.
```
jkodroff, line 18:
```suggestion
1. Adopting the existing resource to bring it under management by Pulumi.
```

themes/default/content/docs/using-pulumi/adopting-pulumi/import/_index.md
```diff
@@ -12,161 +12,184 @@ aliases:
...
+
+![Where to find the type token and lookup property for a resource](./token-and-lookup.png)
+
+{{< notes >}}
+Make sure the resource provider is configured in a way that allows it to locate the resource you want to import --- e.g., that the resource is in the same region as other resources in your stack --- or the import operation may fail.
+{{< /notes >}}
+
+## Importing resources with the CLI
+
+The `pulumi import` command looks up the resource using the specified type token and resource identifier, adds the resource to the stack's current state, and emits the code required to manage the resource with Pulumi from that point forward. This option requires the least manual effort, so is generally recommended, and is best suited to projects consisting consisting of only one stack.
```
jkodroff, line 52: I think we need to explain what a type token is. I prefer "type identifier" as it's less jargon-y (if it won't make our docs wildly inconsistent) and may not require much more than an example, e.g. `aws:ec2/vpc:Vpc`, to make it clear what it is (a string that tells Pulumi what resource type it is).

themes/default/content/docs/using-pulumi/adopting-pulumi/import/_index.md
```diff
@@ -12,161 +12,184 @@ aliases:
...
+
+To import an existing cloud resource with the Pulumi CLI, use the following syntax:
+
+```bash
+$ pulumi import <type> <name> <id>
+```
+
+* The first argument, `type`, is the Pulumi type token to use for the imported resource.
+
+    As [mentioned above](#how-import-works), you'll find the type token for a given resource by navigating to the Import section of the resource's API documentation in the [Pulumi Registry](/registry/). The type token of an [Amazon S3 Bucket](/registry/packages/aws/api-docs/s3/bucket/#import) resource, for example, is `aws:s3/bucket:Bucket`.
```
jkodroff, line 62:
```suggestion
    As mentioned in [Where to find the type token and lookup property](#how-import-works), you'll find the type token for a given resource by navigating to the Import section of the resource's API documentation in the [Pulumi Registry](/registry/). For example, the type token of an [Amazon S3 Bucket](/registry/packages/aws/api-docs/s3/bucket/#import) resource, for example, is `aws:s3/bucket:Bucket`.
```

themes/default/content/docs/using-pulumi/adopting-pulumi/import/_index.md
```diff
@@ -12,161 +12,184 @@ aliases:
...
+
+* The second argument, `name`, is the [resource name](/docs/concepts/resources/names) to apply to the resource once it's imported. The generated code will use this name for the resource declaration, so like all Pulumi resource names, it must be unique within the scope of the containing project.
```
jkodroff, line 64:
```suggestion
* The second argument, `name`, is the [resource name](/docs/concepts/resources/names) to apply to the resource once it's imported. The generated code will use this name for the resource declaration (the first parameter in any resource), so like all Pulumi resource names, it must be unique among all resources for this type within the scope of the containing project. (That is, you may have an S3 bucket and a VPC named `foo`, but you can not have 2 S3 buckets named `foo`.)
```

themes/default/content/docs/using-pulumi/adopting-pulumi/import/_index.md
```diff
@@ -12,161 +12,184 @@ aliases:
...
+the destroy will take effect.
 
-```typescript
 import * as pulumi from "@pulumi/pulumi";
 import * as aws from "@pulumi/aws";
 
-const demo = new aws.s3.Bucket("infra-logs", {
-    acl: "private",
+const my_bucket = new aws.s3.Bucket("infra-logs", {
+    arn: "arn:aws:s3:::company-infra-logs",
```
jkodroff, line 123: It really returns the ARN? I thought that was strictly an output.

jkodroff (commented):
This is a major improvement! All suggestions are "could do", no blockers.

jkodroff (approved):
Whoops forgot to approve.

themes/default/content/docs/using-pulumi/adopting-pulumi/import/_index.md
cnunciato: It does! This is the actual output, even on a brand-new empty bucket. I actually wonder whether it's worth filing an issue on this, as you can remove everything except `bucket`, run `up`, and get no changes. What do you think?

themes/default/content/docs/using-pulumi/adopting-pulumi/import/_index.md
cnunciato: Ah yeah, that's better. I'll do that.

cnunciato: It looks like we refer to this as "type token" in more places than I realized. Maybe best to keep to that for consistency for now, but it's something worth discussing for sure.

PR #3996, opened 2024-03-02
Add a note about  precedence in Node.js projects

### LINKED ISSUE #2189: Clarify `main` flag behavior.
[themes/default/content/docs/reference/pulumi-yaml.md](https://www.pulumi.com/docs/reference/pulumi-yaml/)

A user was having pain from unintuitive behavior with the `main` flag.
https://github.com/pulumi/pulumi/issues/11243

We should make it more clear.

cnunciato:
Clarifies that `Pulumi.yaml`'s `main` value takes precedence over `package.json`'s and that paths are relative to `Pulumi.yaml`. Does a bit of minor cleanup as well.

Fixes #2189.

### DIFF
--- a/themes/default/content/docs/concepts/projects/project-file.md
+++ b/themes/default/content/docs/concepts/projects/project-file.md
@@ -14,22 +14,22 @@ aliases:
 - /docs/reference/project-file/
 ---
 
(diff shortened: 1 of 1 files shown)


interurban (approved):
looks good!

PR #4003, opened 2024-03-04
Adding OpenTofu to docs

aaronkao:
## Description
Just adding in a vs page for OpenTofu but also added a TF vs OpenTofu page to capture SEO


phillipedwards (approved):
LGTM

PR #4021, opened 2024-03-07
fix: callback function docs not available

corymhall:
## Description

The `CallbackFunction` is a node only function which means that it does not have any general documentation. The link that is provided to find more information does not have any information on the `CallbackFunction`. Since we do not publish any language specific docs, I don't think there is any better place to point the user to.

fixes https://github.com/pulumi/pulumi-aws/issues/2035

## Checklist:

- ~[ ] If blogging, I have reviewed the [blogging guide](https://github.com/pulumi/pulumi-hugo/blob/master/BLOGGING.md).~
- ~[ ] I have manually confirmed that all new links work.~
- ~[ ] I added aliases (i.e., redirects) for all filename changes.~
- ~[ ] If making css changes, I rebuilt the bundle.~

### DIFF
--- a/themes/default/content/docs/clouds/aws/guides/lambda.md
+++ b/themes/default/content/docs/clouds/aws/guides/lambda.md
@@ -162,9 +162,6 @@ docsBucket.onObjectCreated("docsHandler", new aws.lambda.CallbackFunction("docsH
 });
 ```
 
-For more information about the properties available on `CallbackFunction`, refer to the [API documentation](
-/registry/packages/aws/api-docs/lambda).
-
 ### Register an Event Handler by Creating a Lambda Function Resource
 
 It is possible to create and register serverless event handlers by allocating `aws.lambda.Function` objects


corymhall:
> Your site preview for commit [7d26148](https://github.com/pulumi/pulumi-hugo/commit/7d2614807eeee1bbc95b58af6506af8310357b89) is ready! 🎉
> 
> http://pulumi-hugo-origin-pr-4021-7d261480.s3-website.us-west-2.amazonaws.com.

link has been removed here (below the code snippet)
http://pulumi-hugo-origin-pr-4021-7d261480.s3-website.us-west-2.amazonaws.com/docs/clouds/aws/guides/lambda/#specifying-attributes-on-your-magic-lambda-function

cnunciato (approved):
Ah, I'm actually working on documenting some of these properties right now! 

I think we can do one of two things. We could merge this change now to close out the issue, as what you have here is technically fine. On the other hand, since the user was presumably looking for docs they weren't able to find, another option would be to wait and close out the issue with an updated link that actually does point to more info on `CallbackFunction` (which I'm planning on submitting today, likely by adding to https://www.pulumi.com/docs/concepts/function-serialization/). 

Either one would be fine with me, so I'll leave you an approval for now to keep you unblocked. Up to you.

corymhall:
> Either one would be fine with me, so I'll leave you an approval for now to keep you unblocked. Up to you.

That's great, I'll definitely take the option of more documentation!

PR #4030, opened 2024-03-08
Fix aws-ec2-sg-nginx-server-javascript example

cnunciato:
Looks like this one just didn't get fleshed out with the others alongside it. This should fix it up.


toriancrane (approved):
lgtm

PR #4032, opened 2024-03-08
Inputs Outputs Updates and Linting Fixes

toriancrane:
## Description

This PR provides the following updates:

- [x] Inputs Outputs Lifting code examples and content
- [x] Applies linting to fix Upgrade Examples job
- [x] Adds a redirect to a 404ing page
- [x] Updates outputs and json content on Apply page
- [x] Updates outputs and json content on All page


themes/default/static/programs/aws-simulated-server-interpolate-csharp/Program.cs
```diff
@@ -9,11 +9,12 @@
         Port = "8080",
     });
 
-    // Format takes a FormattableString and expands outputs correctly:
-    var url = Output.Format($"http://{webServer.HostName}:{webServer.Port}/");
+    // // Format takes a FormattableString and expands outputs correctly:
```
cnunciato, line 12: I'm a bit of a stickler when it comes to checking in dead and/or commented code. 😇 (Always some exceptions of course, but rare, and generally accompanied by an additional comment explaining why the code is commented and not just deleted.) 

If this doesn't yet work and isn't referenced anywhere, I'd prefer we removed it and put it back when we're ready to use it.

cnunciato (commented):
Looking good! Few little things for ya.

themes/default/content/docs/concepts/inputs-outputs/all.md
```diff
@@ -165,7 +165,193 @@ Server=tcp:myDbServer.database.windows.net;initial catalog=myExampleDatabase;
...
 There is an easier way to generate a concatenated string value using multiple outputs, and that is by using interpolation. Pulumi exposes interpolation helpers that enables you to create strings that contain outputs. These interpolation methods wrap [all](/docs/concepts/inputs-outputs/all/) and [apply](/docs/concepts/inputs-outputs/apply/) with an interface that resembles your language's native string formatting functions. The below example demonstrates how to create a URL from the hostname and port output values of a web server.
 
-{{< example-program path="aws-simulated-server-interpolate" >}}
+{{< chooser language "javascript,typescript,python,go,csharp,java,yaml" >}}
+
+{{% choosable language javascript %}}
+
+```javascript
+const pulumi = require("@pulumi/pulumi");
+/* Simulated outputs of a webServer resource
```
cnunciato, line 174: I know you're trying to get this wrapped, so apologies for slowing it down! 😇 But I do think this example and the others on this page would be better served if we used actual resources and outputs. That'd make this both clearer and more testable, IMO.

The simulated outputs _look_ simpler on the surface -- but in a way they're actually less clear in that you have to imagine what they stand in for. It'd be better if we just used a real resource or two. We could use just about anything, but here's an example that uses a bucket and bucket object to construct the two URLs the current example exports at the end:

```typescript

const bucket = new aws.s3.Bucket("bucket");

const file = new aws.s3.BucketObject("bucket-object", {
    bucket: bucket.id,
    key: "some-file.txt",
    content: "some-content",
});

// concat takes a list of args and concatenates all of them into a single output:
export const s3Url1: pulumi.Output<string> = pulumi.concat("s3://", bucket.bucket, "/", file.key);

// interpolate takes a JavaScript template literal and expands outputs correctly:
export const s3Url2: pulumi.Output<string> = pulumi.interpolate`s3://${bucket.bucket}/${file.key}`;
```

Any reason _not_ to do this? I just think it'd be better all 'round. What do you think?

toriancrane: Hi @cnunciato ,

Please see my comment in [this thread](https://pulumi.slack.com/archives/C026M35HVDJ/p1709534734659279?thread_ts=1709074658.591829&cid=C026M35HVDJ). Until I can get example code across all languages that accurately reflects what the experience of interacting with a resource is like, the example output is the best I can do in the interim. I don't think this should be something that blocks as we can always add the interactive code later.

toriancrane: I was kind of hoping to avoid reinventing the wheel on examples where possible, especially given that it can be challenging recreating the same functionality across the different languages in a way that the code actually works. I will work on updating with the above but it will definitely extend my timeline on wrapping this up.

themes/default/content/docs/concepts/inputs-outputs/all.md
```diff
@@ -165,7 +165,7 @@ Server=tcp:myDbServer.database.windows.net;initial catalog=myExampleDatabase;
 
 There is an easier way to generate a concatenated string value using multiple outputs, and that is by using interpolation. Pulumi exposes interpolation helpers that enables you to create strings that contain outputs. These interpolation methods wrap [all](/docs/concepts/inputs-outputs/all/) and [apply](/docs/concepts/inputs-outputs/apply/) with an interface that resembles your language's native string formatting functions. The below example demonstrates how to create a URL from the hostname and port output values of a web server.
```
cnunciato, line 166:
```suggestion
There is an easier way to generate a concatenated string value using multiple outputs, and that is by using interpolation. Pulumi exposes interpolation helpers that enables you to create strings that contain outputs. These interpolation methods wrap [all](/docs/concepts/inputs-outputs/all/) and [apply](/docs/concepts/inputs-outputs/apply/) with an interface that resembles your language's native string formatting functions. The example below demonstrates how to create a URL from the hostname and port output values of a web server.
```

themes/default/content/docs/concepts/inputs-outputs/apply.md
```diff
@@ -1120,555 +1059,64 @@ The result of the call to {{< pulumi-apply >}} is a new Output<T>, meaning the `
...
-            Action: "s3:GetObject",
-            Resource: `arn:aws:s3:::${bucketName}/*`,
-          },
-        ],
-      })
-    ),
-});
-```
+- take native JSON objects and convert them into JSON strings (i.e. serialization)
+- take JSON strings and convert them into native JSON objects (i.e. deserialization)
```
cnunciato, line 1071:
```suggestion
- convert native objects into JSON strings (known as serialization)
- convert JSON strings into native objects (known as deserialization)
```

themes/default/content/docs/concepts/inputs-outputs/apply.md
```diff
@@ -1120,555 +1059,64 @@ The result of the call to {{< pulumi-apply >}} is a new Output<T>, meaning the `
...
+- be converted into a JSON string representation
 
-{{% notes type="info" %}}
-The Pulumi Go SDK does not currently support serializing or deserializing maps with unknown values.
-It is tracked [here](https://github.com/pulumi/pulumi/issues/12460).
+you can easily do so using Pulumi's JSON stringify helper. This helper function unwraps Pulumi outputs without needing to explicitly use `apply` and converts the entire JSON object into a JSON string.
 
-The following is a simplified example of using `pulumi.JSONMarshal` in Go.
-{{% /notes %}}
+For example, you can write the definition of a AWS Step Function State Machine as a native JSON object, embed outputs from other resources (such as a Lambda Function ARN) within the JSON object, and then convert the entire definition into the JSON string representation that is required by the State Machine resource definition:
```
cnunciato, line 1082:
```suggestion
For example, you can write the definition of an AWS Step Function State Machine as a native JSON object, embed outputs from other resources (such as a Lambda Function ARN) within the JSON object, and then convert the entire definition into the JSON string representation that is required by the State Machine resource definition:
```

cnunciato (approved):
Several suggestions, but leaving an approval to keep you unblocked. 🙌

themes/default/static/programs/aws-s3websitebucket-oai-bucketpolicy-go/main.go
```diff
@@ -0,0 +1,70 @@
...
+					}
+					return pulumi.String(policy).ToStringOutput(), nil
+				}).(pulumi.StringOutput),
+		}, pulumi.Parent(bucket))
+		if err != nil {
+			return err
+		}
+		return nil
+	})
+}
```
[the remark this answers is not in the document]
toriancrane: @cnunciato Isn't the JSON Marshall reference on line 44?

```
policy, err := json.Marshal(map[string]interface{}
```

themes/default/content/docs/concepts/inputs-outputs/apply.md
```diff
@@ -952,46 +927,10 @@ resources:
...
-```
-
-{{% /choosable %}}
-
-{{< /chooser >}}
-
 ## Creating new output values
 
+### Outputs and Strings
+
```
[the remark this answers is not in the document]
toriancrane: Outputs and strings was actually a part of the original content and this header was just something I forgot to re-add from the previous iteration. I can update the text to better reflect that it is about strings, but I'd like to keep the header.

themes/default/content/docs/concepts/inputs-outputs/apply.md
```diff
@@ -1120,555 +1059,64 @@ The result of the call to {{< pulumi-apply >}} is a new Output<T>, meaning the `
...
-        ],
-      })
-    ),
-});
-```
+- take native JSON objects and convert them into JSON strings (i.e. serialization)
+- take JSON strings and convert them into native JSON objects (i.e. deserialization)
 
-{{% /choosable %}}
+### Converting JSON object to string
```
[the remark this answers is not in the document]
toriancrane: I actually found the previous headers to be very confusing which is why I didn't initially understand what they were trying to explain. Maybe I can make these level 4 headers to fall under Outputs and JSON to make that connection more clear? The downside is that we'll lose them in the sidebar.

themes/default/content/docs/concepts/inputs-outputs/apply.md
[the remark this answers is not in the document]
toriancrane: Yeah I was worried about the same thing (primarily why I asked if the example was too long in the docs daily). I will get this updated.

PR #4034, opened 2024-03-08
Temporarily redirect some deleted content

cnunciato:
We deleted a number of outdated pages recently, but several are still being linked to from external sources (email, etc.). This change adds a few redirects to handle those while we await a more complete restoration and proper de-listing of the pages that were removed.

Fixes https://github.com/pulumi/marketing/issues/983.


sean1588 (approved):
Approving assuming this works, but dumb question, do these need to be nested in a folder under /resources, eg. `/resources/exploring-the-intersection-of-cluster-api-and-infrastructure-as-code/index.md`? If not, ignore me and ship.

cnunciato:
@sean1588 Nope, this is good. I did them this way just to have them stand out a little, too.

PR #4035, opened 2024-03-09
Fix ownershipControls definition and add fmt to go imports

mvadillo-evi:
## Description
While I was following the guide I stumble upon the following error:
```
./main.go:47:8: undefined: ownershipControls
./main.go:54:16: undefined: fmt
```

~~I was able to successfully create the infrastructure by adding the missing fmt library and removing the ownershipControls dependency.~~ 
- This could be fixed by adding the missing fmt library and set correctly the ownershipControls variable definition. 
## Checklist:

- [ ] If blogging, I have reviewed the [blogging guide](https://github.com/pulumi/pulumi-hugo/blob/master/BLOGGING.md). N/A
- [ ] I have manually confirmed that all new links work. N/A
- [ ] I added aliases (i.e., redirects) for all filename changes. N/A
- [ ] If making css changes, I rebuilt the bundle. N/A

### DIFF
--- a/themes/default/content/docs/clouds/aws/get-started/deploy-changes.md
+++ b/themes/default/content/docs/clouds/aws/get-started/deploy-changes.md
@@ -276,7 +276,13 @@ bucket_object = s3.BucketObject(
 {{% choosable language go %}}
 
 ```go
-_, err = s3.NewBucketOwnershipControls(ctx, "ownership-controls", &s3.BucketOwnershipControlsArgs{
+import (
+	"fmt"
+	"github.com/pulumi/pulumi-aws/sdk/v6/go/aws/s3"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+ownershipControls, err := s3.NewBucketOwnershipControls(ctx, "ownership-controls", &s3.BucketOwnershipControlsArgs{
     Bucket: bucket.ID(),
     Rule: &s3.BucketOwnershipControlsRuleArgs{
         ObjectOwnership: pulumi.String("ObjectWriter"),
@@ -301,6 +307,7 @@ _, err = s3.NewBucketObject(ctx, "index.html", &s3.BucketObjectArgs{
     Acl:         pulumi.String("public-read"),
 }, pulumi.DependsOn([]pulumi.Resource{
 			publicAccessBlock,
+			ownershipControls,
 }))
 if err != nil {
     return err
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/clouds/aws/get-started/deploy-changes.md
```diff
@@ -301,7 +301,6 @@ _, err = s3.NewBucketObject(ctx, "index.html", &s3.BucketObjectArgs{
     Acl:         pulumi.String("public-read"),
 }, pulumi.DependsOn([]pulumi.Resource{
 			publicAccessBlock,
-			ownershipControls,
```
cnunciato, line 304: This line actually does need to be there, I think. I'm guessing the issue you saw was instead because of [this line](https://github.com/pulumi/pulumi-hugo/pull/4035/files#diff-a06a85181197ce448e17d363732f21ac16ccdd16ca05156e7329cb467091bcb0R279), which should be defining that variable, but is instead leaving it ignored (as `_`). If you change that line to this:

```
ownershipControls, err := s3.NewBucketOwnershipControls(ctx, "ownership-controls", &s3.BucketOwnershipControlsArgs{
``` 

... it should be good. (At least that's what our automated tests are telling me.) GitHub won't let me make that suggestion directly, though, so if you wouldn't mind confirming that and changing it up, I'd appreciate it!

themes/default/content/docs/clouds/aws/get-started/deploy-changes.md
```diff
@@ -450,6 +449,12 @@ pulumi.export('bucket_endpoint', pulumi.Output.concat('http://', bucket.website_
 {{% choosable language go %}}
 
 ```go
+import (
+  "fmt"
+  "github.com/pulumi/pulumi-aws/sdk/v6/go/aws/s3"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
```
cnunciato, line 457: Ah, good catch. I see with the other languages, we add these imports in the block just above this one. I wonder if we could do that there instead (e.g., as we do with the Java example)? 🤔

cnunciato (changes requested):
Thanks for submitting this, @mvadillo-evi! Bit of initial feedback for you.

themes/default/content/docs/clouds/aws/get-started/deploy-changes.md
mvadillo-evi: Mmm, I see, the problem I was facing was the `=` instead of `:=` First I tried to do what you mentioned but I didn't see that. Thanks for your great feedback 🙂

themes/default/content/docs/clouds/aws/get-started/deploy-changes.md
mvadillo-evi: Thank you, you are right, the stile doesn’t align with the rest of the languages. 🙂

cnunciato (approved):
Looks great -- thank you again for the contribution!

PR #4042, opened 2024-03-11
update aliases to workshops list to redirect all deleted resource urls

shughes26:
## Description
Adding aliases for all deleted workshop URL so they redirect to the full list (more restoration to come for select workshops, but for now this fixes 404s)


cnunciato (approved):
Thanks! Then I guess once we know which ones we want to "undelete", we can remove them from this list as we restore them.

PR #4044, opened 2024-03-12
Add new transforms docs

Frassle:
## Checklist:


themes/default/content/docs/concepts/options/parent.md
```diff
@@ -113,3 +113,5 @@ Child resources inherit default values for many other resource options from thei
...
 
 * [`transformations`](/docs/concepts/options/transformations):  Transformations applied to a parent will run on the parent and on all child resources. This allows a transformation to be applied to a component to intercept and modify any resources created by it's children. As a special case, [Stack transformations](/docs/concepts/options/transformations#stack-transformations) will be applied to *all* resources (since all resources ultimately are parented directly or indirectly by the root stack resource).
+
+* [`transforms`](/docs/concepts/options/transforms):  Transforms applied to a parent will run on the parent and on all child resources. This allows a transform to be applied to a component to intercept and modify any resources created by it's children. As a special case, [Stack transforms](/docs/concepts/options/transforms#stack-transforms) will be applied to *all* resources (since all resources ultimately are parented directly or indirectly by the root stack resource).
```
ericrudder, line 117:
```suggestion
* [`transforms`](/docs/concepts/options/transforms):  Transforms applied to a parent will run on the parent and on all child resources. This allows a transform to be applied to a component to intercept and modify any resources created by its children. As a special case, [Stack transforms](/docs/concepts/options/transforms#stack-transforms) will be applied to *all* resources (since all resources ultimately are parented directly or indirectly by the root stack resource).
```

Frassle: Thanks, couple of other instances of this mistake on the page as well. Will get them fixed up with.

themes/default/content/docs/concepts/options/parent.md
```diff
@@ -106,10 +106,12 @@ Previewing update (dev):
...
+* [`transformations`](/docs/concepts/options/transformations):  Transformations applied to a parent will run on the parent and on all child resources. This allows a transformation to be applied to a component to intercept and modify any resources created by its children. As a special case, [Stack transformations](/docs/concepts/options/transformations#stack-transformations) will be applied to *all* resources (since all resources ultimately are parented directly or indirectly by the root stack resource).
+
+* [`transforms`](/docs/concepts/options/transforms):  Transforms applied to a parent will run on the parent and on all child resources. This allows a transform to be applied to a component to intercept and modify any resources created by its children. As a special case, [Stack transforms](/docs/concepts/options/transforms#stack-transforms) will be applied to *all* resources (since all resources ultimately are parented directly or indirectly by the root stack resource).
```
tgummerer, line 117: Should this call out in a sentence on how `transforms` are different than `transformations`? Just reading this page I wouldn't be able to tell one from another and why I would go deeper into one or another.

themes/default/content/docs/concepts/options/transforms.md
```diff
@@ -0,0 +1,329 @@
...
+- /docs/intro/concepts/resources/options/transforms/
+---
+
+The `transforms` resource option provides a list of transforms to apply to a resource and all of its children. This option is used to override or modify the inputs to the child resources of a component resource. One example is to use the option to add other resource options (such as `ignoreChanges` or `protect`). Another example is to modify an input property (such as adding to tags or changing a property that is not directly configurable).
+
+Each transform is a callback that gets invoked by the Pulumi engine. It receives the resource type, name, input properties, and resource options. The callback returns a new set of resource input properties and resource options that will be used to construct the resource instead of the original values.
+
+## Transforms vs Transformations
+
+The `transforms` option is the replacement system to [transformations](/docs/concepts/options/transformations/). Calling transform functions is managed by the engine, rather than by each SDK and so transforms apply to resources created inside remote component resources (such as those in [AWSX](https://www.pulumi.com/registry/packages/awsx/)).
```
tgummerer, line 22: Maybe a sentence that compared to Transformations this works with MLCs would be helpful here?

themes/default/content/docs/concepts/options/transforms.md
```diff
@@ -0,0 +1,329 @@
...
+
+However because `transforms` are invoked by the engine the properties are the values after being serialised and deserialised across the wire protocol to the engine and back to the SDK runtime. As such unlike with `transformations` there is no `Resource` object, and the properties will be raw property maps not typed resource arg classes.
+
+The new transform system is also natively async. As such depending on the language your using you may have to lift your transform function to be `async` and/or to handle an extra parameter for the async context (e.g. a `context.Context` in Go, or `System.Threading.CancellationToken` in C#).
+
+As such when moving from `transformations` to `transforms` you will need to update your transform code to handle these differences.
+
+### No resource type
+
+There is no `Resource` object passed to transform functions. But most of the information you could have retrieved from that object is presented on the transform arguments directly.
```
tgummerer, line 32:
```suggestion
There is no `Resource` object passed to transform functions. Most of the information you could have retrieved from that object is presented on the transform arguments directly.
```

themes/default/content/docs/concepts/options/transforms.md
```diff
@@ -0,0 +1,329 @@
...
+    {
+        var resultOpts = CustomResourceOptions.Merge((CustomResourceOptions)args.Options,
+            new CustomResourceOptions {AdditionalSecretOutputs = {"length"}});
+        return new ResourceTransformationResult(resultArgs, resultOpts);
+    }
+    return null;
+}
+```
+
+We can rewrite this quite easily to:
```
tgummerer, line 49:
```suggestion
We can rewrite this to:
```

.mono/registry/CurrentUser/software/microsoft/csdevkit/v1/pids/12603_133547972322250879/values.xml
```diff
@@ -0,0 +1,3 @@
+﻿<values>
+  <value name="StillAlive" type="qword">133554269402339649</value>
+</values>
```
lukehoban, line 3: Unintentionally added file?

themes/default/content/docs/concepts/options/_index.md
```diff
@@ -29,4 +29,5 @@ All resource constructors accept an options argument that provide the following
 - [replaceOnChanges](/docs/concepts/options/replaceonchanges/): declare that changes to certain properties should be treated as forcing a replacement.
 - [retainOnDelete](/docs/concepts/options/retainondelete/): if true the resource will be retained in the backing cloud provider during a Pulumi delete operation.
 - [transformations](/docs/concepts/options/transformations/): dynamically transform a resource’s properties on the fly.
```
lukehoban, line 31: Do we need a comment at this level to help folks understand which of these to use?

Something like this?

```suggestion
- [transformations](/docs/concepts/options/transformations/): dynamically transform a resource’s properties on the fly.  Prefer `transforms` if possible. `transformations` will be deprecated in the future in favor of `transforms`. 
```

themes/default/content/docs/concepts/options/parent.md
lukehoban: Yes - we must help users with this. Left a comment on the previous place we have this issue - could borrow something similar here. 

Honestly - probably easiest for the docs to just say it’s deprecated, even if we’re not yet issues deprecation warnings in the product.

themes/default/content/docs/concepts/options/transforms.md
```diff
@@ -0,0 +1,329 @@
...
+
+## Transforms vs Transformations
+
+The `transforms` option is the replacement system to [transformations](/docs/concepts/options/transformations/). Calling transform functions is managed by the engine, rather than by each SDK and so transforms apply to resources created inside remote component resources (such as those in [AWSX](https://www.pulumi.com/registry/packages/awsx/)).
+
+However because `transforms` are invoked by the engine the properties are the values after being serialised and deserialised across the wire protocol to the engine and back to the SDK runtime. As such unlike with `transformations` there is no `Resource` object, and the properties will be raw property maps not typed resource arg classes.
+
+The new transform system is also natively async. As such depending on the language your using you may have to lift your transform function to be `async` and/or to handle an extra parameter for the async context (e.g. a `context.Context` in Go, or `System.Threading.CancellationToken` in C#).
```
lukehoban, line 26:
```suggestion
The new transform system is also natively async. As such, depending on the language you are using, you may have to lift your transform function to be `async` and/or to handle an extra parameter for the async context (e.g. a `context.Context` in Go, or `System.Threading.CancellationToken` in C#).
```

themes/default/content/docs/concepts/options/transforms.md
```diff
@@ -0,0 +1,329 @@
...
+    weight: 15
+aliases:
+- /docs/intro/concepts/resources/options/transforms/
+---
+
+The `transforms` resource option provides a list of transforms to apply to a resource and all of its children. This option is used to override or modify the inputs to the child resources of a component resource. One example is to use the option to add other resource options (such as `ignoreChanges` or `protect`). Another example is to modify an input property (such as adding to tags or changing a property that is not directly configurable).
+
+Each transform is a callback that gets invoked by the Pulumi engine. It receives the resource type, name, input properties, and resource options. The callback returns a new set of resource input properties and resource options that will be used to construct the resource instead of the original values.
+
+## Transforms vs Transformations
```
lukehoban, line 20: I suggest we split this into two parts:
1. A list of differences in capabilities
2. A migration guide

For the migration guide, I’d love to somehow move it out of the mainline of this page - but not sure how.  I almost wonder if we should put it in the `transformations` page as part of mentioning that that is deprecated, and then cross link it from here?

themes/default/content/docs/concepts/options/transforms.md
```diff
@@ -0,0 +1,329 @@
...
+{
+    if (args.Resource.GetResourceType() == "random:index/randomString:RandomString")
+    {
+        var resultOpts = CustomResourceOptions.Merge((CustomResourceOptions)args.Options,
+            new CustomResourceOptions {AdditionalSecretOutputs = {"length"}});
+        return new ResourceTransformationResult(resultArgs, resultOpts);
+    }
+    return null;
+}
+```
```
lukehoban, line 47: We’ll need these code snippets in all languages.

themes/default/content/docs/concepts/options/parent.md
lukehoban: We’ll also need to add a note into the transformations page.

themes/default/content/docs/concepts/options/transforms.md
Frassle: This one only really makes sense for C#. The other languages don't have `GetResourceType` like this.

PR #4049, opened 2024-03-13
Add a note about checking encryption salts into source control

cnunciato:
Adds a note to clarify that checking these in is also safe and recommented. Also tidies up a bit of formatting and adds a few keywords to make this easier to find. ([Internal thread for reference](https://pulumi.slack.com/archives/C011EMDQ8JE/p1710371997870229).)


themes/default/content/docs/concepts/secrets.md
```diff
@@ -95,8 +95,8 @@ As an example, let’s create an AWS Parameter Store secure value. Parameter Sto
 ```javascript
 const cfg = new pulumi.Config()
 const param = new aws.ssm.Parameter("a-secret-param", {
-    type: "SecureString",
```
cnunciato, line 98: Gah, these removals are unintended -- I'll fix.

toriancrane (approved):
LGTM

PR #4050, opened 2024-03-14
Add Nx monorepo blog

julienp:
## Description

Adds a blog entry about using Nx to manage a monorepo with resource components.

## Checklist:


themes/default/content/blog/nx-monorepo/index.md
```diff
@@ -0,0 +1,270 @@
...
+Do you want to perform this update? yes
+Updating (dev)
+
+...
+NX   Successfully ran target deploy for project infra and 4 tasks it depends on (27s)
+```
+
+The output tells us that Nx found the 4 dependant tasks for the `deploy` task of the `infra` package and ran them successfully.
+
+If we now run `npx nx deploy infra` again we can see Nx's caching mechanism in action. Nx notices that there are no code changes to the `infra` package, and that none of its dependencies have changed either, so it just replays the output from the previous run.
```
interurban, line 191:
```suggestion
If we now run `npx nx deploy infra` again, we can see Nx's caching mechanism in action. Nx notices that there are no code changes to the `infra` package, and that none of its dependencies have changed either, so it just replays the output from the previous run.
```

interurban (approved):
Thx Julien, I also approved Luke's additions.

PR #4058, opened 2024-03-15
blocking and async in python

PollRobots:
Add some guidance for using blocking and async code in python pulumi programs

## Checklist:

### DIFF
--- a/themes/default/content/docs/languages-sdks/python/python-blocking-async.md
+++ b/themes/default/content/docs/languages-sdks/python/python-blocking-async.md
@@ -1,0 +1,199 @@
+---
+title_tag: "Blocking and Async Python with Pulumi | Languages & SDKs"
+meta_desc: An overview of how to use Python with Pulumi for infrastructure as code on any cloud (AWS, Azure, Google Cloud, Kubernetes, etc.).
+title: Blocking and Async
+h1: Blocking and Async Python with Pulumi
+meta_image: /images/docs/meta-images/docs-meta.png
+menu:
+  languages:
+    parent: python
+    weight: 5
+aliases:
+- /python/async/
+---
+
+In large and complex Python Pulumi programs, blocking calls can prevent progress
+from being made by the Pulumi runtime by preventing the event loop from
+processing asynchronous tasks.  
+
+If the output from the blocking code will be used as an input to a resource,
+then the blocking code may be executed on another thread with the future result
+captured as a `pulumi.Output` object. In the same way, asynchronous code (using
+Python’s [`asyncio`](https://docs.python.org/3/library/asyncio.html)) can also
+be evaluated concurrently with the Pulumi runtime and the future result also
+captured using `pulumi.Output`.
+
+## Background
+
+As covered in [Concepts](/docs/intro/concepts), when a Pulumi program is run it creates resources and the dependencies or
+connections between them. Consider the following example (taken from the Pulumi
+blog article [Programming the Cloud with Python](/blog/programming-the-cloud-with-python))
+
+```python
+import mimetypes
+import os
+
+from pulumi import export, FileAsset
+from pulumi_aws import s3
+
+web_bucket = s3.Bucket('s3-website-bucket', website={
+    "index_document": "index.html"
+})
+
+content_dir = "www"
+for file in os.listdir(content_dir):
+    filepath = os.path.join(content_dir, file)
+    mime_type, _ = mimetypes.guess_type(filepath)
+    obj = s3.BucketObject(file,
+        bucket=web_bucket.id,
+        source=FileAsset(filepath),
+        content_type=mime_type)
+```
+
+Here the resources are an S3 Bucket and some number of S3 BucketObjects---one
+for each file in the `www` directory. The bucket objects are told which bucket
+they will be put in using the id of the bucket. This establishes the dependency
+between the BucketObject and the Bucket.
+
+The type of the bucket id is `Output[str]`. The
+[`pulumi.Output`](/docs/reference/pkg/python/pulumi/#pulumi.Output) type holds a
+piece of data, in this case the id of the bucket, this is not necessarily known
+during the execution of the program, and is held internally within the output
+object as a future---the awaitable result of an asynchronous operation.
+
+In this way the Pulumi program can proceed and build the graph of resources
+without having to wait for each resource in turn. When Pulumi is previewing a
+stack the value of an output may never be known. When resources are then created
+that can be done concurrently according to the dependency graph that the Pulumi
+program defines.
+
+## Blocking and Asynchronous Code
+
+The example above, and many or most Python Pulumi programs, contains no explicit
+asynchronous code. While the resources are connected using futures, the
+asynchronous evaluation all happens behind the scenes.
+
+This can cause problems both with blocking code and with explicitly asynchronous code
+
+- Blocking code will prevent the event loop on the Pulumi program’s thread from
+  pumping while the blocking code is executing
+- Asynchronous code may not be executed unless it is explicitly scheduled. It is
+  not appropriate to call `asyncio.run` from within a Pulumi program
+
+### Blocking Code
+
+A Pulumi program is a single threaded Python program. This means that the
+asynchronous operations on resources cannot happen while blocking code is
+evaluating. For simple Pulumi programs this is not consequential, but for more
+complex programs it is possible to be waiting on blocking code that causes all
+progress to be stalled waiting for that blocking code to complete.
+
+Examples of this might include calls to requests.get, or subprocess.run that are
+used to collect information needed to configure the Pulumi program or provide
+input from external sources.
+
+If the result from the blocking code is used to provide the input for a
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/languages-sdks/python/python-blocking-async.md
```diff
@@ -0,0 +1,199 @@
...
+  languages:
+    parent: python
+    weight: 5
+aliases:
+- /python/async/
+---
+
+In large and complex Python Pulumi programs, blocking calls can prevent progress
+from being made by the Pulumi runtime by preventing the event loop from
+processing asynchronous tasks.
```
AaronFriel, line 17: Should we start this with a brief description of how Pulumi Python programs rely on an event loop to handle resource declarations, outputs and applies, and more asynchronously? Then this sentence follows naturally.

Suggest nixing "In large and complex ..." and "may", I think we can be prescriptive: blocking calls block the Pulumi runtime.

themes/default/content/docs/languages-sdks/python/python-blocking-async.md
```diff
@@ -0,0 +1,199 @@
...
+In large and complex Python Pulumi programs, blocking calls can prevent progress
+from being made by the Pulumi runtime by preventing the event loop from
+processing asynchronous tasks.  
+
+If the output from the blocking code will be used as an input to a resource,
+then the blocking code may be executed on another thread with the future result
+captured as a `pulumi.Output` object. In the same way, asynchronous code (using
+Python’s [`asyncio`](https://docs.python.org/3/library/asyncio.html)) can also
+be evaluated concurrently with the Pulumi runtime and the future result also
+captured using `pulumi.Output`.
```
AaronFriel, line 24: This graf seems like it follows logically from the rest of the document, but read on its own applies that this is somehow done for the user. Our are often written as though the Pulumi engine or runtime is the subject, so someone could read "the blocking code may be executed on another thread" as though the Pulumi runtime is doing that for them or there's some heuristic involved. 

Perhaps "Blocking could _should_ be executed on another thread"?

Or using a more active voice:

> The Pulumi runtime enables you to await synchronous and asnychronous calls without blocking the event loop. ... [describe that for synchronous calls, we use `asyncio.to_thread` to make synchronous calls async] [describe that for both, we then use `pulumi.Output.from_input`] ...

themes/default/content/docs/languages-sdks/python/python-blocking-async.md
```diff
@@ -0,0 +1,199 @@
...
+The example above, and many or most Python Pulumi programs, contains no explicit
+asynchronous code. While the resources are connected using futures, the
+asynchronous evaluation all happens behind the scenes.
+
+This can cause problems both with blocking code and with explicitly asynchronous code
+
+- Blocking code will prevent the event loop on the Pulumi program’s thread from
+  pumping while the blocking code is executing
+- Asynchronous code may not be executed unless it is explicitly scheduled. It is
+  not appropriate to call `asyncio.run` from within a Pulumi program
```
AaronFriel, line 81: What happens?

themes/default/content/docs/languages-sdks/python/python-blocking-async.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Blocking Code
+
+A Pulumi program is a single threaded Python program. This means that the
+asynchronous operations on resources cannot happen while blocking code is
+evaluating. For simple Pulumi programs this is not consequential, but for more
+complex programs it is possible to be waiting on blocking code that causes all
+progress to be stalled waiting for that blocking code to complete.
+
+Examples of this might include calls to requests.get, or subprocess.run that are
```
AaronFriel, line 91:
```suggestion
Examples of this might include calls to `requests.get`, or `subprocess.run` that are
```

themes/default/content/docs/languages-sdks/python/python-blocking-async.md
```diff
@@ -0,0 +1,199 @@
...
+    As example function that uses the Command package to run a local process
+    """
+    my_cli = local.Command('my-cli', create=f'my-cli {shlex.quote(foo)})')
+    return my_cli.stdout
+
+# The return from call_my_cli is already a pulumi.Output
+mycli_output = call_my_cli('bar')
+```
+
+If the input to the function needs to be a `pulumi.Output` itself---_i.e._ if it is the result of creating another resource, or calling another command, then we can use [`pulumi.Output.concat`](/docs/reference/pkg/python/pulumi#pulumi.Output.concat) as follows
```
AaronFriel, line 179: nit: is it normal for us to go into a code block without a colon?

themes/default/content/docs/languages-sdks/python/python-blocking-async.md
```diff
@@ -0,0 +1,199 @@
+---
+title_tag: "Blocking and Async Python with Pulumi | Languages & SDKs"
```
AaronFriel, line 2: Could we add a note to the Python page linking to this?

themes/default/content/docs/languages-sdks/python/python-blocking-async.md
PollRobots: It will raise an exception because there is already an event loop

themes/default/content/docs/languages-sdks/python/python-blocking-async.md
PollRobots: Yeah, these are good points. I mentioned large and complex, because in general it is a nothing burger for a small program, but when the stack is super complicated its much harder to tell where the slowdown is coming from.

I'll reword it

themes/default/content/docs/languages-sdks/python/python-blocking-async.md
PollRobots: I'll add colons

themes/default/content/docs/languages-sdks/python/python-blocking-async.md
PollRobots: Done

themes/default/content/docs/languages-sdks/python/python-blocking-async.md
PollRobots: Updated the wording here.

themes/default/content/docs/languages-sdks/python/python-blocking-async.md
PollRobots: Thanks! I reworded this

themes/default/content/docs/languages-sdks/python/python-blocking-async.md
PollRobots: Done

themes/default/content/docs/languages-sdks/python/python-blocking-async.md
```diff
@@ -0,0 +1,206 @@
...
+[`asyncio`](https://docs.python.org/3/library/asyncio.html)) event loop to
+create the resources that the Pulumi program describes---this isn't directly
+visible to the Python program.
+
+A consequence of this is that blocking calls prevent progress from being made by
+the Pulumi runtime by preventing the event loop from processing asynchronous
+tasks.  
+
+The `pulumi.Output` object can be used to pass the output from both synchronous
+and asynchronouse calls without blocking the event loop.
```
Frassle, line 25:
```suggestion
and asynchronous calls without blocking the event loop.
```

themes/default/content/docs/languages-sdks/python/python-blocking-async.md
```diff
@@ -0,0 +1,206 @@
...
+
+For synchronous or blocking code, this should be executued on another thread
```
Frassle, line 27:
```suggestion
For synchronous or blocking code, this should be executed on another thread
```

themes/default/content/docs/languages-sdks/python/python-blocking-async.md
PollRobots: aargh, I always make this one typo! It makes me think of a mouse

themes/default/content/docs/languages-sdks/python/python-blocking-async.md
```diff
@@ -12,22 +12,29 @@ aliases:
...
+tasks.  
+
+The `pulumi.Output` object can be used to pass the output from both synchronous
+and asynchronous calls without blocking the event loop.
+
+For synchronous or blocking code, this should be executed on another thread
+with the future result captured as a `pulumi.Output`---using `asyncio.to_thread`
+and `pulumi.Output.from_input` respectively.
+
+For asynchronous code the coroutine or other awaitable resilt can also be passed to `pulumi.Output.from_input` to create a `pulumi.Output` object.
```
AaronFriel, line 31:
```suggestion
For asynchronous code the coroutine or other awaitable result can also be passed to `pulumi.Output.from_input` to create a `pulumi.Output` object.
```

AaronFriel (commented):
LGTM!

AaronFriel (approved):

Frassle (approved):

PR #4059, opened 2024-03-15
attempting to fix redirect build error

sean1588:
There is a [build error due to redirects](https://github.com/pulumi/docs/actions/runs/8300711991/job/22719219189?pr=11039), so attempting to resolve that here. I suspect it is the anchor at the end of the url that may be causing this, but not sure. 

cc @toriancrane


cnunciato:
@sean1588 I think we can close this now, so I'll go ahead and do that. Thanks for hopping in to help with it though!

PR #4073, opened 2024-03-20
Remove broken link to community-engineering page.

sean1588:
## Description

The docs build is failing due to this broken relref link that is trying to link to `/community/community-engineering`. This appears to be a top level page that I'm guessing would list out all the community engineers. This page seems to no longer exist. So I'm removing the link for now, to unblock the build  and we can decide if we want to bring that page back later and then we can add back the link.


cnunciato (approved):
Thanks Sean!

PR #4079, opened 2024-03-21
Fixes a small code issue to resolve test pipeline error

toriancrane:
## Description

Resolves [this issue](https://github.com/pulumi/pulumi-hugo/issues/4078).


cnunciato (changes requested):
The change looks okay, but we should remove these go.* files.

cnunciato (approved):

PR #4082, opened 2024-03-23
Pull older AWS guides into a More page under AWS > Guides

### LINKED ISSUE #3106: No way to navigate to our older AWS Guides
Pages like these still rank highly in search results (external as well as internal), but there's no way to navigate them directly in the UI:

image

https://www.pulumi.com/docs/clouds/aws/aws-guides/ec2/

Following the breadcrumb to the "top level" of this section (by clicking "AWS Guides") also leads to this page, which appears to have no descendants and also contains @pulumi/cloud content that we may or may not want to continue featuring in this way. 

We should decide on a path forward for these pages and get their content integrated properly, either into existing AWS Guides or other tutorials.

cnunciato:
This change pulls the "orphaned" content that currently exists at https://www.pulumi.com/docs/clouds/aws/aws-guides/ into a single page under AWS > Guides. As part of this, I made sure all of these example also worked (some didn't), formatted the code, and added a brief summary for each explaining what it does.

Note this also removes [the `@pulumi/cloud` "multicloud" snippet](https://www.pulumi.com/docs/clouds/aws/aws-guides/). I did this deliberately because we aren't actively developing this library, so I'd rather not point users to it in a prominent location like this.

All redirects should be handled, so we shouldn't see any 404s or loss of link juice with these changes.  

Fixes #3106.


toriancrane (approved):
LGTM

PR #4117, opened 2024-03-28
Update AWS Guide S3 Examples

toriancrane:
## Description

This PR adds examples for the other supported languages to the S3 section.


themes/default/static/programs/aws-s3-bucket-resources-csharp/aws-s3-bucket-resources-csharp.csproj
```diff
@@ -0,0 +1,14 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net7.0</TargetFramework>
```
cnunciato, line 5: Similar to your other PR, we should make this `net6.0`:

```suggestion
    <TargetFramework>net6.0</TargetFramework>
```

themes/default/static/programs/aws-s3-bucket-resources-go/go.mod.txt
```diff
@@ -0,0 +1,10 @@
+module aws-s3-bucket-resources-go
+
+go 1.21
+
+toolchain go1.22.1
```
cnunciato, line 5: I think you may need to make sure you're running the earliest supported versions of these runtimes (Go, .NET, etc.) locally, or at least are able to switch to them easily. (I think this should be <= 1.21.)

cnunciato (commented):
Thanks for doing this! I'll give it a more thorough review tomorrow but wanted to get you a quick response to do something with in the meantime.

themes/default/static/programs/aws-s3-bucket-resources-go/go.mod.txt
toriancrane: I'm not sure that this is true given that the `make test` was failing for certain things. If a user does not have an earlier version of a runtime installed and they run pulumi up, won't that give them the same errors that I've been given?

cnunciato: If the user is a .NET user and doesn't have a runtime installed that aligns with what's specified in the program's `TargetFramework` setting, then yes, they'll get an error (and this may be what you ran into). To get past it, they just install the specified runtime. This happens in both directions; if you're running 6 and the program calls for 7, you'll need to install 7. If you're running 7 and it calls for 6, you'll need to install 6. Since .NET supports running multiple runtimes side-by-side (e.g., `dotnet --list-runtimes` may list several; I currently have 6 and 7), I believe .NET users are generally comfortable navigating this sort of thing.

The other languages are more "minimum-version" oriented, meaning newer runtimes are generally fine running code that was written for older versions. For example with Go, if the minimum Go version (the `go` parameter in a `go.mod` file) is 1.22, and you have 1.21, you won't be able to run it. But if you're on 1.22 and `go.mod` says `1.21`, you generally will. It's similar for Node; you can generally run really old JS code on the latest version of Node, and I think it's the same for Python as well. 

So for our purposes, we should always target the oldest runtimes and use the latest Pulumi SDKs, as this means folks still running on older runtimes can expect the code we show them in the docs to work -- even if the runtime we happen to be targeting in our tests is an older version than the one they currently have. (And if they want to know which runtime we _are_ testing against, they can always click the GitHub icon to find out.)

cnunciato (approved):

PR #4120, opened 2024-03-28
Small update to program code packages/config

toriancrane:
## Description

This PR makes a small update to some of the example program code to resolve pipeline issue.


themes/default/static/programs/aws-lambda-stepfunctions-jsonhelper-csharp/aws-lambda-stepfunctions-jsonhelper-csharp.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net7.0</TargetFramework>
```
cnunciato, line 5: We should actually keep this at 6, as it's still on the list of [supported versions](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core#lifecycle) (and we [officially support all supported versions](https://www.pulumi.com/docs/languages-sdks/dotnet/)). Is there functionality here that doesn't work in .NET 6?

themes/default/static/programs/aws-lambda-stepfunctions-jsonhelper-go/main.go
```diff
@@ -33,9 +30,11 @@ func main() {
...
 			// The following functionality is currently unsupported as myResource
 			// is an unknown value.
-			"unknown": myResource.ApplyT(func(res interface{}) (interface{}, error) {
-				return "Hello World!", nil
-			}),
+			/*
+				"unknown": myResource.ApplyT(func(res interface{}) (interface{}, error) {
+					return "Hello World!", nil
+				}),
+			*/
```
cnunciato, line 37: Why is this commented?

toriancrane: Because the make test was failing at `myResource`

themes/default/static/programs/aws-lambda-stepfunctions-jsonhelper-csharp/aws-lambda-stepfunctions-jsonhelper-csharp.csproj
toriancrane: `make test` was failing for me when this was 6. i think its because i have 7 on mine instance

cnunciato: Yeah I think if you install 6 locally, you'll be good. https://dotnet.microsoft.com/en-us/download/dotnet

themes/default/static/programs/aws-lambda-stepfunctions-jsonhelper-go/main.go
cnunciato: Ok. Do we need that code, then? If not, can we remove it? I'm honestly not sure, as a user, what this commented bit is supposed to be telling me. (Like I personally know that it's telling me `JSONMarshal` doesn't support straight-up `interface{}`, but as a novice user, I'm not at all sure this particular snippet of code would make this clear to me or why it should matter. For example, what is `myResource` here? Is it a Pulumi resource? Because if so, resources don't have `.ApplyT` methods. Is it a property? Why are we returning "Hello World!"? I just find this particular example confusing. 😅)

I'm remembering now [some suggestions I had to use IAM policies](https://github.com/pulumi/pulumi-hugo/pull/4032#discussion_r1529291418) for this helper, and there was a Go example in there that I think worked okay. Should we use that instead, maybe? (I'd also be happy apply that example to all languages if you wanted, as it'd perobably be nice if they were all "doing the same thing". (Right now, they're all doing step-function things except this one.)

toriancrane: This was the code that was originally there. I didn't understand what was going on because I'm not really familiar with this programming language but I assumed it was important. If you have a better example that represents the content of this section (and the associated limitation of this language), please feel free to add.

cnunciato: Ok cool, I'll put something together for you to have a look at, then.

cnunciato: Ok I just pushed a commit that makes the Go example work. Take a look and let me know what you think.

toriancrane: Hi @cnunciato ,

Yes I agree that a simple example would be better, ideally an example that doesn't additionally require the use of `interpolate` and one that will actually deploy when running `pulumi up`. I think that was my main concern where we last left off with this section.

cnunciato: Ok, so I've pushed a commit to your branch here that adds an S3 bucket policy example that's hopefully simpler (and definitely works and deploys). The example shows both a "simple" case of using an output as a JSON property value directly and the "less simple" case of having to use an output to compose a value. I've also adjusted the text to explain this. Let me know what you think.

I've also left the Step Functions examples in the PR since they work -- we might as well keep 'em!

cnunciato (approved):

PR #4137, opened 2024-04-03
Fix invalid example in ESC REST API docs

lblackstone:
The create example included an incorrect `curl` command, which included YAML content in the POST method. The API actually only supports updating the environment content in the PATCH method.

## Checklist:


MMartyn (approved):
lgtm

cnunciato (approved):

PR #4151, opened 2024-04-08
Adds interpolate content to apply page

toriancrane:
## Description

This PR adds interpolate helper content to the Apply page.


cnunciato (changes requested):
This "fixes" the absence of this content on this page, but now we have two pieces of content on two separate pages that duplicates and must be maintained independently. This also leaves us with the related problem of having the JSON helpers on the Apply page, but still not on the All page.

I really do think the proper solution here is just to give the helpers a page of their own, as I've suggested, and link to this page from both the Apply and All pages. There are benefits to this beyond just not having the duplication as well -- discoverability, clarity/distinguishing from Apply and All as I've mentioned, room for more examples, room for more helpers as we add them, etc. I'd much rather see us do this than duplicate this content (and presumably the JSON helper content as well). The effort should minimal since we've already written pretty much all of the content we need already.

PR #4158, opened 2024-04-09
Sarah/filter

shughes26:
## Description
Filters for resources, added filters for skill level, cloud provider, topic, and programming language

Fixes https://github.com/pulumi/marketing/issues/822.

## Checklist:


sean1588:
would there be a way for you to have the groups within each filter only allow one selection, so if you choose aws, it deselects azure? When both are selected it says there are no matches. I guess maybe another option would be to OR the filters selections together rather than AND, so it would show all workshops that are aws or azure.


Screen Shot 2024-04-10 at 2 48 24 PM

cnunciato:
> I guess maybe another option would be to OR the filters selections together rather than AND

This is how it works in the Registry filters, too (i.e., the checkboxes are ORs rather than ANDs).

shughes26:
@sean1588 @cnunciato I'll change the filters from AND to OR

themes/default/theme/src/scss/_event-list.scss
```diff
@@ -59,6 +59,51 @@
             }
         }
     }
+
+    pulumi-filter-select {
+        align-items: end;
+        padding-bottom: 1rem;
```
cnunciato, line 65: Could this be this?

```suggestion
        @apply items-end pb-4;
```

cnunciato (commented):
This looks great! I agree with Sean, though, that we probably want the filters to be ORs rather than ANDs. Otherwise, only a few tiny nits with suggestions.

themes/default/content/resources/getting-started-with-infrastructure-as-code-on-digital-ocean/index.md
```diff
@@ -69,8 +69,9 @@ main:
     # case-sensitive
     tags:
         level: Beginner # Beginner, Intermediate, Advanced
-        topics: ["DigitalOcean"]
+        topics: []
         languages: []
+        clouds: ["DigitalOcean"]
```
cnunciato, line 74: Why is it when I choose this, I get no results? 

image

I get this with others as well, like CodeFresh in the Topics menu, Azure, etc. Is this because they're in the past, and that's preventing them from displaying?

shughes26: The filters are OR'd except for upcoming/on-demand (since the content lives under tabs) so the only DigitalOcean content shows up under the on-demand tab.

cnunciato: Oh -- so the tabs behave like a filter also, then. That seems like a design bug -- makes it look like we don't have content we actually do. 😬

We don't need to block on it, but we should file a follow-up issue for it, as I'm sure someone else will bump into it and get the wrong impression. Unless you can think of an easy solution to include in this PR?

cnunciato (approved):
Approving to unblock, but we should file an issue for the tabs hiding filtered content and link it to this one if we decide to go that route.

themes/default/content/resources/getting-started-with-infrastructure-as-code-on-digital-ocean/index.md
shughes26: I don't really have a solution in mind so maybe we need some UX support? I created a [follow-up issue](https://github.com/pulumi/pulumi-hugo/issues/4175) for this

PR #4164, opened 2024-04-11
Add ESC + Direnv guide

pgavlin:
Add a page that explains how to integrate ESC with Direnv.

### DIFF
--- a/themes/default/config.yml
+++ b/themes/default/config.yml
@@ -1,3 +1,12 @@
 module:
   hugoVersion:
     min: 0.92.0
+
+menus:
+  # This allows the "Integrate With Pulumi IaC" to appear in both the ESC "Getting Started" menu and the
+  # and the "Other Integrations" menu.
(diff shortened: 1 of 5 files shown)


pgavlin:
I'm not entirely sure that this should be part of the getting started, but I didn't have a better idea of where to put it. Maybe we should have an "integrations" section?

pgavlin:
(we should update https://github.com/direnv/direnv/wiki#tools-integration and add a link to this page once it is deployed)

pgavlin:
Okay, this should be ready to go.

These changes now add a new section to the ESC menu, "Other Integrations", which links to the IaC guide and the new Direnv guide.

pgavlin:
PTAL @arunkumar611

arunkumar611 (approved):
Looks good. 

We need to create additional content for Pulumi IaC page and add it there. Experience of clicking on it its pretty jarring.

PR #4173, opened 2024-04-12
fix broken credijusto links

sean1588:
## Description

This PR addresses some broken links, regarding Credijusto. Credijusto is now Covalto, so I updated the website to covalto.com. I think this is good enough, but maybe we should change some of the wording to use the covalto name rather than just updating the link?


interurban (approved):

sean1588:
Going to close this, looks like it was already addressed by Christian
https://github.com/pulumi/pulumi-hugo/pull/4174/files

PR #4206, opened 2024-04-19
Update PulumiUP images

SaraDPH:
## Description

Updating images for PulumiUP

- Banner
- Ad
- Meta

## Checklist:


cnunciato (commented):
I might be in the minority here, but the "fake buttons on a static image" thing always sort of bugs me, to be honest. I can be okay with one button, but two buttons that aren't really buttons and also look like they do different things, but don't (I'm assuming they both point to the same URL), feels like a bit much to me. Is there some way we could communicate both of these ideas in the same image?

SaraDPH:
> I might be in the minority here, but the "fake buttons on a static image" thing always sort of bugs me, to be honest. I can be okay with one button, but two buttons that aren't really buttons and also look like they do different things, but don't (I'm assuming they both point to the same URL), feels like a bit much to me. Is there some way we could communicate both of these ideas in the same image?

It leads to the PulumiUP landing page, where they do have the two buttons. I understand what you mean, but the ad space is so small that it doesn't really make it user-friendly to add "descriptions" or "explanations" to it, but we do want them to know that they can 1) submit a proposal and/or 2) register.

I'm open to suggestions, if you have ideas.

cnunciato (approved):
Approving to unblock the imagery needed for the landing page. I do still think we could do better on this ad, but needn't block the other updates for that, since that won't be visible to users yet.

PR #4210, opened 2024-04-19
update pulumiUp page for 2024 save-the-date

shughes26:
## Description
https://github.com/pulumi/marketing/issues/1029
updated both `/pulumi-up` and `/pulumi-up/thank-you`

I believe we still need some sort of disclaimer that explains the content on the page is from a prior year if anyone knows how we want that to look / what we want that to say


sean1588 (approved):

SaraDPH:
@shughes26 can you please also update the nav bar to 2024, and add AI / ML after IaC:
image

PR #4216, opened 2024-04-23
Fix OpenID and tighten up some language

cleverguy25:
## Description

- Fix OIDC connect to OpenID Connect as the former is wrong
- Simplify some language

### DIFF
--- a/themes/default/content/blog/oidc-trust-relationships/index.md
+++ b/themes/default/content/blog/oidc-trust-relationships/index.md
@@ -75,7 +75,7 @@ In this demo, we are going to use Github Actions to retrieve Pulumi credentials 
 
 * **Sub**: repo:***organization***/***repo***:*
 <!-- markdownlint-enable no-bare-urls -->
-4. Create a GitHub action. Here is a sample code. Ensure to replace the organization name with your organization in the `fetch pulumi token` step.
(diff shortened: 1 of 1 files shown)


lukehoban:
Obviously a nit at this point...

We call this "OIDC" almost everywhere in our product, docs and previous blogs.  On first mention we say "OpenID Connect (OIDC)", but then just "OIDC".    (~900 mentions of "OIDC" and ~100 of "OpenID Connect", the latter of which are mostly in the form "OpenID Connect (OIDC)" or similar)

The updates here make it technically correct, but still seemingly different than the norm in most of our other content.  I think it's still worth bringing this more inline - spell it out on first mention, then use the acronym.

arunkumar611:
Language changes look good. I also made two minor edits. 

As Luke mentions, we'd want OIDC vs. OpenID connect to be consistent. In the first paragraph and in the first instance, we can say "OpenID Connect (OIDC)". Everywhere else we can just say OIDC.

Right now we have ~20 OIDCs and ~10 OpenID Connect in the blog.

arunkumar611:
Do we need to update the meta.png as well? If yes, what text do we want there?

arunkumar611 (approved):
LGTM

PR #4220, opened 2024-04-24
update newsroom

shughes26:
## Description
newsroom and releases page updates


aaronkao:
Let's also add https://info.pulumi.com/announcement/infrastructure-lifecycle-management?hs_preview=vVYbZUFh-165311542814 which is the release tomorrow that will go out at 9am

aaronkao (approved):

PR #4233, opened 2024-04-26
Fix indentation in yaml reference for explicit providers

VenelinMartinov:
## Description
Small fix in the yaml reference - was missing the resources declaration.

## Checklist:

### DIFF
--- a/themes/default/content/docs/languages-sdks/yaml/yaml-language-reference.md
+++ b/themes/default/content/docs/languages-sdks/yaml/yaml-language-reference.md
@@ -135,7 +135,8 @@ resources:
 To create an explicit provider instance, preferably with a specific version, use the [`resources`](#resources) section and prefix the name of the provider with `pulumi:providers` which will the value of the `type` property.
 
 ```yaml
-provider:
+resources:
+  provider:
     type: pulumi:providers:azure
     options:
       version: 5.1.0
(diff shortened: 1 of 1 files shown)


themes/default/content/docs/languages-sdks/yaml/yaml-language-reference.md
```diff
@@ -135,7 +135,8 @@ resources:
 To create an explicit provider instance, preferably with a specific version, use the [`resources`](#resources) section and prefix the name of the provider with `pulumi:providers` which will the value of the `type` property.
```
cnunciato, line 135:
```suggestion
To create an explicit provider instance, preferably with a specific version, use the [`resources`](#resources) section to declare a `provider`, prefixing the name with `pulumi:providers`.
```

cnunciato (approved):
LGTM, but the preceding sentence needs fixing up too. This look okay?

themes/default/content/docs/languages-sdks/yaml/yaml-language-reference.md
VenelinMartinov: Thanks for spotting! I edited it as: 

> For the `type` property, prefix the name of the provider with `pulumi:providers`.

I believe this is slightly more clear but feel free to disagree, happy to go with your original suggestion.

PR #4235, opened 2024-04-26
Logo update for bmw group

interurban:
## Description
Swapping out the BMW Group logo


usrbinkat (approved):
double checked on preview build, LGTM

PR #4269, opened 2024-05-13
Fix CSS background color to be more readable for code listings with highlighted lines

thoward:
This makes a small change to the CSS for code blocks that have highlighted lines. I ran the bundler to update the `bundle.css` file and also changed it in the source `.scss` file.

A code block like this:

```typescript {hl_lines=[2]}
import * as childProcess from "child_process";
```

Will currently render with this background color, which is hard to read:

Screenshot 2024-05-13 at 1 39 19 PM

After this change it will look like this:

Screenshot 2024-05-13 at 1 39 05 PM

### DIFF
--- a/themes/default/theme/src/scss/_chroma.scss
+++ b/themes/default/theme/src/scss/_chroma.scss
@@ -6,7 +6,7 @@
 /* LineLink */ .chroma .lnlinks { outline: none; text-decoration: none; color: inherit }
 /* LineTableTD */ .chroma .lntd { vertical-align: top; padding: 0; margin: 0; border: 0; }
 /* LineTable */ .chroma .lntable { border-spacing: 0; padding: 0; margin: 0; border: 0; }
-/* LineHighlight */ .chroma .hl { background-color: #474747 }
+/* LineHighlight */ .chroma .hl { background-color: #4d3675 }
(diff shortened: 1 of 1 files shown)


cnunciato (approved):
Much better, thanks! ⭐ 

FWIW and for future reference, we also have a color palette with a bunch of grays and other colors that we use to power [TailwindCSS](https://tailwindcss.com/). You'll find the Tailwind config itself here:

https://github.com/pulumi/pulumi-hugo/blob/master/themes/default/theme/tailwind.config.js

... and the tokens that back that config here:

https://github.com/pulumi/pulumi-hugo/blob/master/themes/default/theme/tokens.json

In general we try to keep to only the colors that belong to this palette (for overall consistency), but that needn't block here. If you do see one you think works well, by all means feel free to use it, but up to you. This is definitely an improvement for now. Thanks again!

thoward:
> In general we try to keep to only the colors that belong to this palette (for overall consistency), but that needn't block here. If you do see one you think works well, by all means feel free to use it, but up to you. This is definitely an improvement for now. Thanks again!

Good call! I'll update to use one of the colors defined there

thoward:
> > In general we try to keep to only the colors that belong to this palette (for overall consistency), but that needn't block here. If you do see one you think works well, by all means feel free to use it, but up to you. This is definitely an improvement for now. Thanks again!
> 
> Good call! I'll update to use one of the colors defined there

New color:

Screenshot 2024-05-13 at 3 53 33 PM

PR #4274, opened 2024-05-14
Add python code samples to packages ecosystem blog post

thoward:
## Description

Adds Python code samples to the packages ecosystem post.

## Checklist:

- [X] If blogging, I have reviewed the [blogging guide](https://github.com/pulumi/pulumi-hugo/blob/master/BLOGGING.md).

### DIFF
--- a/themes/default/content/blog/next-level-iac-package-ecosystems/index.md
+++ b/themes/default/content/blog/next-level-iac-package-ecosystems/index.md
@@ -28,7 +28,7 @@ Let’s have a look at a concrete example where this package ecosystem can get y
 
 ## Use Case: Determining S3 MIME types on-the-fly
 
-A popular example of cloud infrastructure automation is creating an S3 bucket and pushing some files to it. This is a common task that many of us have had to do, and most infrastructure-automation tools have a way to accomplish this. 
+A popular example of cloud infrastructure automation is creating an S3 bucket and pushing some files to it. This is a common task that many of us have had to do, and most infrastructure automation tools have a way to accomplish this. 
 
 Here’s our public answer for [how to do that in Pulumi][s3-bucket-example]: 
 
@@ -39,23 +39,19 @@ Here’s our public answer for [how to do that in Pulumi][s3-bucket-example]:
 ```typescript
 import * as pulumi from '@pulumi/pulumi';
 import * as aws from '@pulumi/aws';
-import * as fs from 'fs';
 
 // Create an S3 bucket where we will upload the file.
 const bucket = new aws.s3.Bucket("my-bucket", {
     // You can apply additional bucket settings here.
 });
 
-const filePath = "path/to/your/local/file.ext"; // Replace with the path to the file you want to upload
-
-// Read the contents of the file to be uploaded to S3.
-const fileContents = fs.readFileSync(filePath, {encoding: 'utf-8'});
+const filePath = "path/to/your/local/file.txt"; // Replace with the path to the file you want to upload
 
 // Upload the file to the S3 bucket.
 const bucketObject = new aws.s3.BucketObject("my-bucket-object", {
     bucket: bucket, // Reference to the bucket created above.
-    key: "file.ext", // This will be the file's name in the bucket.
-    content: fileContents, // Contents of the file.
+    key: "file.txt", // This will be the file's name in the bucket.
+    source: pulumi.FileAsset(filePath), // Contents of the file.
     contentType: "text/plain", // MIME-type of the file. Adjust if you upload a different type of file.
 });
 
@@ -78,17 +74,13 @@ import pulumi_aws as aws
 bucket = aws.s3.Bucket('my-bucket')
 
 # The path to the file to upload to S3
-file_path = "path/to/your/local/file.ext"  # Replace with the path to the file you want to upload
-
-# Read the contents of the file to be uploaded to S3.
-with open(file_path, 'rb') as file:
-    file_contents = file.read()
+file_path = "path/to/your/local/file.txt"  # Replace with the path to the file you want to upload
 
 # Upload the file to the S3 bucket
 bucket_object = aws.s3.BucketObject('my-bucket-object',
                                 bucket=bucket.id, # Reference to the bucket created above.
-                                key='file.ext',  # This will be the file's name in the bucket.
-                                content=file_contents,
+                                key='file.txt',  # This will be the file's name in the bucket.
+                                source=pulumi.FileAsset(file_path),
                                 content_type='text/plain')  # MIME-type of the file. Adjust if you upload a different type of file.
 
 # Export the name of the bucket
@@ -122,7 +114,7 @@ content_type='text/plain')  # MIME-type of the file. Adjust if you upload a diff
 
 Yeah, hard-coding the MIME type for every file I need to upload is a not going to scale. In fact, I’m probably going to get a random directory full of files I need to iterate over, each with different file types, and it will change every time it runs, and we won’t know what type of file it is ahead of time.
 
-Packages to the rescue!! Turns out that determining the MIME type of a file in TypeScript only involves importing a community-created [`mime-types`][npm-mimetypes] [library][python-mimetypes] and a single function call to guess the mimetype:
+Packages to the rescue!! Turns out that determining the MIME type of a file only involves importing a community-created module like [`mime-types`][npm-mimetypes] in TypeScript, or using the [mimetypes][python-mimetypes] module from Python's standard library, and a single function call to guess the MIME type:
 
 {{< chooser language "typescript,python" />}}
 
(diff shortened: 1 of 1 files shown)


themes/default/content/blog/next-level-iac-package-ecosystems/index.md
```diff
@@ -60,26 +64,95 @@ export const bucketName = bucket.id;
...
+```
+{{% /choosable %}}
+
+{{% choosable language python %}}
+
+```python
+import mimetypes
 
+# Detect the MIME type of the file
+mime_type, _ = mimetypes.guess_type('path/to/your/local/file.ext')
```
cnunciato, line 146: Hm -- does this actually give the expected MIME type? It seems to give me `application/vnd.novadigm.ext`. Should we not just stick with `.txt`?

interurban (approved):
Thx Troy

themes/default/content/blog/next-level-iac-package-ecosystems/index.md
```diff
@@ -50,7 +54,7 @@ const fileContents = fs.readFileSync(filePath, {encoding: 'utf-8'});
 // Upload the file to the S3 bucket.
 const bucketObject = new aws.s3.BucketObject("my-bucket-object", {
     bucket: bucket, // Reference to the bucket created above.
-    key: "file.txt", // This will be the file's name in the bucket.
+    key: "file.ext", // This will be the file's name in the bucket.
```
cnunciato, line 57: Curious why we're changing this from `txt` to `ext`?

themes/default/content/blog/next-level-iac-package-ecosystems/index.md
```diff
@@ -60,26 +64,95 @@ export const bucketName = bucket.id;
...
+{{% /choosable %}}
+
+So, to work that into our Pulumi program from above, we only need to make a few small edits:
+
+{{< chooser language "typescript,python" />}}
+
+{{% choosable language typescript %}}
 
 ```typescript {hl_lines=[4,21]}
 import * as pulumi from '@pulumi/pulumi';
```
cnunciato, line 158: There's a bug in the code two lines down from here. It should be:

```
import * as mime from 'mime-types';
```

themes/default/content/blog/next-level-iac-package-ecosystems/index.md
```diff
@@ -111,6 +184,45 @@ export const bucketName = bucket.id;
...
+import mimetypes
+
+# Create an S3 bucket
+bucket = aws.s3.Bucket('my-bucket')
+
+# The path to the file to upload to S3
+file_path = "path/to/your/local/file.ext"  # Replace with the path to the file you want to upload
+
+# Read the contents of the file to be uploaded to S3.
+with open(file_path, 'rb') as file:
```
cnunciato, line 204: I get an error when I run this as written. Looks like it should be this?

```suggestion
with open(file_path) as file:
```

themes/default/content/blog/next-level-iac-package-ecosystems/index.md
```diff
@@ -60,26 +64,95 @@ export const bucketName = bucket.id;
...
+
+```python
+content_type='text/plain')  # MIME-type of the file. Adjust if you upload a different type of file.
+```
+{{% /choosable %}}
+
 Yeah, hard-coding the MIME type for every file I need to upload is a not going to scale. In fact, I’m probably going to get a random directory full of files I need to iterate over, each with different file types, and it will change every time it runs, and we won’t know what type of file it is ahead of time.
 
-Packages to the rescue!! Turns out that determining the MIME type of a file in TypeScript only involves importing the community-created [`mime-types`][npm-mimetypes] library and a single function call to `mime.lookup(...)`:
+Packages to the rescue!! Turns out that determining the MIME type of a file in TypeScript only involves importing a community-created [`mime-types`][npm-mimetypes] [library][python-mimetypes] and a single function call to guess the mimetype:
```
cnunciato, line 125: This confused me a bit -- partly because it's two links that look like one, but also because in the case of Python, it isn't a community-created library, but part of (I think) the Python standard library. So we may way to choose an actual package instead, to illustrate the point of the post a bit more clearly. (Even if some users may realize this is built into Python itself.)

themes/default/content/blog/next-level-iac-package-ecosystems/index.md
```diff
@@ -60,26 +64,95 @@ export const bucketName = bucket.id;
...
+
+```python
+import pulumi
+import pulumi_aws as aws
+
+# Create an S3 bucket
+bucket = aws.s3.Bucket('my-bucket')
+
+# The path to the file to upload to S3
+file_path = "path/to/your/local/file.ext"  # Replace with the path to the file you want to upload
```
cnunciato, line 81: Noting this here also, in case we want to change it back to `.txt`.

themes/default/content/blog/next-level-iac-package-ecosystems/index.md
```diff
@@ -60,26 +64,95 @@ export const bucketName = bucket.id;
...
+
+# Read the contents of the file to be uploaded to S3.
+with open(file_path, 'rb') as file:
```
cnunciato, line 84: When I run this, I get an error. Seems like it should be this (which results in a `str`)?

```suggestion
with open(file_path) as file:
```

themes/default/content/blog/next-level-iac-package-ecosystems/index.md
```diff
@@ -60,26 +64,95 @@ export const bucketName = bucket.id;
...
+    file_contents = file.read()
+
+# Upload the file to the S3 bucket
+bucket_object = aws.s3.BucketObject('my-bucket-object',
+                                bucket=bucket.id, # Reference to the bucket created above.
+                                key='file.ext',  # This will be the file's name in the bucket.
```
cnunciato, line 90: Also here, again just so it doesn't get lost.

themes/default/content/blog/next-level-iac-package-ecosystems/index.md
thoward: It was meant to be a pseudo-name, with `.ext` short for extension, as in this is a file with an extention, rather than to imply a specific file with a specific file type.

themes/default/content/blog/next-level-iac-package-ecosystems/index.md
thoward: Well, one of the issues was that I wrote this to be Typescript specific, then I was asked to add Python in today. This made it a little complicated because the example works for Typescript, but not as well as for Python. This is one of the issues when we talk about package ecosystems -- they are not homogenous and what is simple and easy for one is not going to be the same in the next one, so multilingual examples will end up being more complicated to show. That said, even though the Python library is part of the core library, it is still a community created library, as the entire Python language and standard library is community created and maintained.

That said, I could update this to use [mimelib](https://pypi.org/project/mimelib/) which is an external package, and all it really does is add some gloss over the built-in package.

themes/default/content/blog/next-level-iac-package-ecosystems/index.md
thoward: oh interesting, something in my editor changed these to fancy single quotes. weird. will fix.

themes/default/content/blog/next-level-iac-package-ecosystems/index.md
thoward: Also, the narrative flow is talking about a single example and how to change it, as opposed to two at once that are subtly different. I'll reword this a bit and update to use `mimelib`.

themes/default/content/blog/next-level-iac-package-ecosystems/index.md
thoward: That said, if it's confusing to you, it will probably be confusing to others. I'll change it back to `.txt` everywhere.

themes/default/content/blog/next-level-iac-package-ecosystems/index.md
thoward: Hmm.. Yeah, I just assumed the `BucketObject.content` wanted binary data. Looking over it, apparently it wants.. text/string? Strange API choice, TBH. So however you can use `source` which takes binary. I noticed that most of the examples out there for Python actually just use `source=pulumi.FileAsset('./file.txt')` rather than using the `content` property. So maybe I'll update both the Typescript and the Python examples to use `FileAsset` instead. I was mostly just working from the [example given in the AI Answers page](https://www.pulumi.com/ai/answers/fZ6JumwyAXHcXys5tFFkqq/uploading-files-to-amazon-s3-using-typescript), which uses `content`.  Realizing now that is probably not best practice, so...

thoward:
Updated per @cnunciato feedback.

themes/default/content/blog/next-level-iac-package-ecosystems/index.md
thoward: Ended up sticking w/ the standard `mimetypes` module because changing it would have made things even more confusing. Instead just changed the language in the text.

cnunciato (approved):
Leaving this here for ya to keep you unblocked. Thanks for entertaining my suggestions! ⭐️

